PR #34, opened 2021-03-23
Gene Set Format documentation

### LINKED ISSUE #25: The gene set data format MUST be documented
Per the requirements in [A curator can include gene sets with published datasets.](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell/104), the gene set data format must be documented for curators and for reference in the Gene Sets tab in the portal.

brianraymor:
Closes #25

Additional requirements to be added per alignment with cellxgene explorer:

The cellxgene gene set data format is a Tidy CSV (comma-separated values) file using ~UTF-8~ ASCII encoding

cellxgene portal will warn and fail when the following illegal ASCII characters or sequences are detected in any **required** field value:

* control characters (decimal 0-31)
* DEL (decimal 127)
* leading spaces (decimal 32) in a field - `     This is an example`
* trailing spaces (decimal 32) in a field - `This is an example     ` 
* multiple spaces (decimal 32) "internal" to a field - `This     is an example`

### DIFF
--- a/docs/gene_sets.md
+++ b/docs/gene_sets.md
@@ -54,7 +54,7 @@ Each subsequent row describes a gene in a gene set.
 
 ### Values
 
-The values for `gene_set_name`, `gene_set_description`, and `gene_symbol` MUST not contain illegal ASCII characters or sequences. If the following cases are detected, validation MUST display a warning and fail the upload:
+The values for `gene_set_name`, `gene_set_description`, and `gene_symbol` MUST NOT contain illegal ASCII characters or sequences. If the following cases are detected, validation MUST display a warning and fail the upload:
 
 * control characters (decimal 0-31)
 * DEL (decimal 127)
--- a/docs/gene_sets.md
+++ b/docs/gene_sets.md
@@ -54,7 +54,7 @@ Each subsequent row describes a gene in a gene set.
 
 ### Values
 
-The values for `gene_set_name`, `gene_set_description`, and `gene_symbol` MUST NOT contain illegal ASCII characters or sequences. If the following cases are detected, validation MUST display a warning and fail the upload:
+The values for `gene_set_name`, `gene_set_description`, and `gene_symbol` MUST NOT contain illegal ASCII characters or sequences. If the following cases are detected, validation MUST display an error message and fail the upload:
 
 * control characters (decimal 0-31)
 * DEL (decimal 127)
@@ -68,7 +68,7 @@ The values for `gene_set_name`, `gene_set_description`, and `gene_symbol` MUST N
 
 The `gene_set_name` column MUST contain a value.
 
-If the `gene_set_name` is missing, validation MUST display a warning and fail the upload. This is illustrated by **~~?~~** in the example:
+If the `gene_set_name` is missing, validation MUST display an error message and fail the upload. This is illustrated by **~~?~~** in the example:
 
 | gene_set_name | gene_set_description | gene_symbol | gene_description |
 |---------------|----------------------|-------------|------------------|
@@ -77,7 +77,7 @@ If the `gene_set_name` is missing, validation MUST display a warning and fail th
 | club.cell     |                      | CYP2F2      | description      |
 
 
-All `gene_symbol(s)` for a `gene_set_name` MUST be on contiguous rows; otherwise, an out-of-order row is interpreted as a duplicate `gene_set_name`. In this case, validation MUST display a warning and fail the upload. This is illustrated by **~~club.cell~~** in the example:
+All `gene_symbol(s)` for a `gene_set_name` MUST be on contiguous rows; otherwise, an out-of-order row is interpreted as a duplicate `gene_set_name`. In this case, validation MUST display an error message and fail the upload. This is illustrated by **~~club.cell~~** in the example:
 
 ---
 | gene_set_name | gene_set_description | gene_symbol | gene_description |
@@ -89,7 +89,7 @@ All `gene_symbol(s)` for a `gene_set_name` MUST be on contiguous rows; otherwise
 | **~~club.cell~~**     |                      | CD163       | description      |
 
 
-When new gene sets are being added to a data collection on the portal, validation MUST detect `gene_set_name` collisions with current gene sets in the collection, display a warning, and fail the upload.  <br><br>
+When new gene sets are being added to a data collection on the portal, validation MUST detect `gene_set_name` collisions with current gene sets in the collection, display an error message, and fail the upload.  <br><br>
 
 ### `gene_set_description`
 
@@ -97,7 +97,7 @@ When new gene sets are being added to a data collection on the portal, validatio
 
 The first instance of a `gene_set_description` column for a specific `gene_set_name` MUST contain a value. All other instances are ignored in subsequent rows for the same `gene_set_name`.
 
-If the first instance of the `gene_set_description` is missing, validation MUST display a warning and fail the upload. This is illustrated by **~~?~~** in the example:
+If the first instance of the `gene_set_description` is missing, validation MUST display an error message and fail the upload. This is illustrated by **~~?~~** in the example:
 
 ---
 
@@ -112,7 +112,7 @@ If the first instance of the `gene_set_description` is missing, validation MUST 
 
 ---
 
-The `gene_symbol` column MUST contain a value that is unique for the `gene_set_name`. Validation MUST display warnings for a duplicate `gene_symbol` and fail the upload. This is illustrated by **~~CCKAR~~** in the example:
+The `gene_symbol` column MUST contain a value that is unique for the `gene_set_name`. Validation MUST display an error message for a duplicate `gene_symbol` and fail the upload. This is illustrated by **~~CCKAR~~** in the example:
 
 | gene_set_name | gene_set_description | gene_symbol | gene_description |
 |---------------|----------------------|-------------|------------------|
--- a/docs/gene_sets.md
+++ b/docs/gene_sets.md
@@ -47,8 +47,6 @@ The first row MUST contain header columns using reserved names in the following 
 
 Publishers MAY include additional header columns. It is RECOMMENDED that these custom columns observe the same self-documenting style - `column_name` and `column_description` for easier comprehension by data consumers.
 
-Only the values of `gene_set_name` and `gene_set_description` are presented to data consumers in the portal UX. 
-
 ## Rows
 
 Each subsequent row describes a gene in a gene set.
@@ -129,4 +127,10 @@ The value for a `gene_symbol` SHOULD follow [cellxgene schema](https://github.co
 
 ---
 
-The `gene_description` column MAY contain a value.
+The `gene_description` column MAY contain a value. <br><br>
+
+## Presentation in the cellxgene UX
+
+`gene_set_name` and `gene_set_description` are presented to data consumers  viewing data collections in the portal. They also may download gene sets.
+
+Users can [**explore**](https://cellxgene.cziscience.com/e/6acb6637-ac08-4a65-b2d1-581e51dc7ccf.cxg/) a dataset and its related gene sets in the data collection. In the visualization, `gene_set_name`, `gene_set_description`,  `gene_symbol`, and `gene_description` are presented to the user. The user may color by the mean expression of the gene set, select cells from a histogram showing the distribution of mean expression, and plot gene sets on the scatter plot.
--- a/docs/gene_sets.md
+++ b/docs/gene_sets.md
@@ -1,0 +1,122 @@
+
+## Overview
+
+*— it made sense until I reached a period. Then what I thought I had comprehended vanished.*
+
+**Ben Lerner – The Reflections of a Reading**
+
+---
+
+Curators may add gene sets to their data collections hosted on the [cellxgene data portal](https://cellxgene.cziscience.com), by uploading a file that is *properly formatted* based on the requirements in this document.
+
+When the portal successfully validates a gene set file, individual gene sets are extracted and stored in the database. If validation fails for an individual gene set in the file, then the entire operation fails. No gene sets from the file are stored in the database.
+
+The portal **does not** store the uploaded file. 
+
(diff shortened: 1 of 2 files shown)


tihuan:
For reading as markdown: https://github.com/brianraymor/single-cell-curation/blob/gene-sets-format/docs/gene_sets.md

docs/gene_sets.md
```diff
@@ -0,0 +1,122 @@
...
+| club.cell     | description          | CCKAR       | description      | Pubmed ID XYZ123 | Primary Pubmed ID       |
+| club.cell     |                      | SCGB3A2     | description      | Pubmed ID ABC456 | Primary Pubmed ID       |
+| club.cell     |                      | CYP2F2      | description      | Pubmed ID DCF678 | Primary Pubmed ID       |
+| macrophage    | description          | CD68        | description      |                  |                         |
+| macrophage    |                      | CD163       | description      |                  |                         |
+
+
+## Mandatory Header
+
+The first row MUST contain header columns using reserved names in the following order:
```
brianraymor, line 41: Should the **reserved** names for column headers be enforced by validation? For example, it may be confusing to consumers if both Column A and Column E (part of custom headers) are `gene_set_name`.

tihuan: hmm great question! When constructing the JSON, additional params are namespaced, so at least there's no key collision there. 

I don't know how frequently users would use `gene_set_name` or any of the reserved header columns as one of the additional column names, and the rationale behind that. So if we don't think it's likely to happen, maybe we can just keep an eye on it until there's an actual need for such validation?

pablo-gar:
LGTM.

Is there a reason for the following?

> All `gene_symbol(s)` for a `gene_set_name` MUST be on contiguous rows; otherwise, an out-of-order row

I think it should be rather simple to ingest non-contiguous gene_set rows as opposed to burdening the user to adhere to this requirement.

brianraymor:
@bkmartinjr and I disagree on this requirement. 

What if an author copied and pasted a gene symbol to use in another gene set and forgot to change the gene set name from the original row? (I see similar cases all the time like last week when we had to update policy notices) I don't want to guess at intentions.

Could you share why you would create a gene set file with unordered rows?

pablo-gar:
Three cases of the top of my head:
1.  Some users won't be as organized and may add unordered genes to the file.
2. Adding one (forgotten) gene after having finalized the file, it may happen they add the gene at the end of the file.
3. (More common I'd think) These files will be likely created out of scripting or pipelines and not manually, it may be the case that ordering isn't taken into consideration.

docs/gene_sets.md
```diff
@@ -0,0 +1,122 @@
...
+| gene_set_name | gene_set_description | gene_symbol | gene_description |
+|---------------|----------------------|-------------|------------------|
+| club.cell     | description          | CCKAR       | description      |
+| club.cell     |                      | SCGB3A2     | description      |
+| club.cell     |                      | CYP2F2      | description      |
+| macrophage    | description          | CD68        | description      |
+| **~~club.cell~~**     |                      | CD163       | description      |
+
+
+When new gene sets are being added to a data collection on the portal, validation MUST detect `gene_set_name` collisions with current gene sets in the collection, display a warning, and fail the upload.  <br><br>
```
jahilton, line 84: What is the process by which a single gene set in a collection can be updated (e.g. adding a gene to a gene set)?

jahilton:
Is there additional documentation to guide submitters on how to define a 'gene set', including how a cellxgene user will interact with them?

docs/gene_sets.md
brianraymor: In a *private* collection, this would require the curator to:

1. Delete the single gene set
2. Upload a gene set file containing the updated single gene set
3. Re-link the single gene set to the datasets (if previously linked)

In the revision of a *public* collection which is in refinement, it currently follows the same model as private. We only support replace for datasets due to the requirement to maintain their CXG permalinks. 

Are you suggesting that the portal interpret collisions as `replace the gene set` in the collection? Based on your wrangling experience how often do you see updates to gene sets?

brianraymor:
RE _additional documentation_, @MaximilianLombardo is preparing a curation tutorial with further details. What precisely do you mean by "how to define a gene set"? 

My objective has been to restrict the content in this document to the file format with limited context for the publishing work flow and presentation hints like:

Only the values of `gene_set_name` and `gene_set_description` are presented to data consumers in the portal UX.

docs/gene_sets.md
jahilton: I wasn't suggesting anything, just seeking clarification.
Still trying to wrap my head around how the information we're capturing might translate to the gene sets you're collecting. But right now, I don't anticipate updates occurring frequently enough to warrant effort toward optimization. The process you described makes sense.

jahilton:
> RE _additional documentation_, @MaximilianLombardo is preparing a curation tutorial with further details. What precisely do you mean by "how to define a gene set"?

I'll await that tutorial & follow-up with clarifying questions at that time.

docs/gene_sets.md
```diff
@@ -0,0 +1,122 @@
...
+The first row MUST contain header columns using reserved names in the following order:
+
+* `gene_set_name`
+* `gene_set_description`
+* `gene_symbol`
+* `gene_description`
+
+Publishers MAY include additional header columns. It is RECOMMENDED that these custom columns observe the same self-documenting style - `column_name` and `column_description` for easier comprehension by data consumers.
+
+Only the values of `gene_set_name` and `gene_set_description` are presented to data consumers in the portal UX.
```
brianraymor, line 50: @signechambers1 - do you want to add a sentence about the use of the values in cellxgene explorer?

signechambers1: Sure, how about:

In the cellxgene explorer UX, users will be able to view `gene_set_name`, `gene_set_description` along with each `gene_symbol` and `gene_description` for the given gene set. Users will be able to color by the mean expression of the gene set, select cells from a histogram showing the distribution of mean expression, and plot gene sets on the scatter plot.

signechambers1 (approved):

docs/gene_sets.md
```diff
@@ -0,0 +1,136 @@
...
+Publishers MAY include additional header columns. It is RECOMMENDED that these custom columns observe the same self-documenting style - `column_name` and `column_description` for easier comprehension by data consumers.
+
+## Rows
+
+Each subsequent row describes a gene in a gene set.
+<br><br>
+
+### Values
+
+The values for `gene_set_name`, `gene_set_description`, and `gene_symbol` MUST NOT contain illegal ASCII characters or sequences. If the following cases are detected, validation MUST display a warning and fail the upload:
```
bkmartinjr, line 57: complete nit (ie, feel free to ignore):  to my eye, "display warning and fail" is contradictory.  Perhaps, "display a message and fail", or "display an error message and fail"?

docs/gene_sets.md
```diff
@@ -0,0 +1,136 @@
...
+If the `gene_set_name` is missing, validation MUST display a warning and fail the upload. This is illustrated by **~~?~~** in the example:
+
+| gene_set_name | gene_set_description | gene_symbol | gene_description |
+|---------------|----------------------|-------------|------------------|
+| club.cell     | description          | CCKAR       | description      |
+|     **~~?~~** |                      | SCGB3A2     | description      |
+| club.cell     |                      | CYP2F2      | description      |
+
+
+All `gene_symbol(s)` for a `gene_set_name` MUST be on contiguous rows; otherwise, an out-of-order row is interpreted as a duplicate `gene_set_name`. In this case, validation MUST display a warning and fail the upload. This is illustrated by **~~club.cell~~** in the example:
```
bkmartinjr, line 80: As Brian noted, I think the contiguity requirement is onerous and should be removed.  It causes as many problems as it solves.

docs/gene_sets.md
```diff
@@ -0,0 +1,136 @@
...
+
+When new gene sets are being added to a data collection on the portal, validation MUST detect `gene_set_name` collisions with current gene sets in the collection, display a warning, and fail the upload.  <br><br>
+
+### `gene_set_description`
+
+---
+
+The first instance of a `gene_set_description` column for a specific `gene_set_name` MUST contain a value. All other instances are ignored in subsequent rows for the same `gene_set_name`.
+
+If the first instance of the `gene_set_description` is missing, validation MUST display a warning and fail the upload. This is illustrated by **~~?~~** in the example:
```
bkmartinjr, line 100: I believe gene set description should be optional.

bkmartinjr (changes requested):
I don't favor the strict requirements on ordering and required values for description fields.  Those are potentially fine for a curation decision, but in a fundamental transfer encoding like this, they rule out other use cases.

Ideally, I would like a more relaxed format that would also be usable by "desktop" cellxgene.  If curators apply a more stringent policy on top of that, all the better.  But coding into the file format seems inappropriate.

docs/gene_sets.md
```diff
@@ -0,0 +1,136 @@
...
+| **~~club.cell~~**     |                      | CD163       | description      |
+
+
+When new gene sets are being added to a data collection on the portal, validation MUST detect `gene_set_name` collisions with current gene sets in the collection, display a warning, and fail the upload.  <br><br>
+
+### `gene_set_description`
+
+---
+
+The first instance of a `gene_set_description` column for a specific `gene_set_name` MUST contain a value. All other instances are ignored in subsequent rows for the same `gene_set_name`.
```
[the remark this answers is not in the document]
brianraymor: In the portal case, the _reader_ is the portal which is presenting **a** gene set description to the consumer in its UX for the list of gene sets. This is already implemented in portal validation.

docs/gene_sets.md
brianraymor: I will update.

docs/gene_sets.md
brianraymor: Pablo and you have convinced me. We will remove the requirement from the portal and this document.

docs/gene_sets.md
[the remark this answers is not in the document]
brianraymor: I'm unsure how to interpret the [desktop requirements:](https://docs.google.com/document/d/1O-Os0Bg6a9FBKAA1KIMR12KESAEhj2pD32lJFC5OGko/edit#)

_gene_set_description (**a field cannot be null**)
The **first non null field** for gene_set_description, all other fields will be ignored_

which reads like at least one `gene_set_description` for a gene set MUST NOT be null which is not quite optional.

docs/gene_sets.md
brianraymor: I'm unsure how to interpret the [desktop requirements:](https://docs.google.com/document/d/1O-Os0Bg6a9FBKAA1KIMR12KESAEhj2pD32lJFC5OGko/edit#)

_gene_set_description (**a field cannot be null**)
The **first non null field** for gene_set_description, all other fields will be ignored_

which reads like at least one `gene_set_description` for a gene set MUST NOT be null which is not quite optional.

brianraymor: Is the requirement:

At least one `gene_set_description` in a `gene_set` MUST have a valid (non-null) value; otherwise, validation fails.. The first valid field for `gene_set_description` sets the gene set description for the gene set. Further `gene_set_description` fields for the `gene_set` on subsequent rows MUST be ignored.

docs/gene_sets.md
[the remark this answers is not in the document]
bkmartinjr: Ah, I was assuming you were writing a file specification, not an implementation description. If the former, we could use this as a component in the supporting REST API doc.  Without, we will need to write another, slightly looser, spec for the actual software API & local app backend format.

How about a compromise:  clearly separate the prose which is indicative of `Portal policy` decisions, leaving the core of this doc usable by the other contexts?  That would avoid the situation where we have different docs, which will likely diverge over time.

docs/gene_sets.md
bkmartinjr: the quoted prose is obsolete (my mistake) - I believe that in discussions with Signe we changed it to better support creator (not just final publication) scenarios.   That said, you guys are the decision maker here, and I think we could go either way.

bkmartinjr: Please look at the REST API spec - it is more accurate, and reflects the final decisions Signe made.  She did not have a chance to fully update the desktop requirements doc.

docs/gene_sets.md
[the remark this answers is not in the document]
brianraymor: I wrote a file specification based on the product requirements that were then implemented in the portal. I'm not documenting an implementation after the fact. Similar to the schema, the audience is data curators which is why this is published in the single-cell-curation repo.

docs/gene_sets.md
brianraymor: The [REST API](https://docs.google.com/document/d/12Wce-r2OvLGnuUWm9abAiZuj-kELz_LYTEvv6KZLXXo/edit#) states [my emphasis]:

> Where a gene set has multiple gene rows, and they differ in gene_set_description value, **any implementation may be used to select an appropriate gene set description.** **Desktop semantics defined** [here](https://docs.google.com/document/d/1O-Os0Bg6a9FBKAA1KIMR12KESAEhj2pD32lJFC5OGko/edit#) (see section 1F).

which returns me to the specification that I quoted above.

In this case, I guess _Portal semantics are what I wrote in this document._

signechambers1: @brianraymor your interpretation and rewording of what I wrote in the desktop requirements is correct (and is worded much more clearly). I agree with you that requiring gene_set_description for the portal is best for the users since the end user of gene sets will likely be different than the people who created them and the description will provide more context.

@bkmartinjr for desktop, I agree that gene_set_description should be optional since the users creating the gene sets are more likely to also be the ones exploring them in the explorer. I'll update the requirements doc to reflect this (and point out the difference with portal).

For posterity, I don't see the difference in requirements between desktop and portal for gene_set_description as problematic. A user may want to download gene sets from explorer and publish these gene sets in portal but at that point adding a description is in their best interest. This workflow happens infrequently (publishing data doesn't happen very often) and there are multiple steps a user needs to take before publishing data anyway.

Issue #27, opened 2021-03-08
Validator should check that one item of dict AnnData.uns["layer_descriptions"] is "raw"


pablo-gar:
According to schema v1.1.0 all layers must be described in `AnnData.uns["layer_descriptions"]` which is a dictionary mapping from `{layer_name: layer_description, ...}`

One of `layer_description` must be "raw" (not "Raw", "Raw counts", etc). Otherwise there [will be an error](https://github.com/chanzuckerberg/corpora-data-portal/blob/ffca067b9e4aea237fa2bd7c7a9cbc5813ebd449/backend/corpora/dataset_processing/process.py#L257-L258) when processing the data in the corpora data portal.

The validator should check for this.

brianraymor:
Outdated.

Issue #22, opened 2021-02-26
HGNC Database version is unknown  


pablo-gar:
The HGNC database version is unknown. Its version should be tracked and documented somewhere.

[https://github.com/chanzuckerberg/single-cell-curation/blob/main/cellxgene_schema_cli/cellxgene_schema/hgnc_complete_set.txt.gz](https://github.com/chanzuckerberg/single-cell-curation/blob/main/cellxgene_schema_cli/cellxgene_schema/hgnc_complete_set.txt.gz)

brianraymor:
If/when we refresh the HGNC download, it should be tied to schema version updates to ensure that we have consistent validation behavior for a schema version. Are there actual HGNC versions or simply [monthly refreshes](https://www.genenames.org/download/statistics-and-files/) ?

And we could probably reverse engineer the current date/version by comparing with older versions in the [HGNC archive](ftp://ftp.ebi.ac.uk/pub/databases/genenames/hgnc/archive/).

pablo-gar:
Correct, it's only monthly refreshes. I checked the archive, the earliest one is from 2020-06-01 and none of them matches with the one we have. I'm assuming Marcus got it earlier than that.

brianraymor:
Also see https://github.com/chanzuckerberg/corpora-data-portal/issues/320

pablo-gar:
Also note that our approach to standardize gene symbols occurs when using `cellxgene-schema apply` [here](https://github.com/chanzuckerberg/single-cell-curation/blob/f2b872edec347acb37959ba006a8387c80550385/cellxgene_schema_cli/cellxgene_schema/gene_symbol.py#L12-L24).

Thus curated datasets that did not go through our apply command are likely to have inconsistent gene names. e.g. external contributors.

brianraymor:
Outdated due to the change to ENSEMBL and pinned references.

PR #46, opened 2021-04-13
Ambrosecarr/schema update

ambrosejcarr:
This PR creates the following changes to the schema: 

* Requirements that were previously weakly worded have been updated to conform to [IETF RFC2119 definitions](https://tools.ietf.org/html/rfc2119). 
* A reference to using cellxgene prepare to create embeddings has been removed. The documentation describing cellxgene prepare has desktop data format requirements and referencing them would be confusing. In my view, it is OK to remove this reference because curators aren't in a position to run this software on behalf of scientists, and the computational scientists we work with would only very rarely want to use the default scanpy commands to generate their embeddings. 
* The section about matrix layers has been rewritten and restructured into "all assays" and "assay specific" requirements. 
* Context has been added to describe how the required "dataset title" field appears on the portal. In my opinion we have no "good" examples of dataset names, and should discuss updates during the revision that will be run prior to broad release. 
* The schema was updated to have an opinion (SHOULD) about where to store raw data (`.raw`). 
* Removed a reference to features being usually "genes or transcripts". Right now we only accept genes. Transcripts are a future data modality and will require a different feature namespace.
* Editorial changes were made to replace future tense with present tense in appropriate cases.

This PR **DOES NOT** address several open questions. If the PR is accepted, tickets should be made to track the following issues, if they do not already exist: 
* Cases where ontology terms and values do not match have been observed (e.g. `organism` string MUST match NCBITaxon term.) The question was raised as to whether validation should simply correct improper casing? I believe it should, and did not make edits to clarify this requirement (and many other similar requirements for other terms) in the schema. 
* Dataset requirements need to be merged into one place. Some are on cellxgene.github.io. This is not addressed by this PR and should be addressed when executing a minimal revision of our documentation. @ambrosejcarr to create this epic and related ticket. 
* We need to be specific about our MGI and HGNC versions. This is not done here and will presumably be done during the first revision.
* `color_map` review validation code in explorer and make a decision whether to include in schema validation.

Finally, for ease of review, I suggest a future reformatting of the schema document so that one sentence is on each line. This will help github track when specific words have been changed in otherwise identical sentences without requiring that the editor maintain the same sentence fragments on each line.


brianraymor:
There are some `must` statements near the end of the document to review:

See color_map in [Presentation Hints](https://github.com/chanzuckerberg/single-cell-curation/blob/e0d3a2ff1dfa85f79b2103004c07654f610a8ff2/docs/corpora_schema.md#presentation-hints):

> Submitters can include a field called "{field}_colors" for any other categorical integer metadata field. The value **must** be an array of one of fourcolor specifications:
> 
> Hex triplet string, as in "#08c0ff"and each string **must** be a hex color code.

and [ScanpyAnndata](https://github.com/chanzuckerberg/single-cell-curation/blob/e0d3a2ff1dfa85f79b2103004c07654f610a8ff2/docs/corpora_schema.md#scanpyanndata)

> The obs dataframe **must** store all the cell-level metadata except for embeddings. So, there **must** be columns named "tissue", "assay", "disease", and "cell_type".

One **should** case to review (and you can use MAY for optional):

>  In anndata files, the layer_descriptions dictionary **should** have a key "X" and **optionally** "raw.X" to describe those layers.

docs/corpora_schema.md
```diff
@@ -34,30 +35,44 @@ schema. If it does not, it will reject the dataset with an appropriate error mes
...
-*   **Unique feature identifiers**. Every feature (usually a gene or transcript) also needs a unique identifier. This is occasionally not present because
+*   **Unique feature identifiers**. Every gene feature MUST be assigned a unique identifier. This is occasionally not present because
     of one-to-many mappings between gene symbols and other gene ids. In cases where there are duplicated feature identifiers, they will need to be appropriately
     combined before submission. For example, raw counts will be summed and logged counts will be exponentiated, summed, and logged.
-*   **No PII**. No metadata can be personally identifiable: no names, dates of birth, specific locations, etc. There's a
-    [list](https://docs.google.com/document/d/1nlUuRiQ8Awo_QsiVFi8OWdhGONjeaXE32z5GblVCfcI/edit?usp=sharing).
+    This is needed for the explorer to function. When a user requests gene expression information, the explorer needs to be able to unambiguously return a
+    single value.
+*   **No PII**. Curators agree to this requirement as part of the data submission policy. However, it is not strictly enforced in our validation tooling because it
```
brianraymor, line 43: Some extra bullets snuck in to the PII text.

docs/corpora_schema.md
```diff
@@ -34,30 +35,44 @@ schema. If it does not, it will reject the dataset with an appropriate error mes
...
+| scRNA-seq (non-UMI, e.g. SS2)         | Values MUST be one of read counts (e.g. FeatureCounts) or  estimated fragments (e.g. output of RSEM).                                         | REQUIRED      |                                                                                                    | STRONGLY RECOMMENDED | OPTIONAL     |
+| Accessibility (e.g. ATAC-seq, mC-seq) |                                                                                                                                               | NOT REQUIRED  | Values MUST correspond to gene features (e.g. HGNC symbols if human)                               | REQUIRED             | OPTIONAL     |
 
 ### Schema Version
 
-Datasets in the Data Portal must store the version of the schema they follow (that is, the version of this document) as
+Datasets in the Data Portal MUST store the version of the schema they follow (that is, the version of this document) as
```
brianraymor, line 75: The link to the encoding is still broken. Pablo [found](https://chanzuckerbergteam.slack.com/archives/C018B64J3HN/p1617907599111700) the missing encoding. Do we really need a separate encoding field?

Also _and describes ~techincal~ technical details_

docs/corpora_schema.md
```diff
@@ -69,21 +84,19 @@ corpora_encoding_version|Follows [SemVer](https://semver.org/) conventions
...
+To support data integration, each cell MUST have the following metadata:
 
 **Field name**|**Constraints**
 :--|:--
-tissue|string (see note below regarding cell culture and organoids)
+tissue|string. This field SHOULD be appended with " (cell culture)" or " (organoid)" if appropriate.
 assay|string
 disease|string
-cell\_type|string
+cell\_type|string.
```
brianraymor, line 94: Nit. Remove the `.` if there are no further requirements or guidance.

docs/corpora_schema.md
```diff
@@ -96,17 +109,21 @@ development\_stage\_ontology\_term\_id|HsapDv term if human, child of EFO:000039
 
 The `tissue_ontology_term_id` field must be appended with " (cell culture)" or " (organoid)" if appropriate.
```
brianraymor, line 110: From my earlier notes:

[**Editorial**: *Inline this sentence in the table. Keep requirements together. This is a SHOULD?*]

docs/corpora_schema.md
```diff
@@ -96,17 +109,21 @@ development\_stage\_ontology\_term\_id|HsapDv term if human, child of EFO:000039
...
 
-The `ontology_term_id` fields may be empty strings when no appropriate ontology value is available, for example if the
-`cell_type` field describes a cell type for which no ontology term exists or if the `ethnicity` is unknown. If the
-field is not empty, then it must be an [OBO-format ID](http://www.obofoundry.org/id-policy.html), meaning it is a CURIE
+cellxgene requires ontology terms to enable search, comparison, and integration of data. When no appropriate ontology value is available, then the most
+precise accurate term MUST be used. For example if the `cell_type` field describes a relay interneuron, but the most specific available term in the CL
+ontology is CL:0000099 ("Interneuron"), then the interneuron term can be used to fulfill this requirement, and ensures that users searching for "neuron"
+are able to find these data. Users will still be able to access any additional more specific cell type annotations that have been submitted with the
```
brianraymor, line 115: _~any additional~ more specific_ - sounds like a Gilbert and Sullivan lyric.

docs/corpora_schema.md
```diff
@@ -119,15 +136,12 @@ layer\_descriptions|Mapping from {layer\_name: layer\_description, ...} Each des
...
-
-**Field name**|**Description**
-:--|:--
-title|String that identifies the dataset
-
+* Each dataset MUST have at least one **embedding**, a mapping from each cell to a tuple of floats of length at least 2. These are usually generated by algorithms
+  like umap or tsne, but can also coordinates of cells in spatial assays. They are used to display the dataset in the Explorer.
+* Each dataset MUST have a title. This is a string that describes and differentiates the dataset from others in the collection. It will be displayed on a page that
+  also has the collection name. For example, in the collection
+  [Cells of the human heart](https://cellxgene.cziscience.com/collections/b52eb423-5d0d-4645-b217-e1c6d38b2e72), the first dataset name is "All — Cells of the
```
brianraymor, line 143: Cells of the **adult** human heart

docs/corpora_schema.md
```diff
@@ -185,17 +186,11 @@ The color code at the nth position in the array corresponds to category n in the
...
-Each of these formats has a way to include metadata along with the count data, so a submission can be entirely contained within a single file. 
-
-
 ### **Scanpy/AnnData**
 
-[Anndata](https://anndata.readthedocs.io/en/stable/) is the hdf5-based file format used by scanpy. There is a python library for interacting with it. The
-count data is stored in an attribute `X` of shape (# of cells, # of genes). `X` can either be a numpy.ndarray or a scipy.sparse.spmatrix.
+The Data Portal requires submitted count matrices and associated metadata to be in [Anndata](https://anndata.readthedocs.io/en/stable/) format, the hdf5-based
+file format used by scanpy. There is a python library for interacting with it. The count data is stored in an attribute `X` of shape (# of cells, # of genes).
+`X` can either be a numpy.ndarray or a scipy.sparse.spmatrix.
```
brianraymor, line 193: `X` ~can either~ MUST be a numpy.ndarray or a scipy.sparse.spmatrix.

docs/corpora_schema.md
```diff
@@ -119,15 +136,12 @@ layer\_descriptions|Mapping from {layer\_name: layer\_description, ...} Each des
...
-like umap or tsne and are used to display the dataset in the Explorer. cellxgene provides a command line tool to add embeddings if they are missing.
-
-Datasets also must have a few other metadata fields for presentation:
-
-**Field name**|**Description**
-:--|:--
-title|String that identifies the dataset
-
+* Each dataset MUST have at least one **embedding**, a mapping from each cell to a tuple of floats of length at least 2. These are usually generated by algorithms
+  like umap or tsne, but can also coordinates of cells in spatial assays. They are used to display the dataset in the Explorer.
```
brianraymor, line 140: _can also coordinates_ ?

docs/corpora_schema.md
```diff
@@ -34,30 +35,44 @@ schema. If it does not, it will reject the dataset with an appropriate error mes
...
-layer. Note that AnnData objects have an attribute called [raw](https://anndata.readthedocs.io/en/latest/anndata.AnnData.raw.html#anndata.AnnData.raw). This
-could be a natural place to store the raw count matrix, but as long as the locations of the layers are specified, it can go anywhere.
-
+cellxgene's data requirements are tailored to optimize data reuse. Because each assay has different characteristics, our requirements differ by assay type. In general,
+cellxgene requires submission of "raw" data suitable for computational reuse, when a standard form exists, and strongly suggests that a "final" matrix suitable for
+visualization in the explorer be included. cellxgene uses `AnnData` for data ingestion, and uses `AnnData`'s `layers` functionality to accept multiple matrix layers,
+such as "raw" and "final". This imposes some requirements on data of all assay types:
+
+*   Anndata provides a [raw](https://anndata.readthedocs.io/en/latest/anndata.AnnData.raw.html#anndata.AnnData.raw) attribute. If a raw matrix is required for an assay
+    type, it SHOULD be stored in `AnnData.raw`, but it MAY be instead stored in `AnnData.layers["raw"]`.
```
brianraymor, line 55: it SHOULD be stored in `AnnData.raw`, but ~it~ MAY be ~instead~ stored in `AnnData.layers["raw"]`.

Also, what are both options supported?

docs/corpora_schema.md
```diff
@@ -34,30 +35,44 @@ schema. If it does not, it will reject the dataset with an appropriate error mes
...
+*   AnnData requires that matrix layers MUST have the same dimension, so raw count matrices MUST include the same cells and genes as the final. Because it is impractical to
+    retain all barcodes in raw and final matrices, we require that cells be filtered from both. By contrast, those wishing to reuse datasets require access to raw gene expression values,
+    so we require that genes are present in both datasets. Summarizing, any cell barcodes that are removed from the data MUST be filtered from both raw and final matrices.
```
brianraymor, line 58: "we" vexes me.

Because it is impractical to retain all barcodes in raw and final matrices, cells MUST be filtered from both. By contrast, those wishing to reuse datasets require access to raw gene expression values, so genes MUST be present in both datasets.

docs/corpora_schema.md
```diff
@@ -11,19 +11,20 @@ Date Last Modified: 2020-12-11
...
-cellxgene aims to support the publication and sharing of single-cell datasets as well as the construction of a data corpus that facilitates data integration
-across multiple tissues and experiments. Achieving the latter goal requires some harmonization of metadata and features in the cellxgene Data Portal. But if
-that harmonization is too onerous, it will burden the goal of rapid data sharing.
+cellxgene aims to support the publication, sharing, and exploration of single-cell datasets.
+Building on those published datasets, cellxgene seeks to create references of the phenotypes and composition of cells that make up human tissues.
+Creating references from multiple datasets requires some harmonization of metadata and features in the cellxgene Data Portal. But if that harmonization is too onerous, it
+will burden the goal of rapid data sharing.
 
-We balance these two goals by requiring datasets hosted in the cellxgene Data Portal to follow a small schema with only a few required fields. These fields
+We balance publishing and reference creation needs by requiring datasets hosted in the cellxgene Data Portal to follow a small schema with only a few required fields. These fields
```
brianraymor, line 19: ~We balance~ - cellxgene balances ...

docs/corpora_schema.md
```diff
@@ -11,19 +11,20 @@ Date Last Modified: 2020-12-11
...
 are expected to be very useful for data integration and also simply and readily available from data submitters.
 
-Note that the requirements in the schema are just the minimum required information. We expect that datasets will have additional metadata, and this
-will be preserved in datasets submitted to the Data Portal.
+Note that the requirements in the schema are just the minimum required information. Datasets often have additional metadata, and this
+is preserved in datasets submitted to the Data Portal.
```
brianraymor, line 23: Datasets often have additional metadata, ~and this~ [which] is preserved in datasets submitted to the Data Portal.

I'm on the fence about whether collisions require further discussion in this document or simply in the curation tools.

brianraymor:
Originally, `AnnData` was one of three possible ingestion formats. Therefore, it made sense to have a section at the end discussing how the schema was modeled in various formats.

At this point, it seems doubtful that additional formats will be supported. With that in mind, I'd recommend merging `AnnData` sections together.

docs/corpora_schema.md
ambrosejcarr: We closed on next steps on this question in the curation sync this morning. We are going to confirm with AnnData that layers are intended to replace raw. If that's the case, and we suspect it is, we will deprecate use of `AnnData.raw`. Both are supported for historic reasons - Marcus noted that the application logic was agnostic to where the data were stored because we required the mapping exist in layers, and therefore could derive the location from the file.

I will make this change in a separate PR.

docs/corpora_schema.md
ambrosejcarr: Thanks for catching this. I inlined it, but missed deleting the original sentence.

docs/corpora_schema.md
ambrosejcarr: I've replaced encoding link. I do not think we need a separate encoding, and in fact think it is confusing that we do, since we're standardizing on a single ingestion format. 

I think the integration will be significant and should be done in a separate PR.

docs/corpora_schema.md
ambrosejcarr: Updated to state _"but can also represent `(x, y)` coordinates of cells in spatial assays"_. Thanks.

docs/corpora_schema.md
ambrosejcarr: I've implemented the proposed edit. 

I didn't add any commentary about collisions at present. My weak opinion is that it belongs in the schema guide, since that document relates to application of a schema to a dataset.

ambrosejcarr:
> Originally, `AnnData` was one of three possible ingestion formats. Therefore, it made sense to have a section at the end discussing how the schema was modeled in various formats.
> 
> At this point, it seems doubtful that additional formats will be supported. With that in mind, I'd recommend merging `AnnData` sections together.

I think it would make sense to implement this change at the same time as integrating the encoding document, if we're going that route. My vote is to integrate the encoding document, but I'd like to hear @brianraymor @pablo-gar and @MaximilianLombardo 's opinions before making a decision.

docs/corpora_schema.md
```diff
@@ -34,32 +35,46 @@ schema. If it does not, it will reject the dataset with an appropriate error mes
...
+| scRNA-seq (UMI, e.g. 10x v3)          | Values MUST be de-duplicated molecule counts.                                                                                                 | REQUIRED      |                                                                                                    | STRONGLY RECOMMENDED | OPTIONAL     |
+| scRNA-seq (non-UMI, e.g. SS2)         | Values MUST be one of read counts (e.g. FeatureCounts) or  estimated fragments (e.g. output of RSEM).                                         | REQUIRED      |                                                                                                    | STRONGLY RECOMMENDED | OPTIONAL     |
+| Accessibility (e.g. ATAC-seq, mC-seq) |                                                                                                                                               | NOT REQUIRED  | Values MUST correspond to gene features (e.g. HGNC symbols if human)                               | REQUIRED             | OPTIONAL     |
```
pablo-gar, line 71: If we are removing the raw data requirement for ATAC, then [this](https://github.com/chanzuckerberg/single-cell-curation/issues/27) has to be addressed ASAP.

> One of `layer_description` must be "raw" (not "Raw", "Raw counts", etc). Otherwise there [will be an error](https://github.com/chanzuckerberg/corpora-data-portal/blob/ffca067b9e4aea237fa2bd7c7a9cbc5813ebd449/backend/corpora/dataset_processing/process.py#L257-L258) when processing the data in the corpora data portal.

docs/corpora_schema.md
```diff
@@ -69,44 +84,44 @@ corpora_encoding_version|Follows [SemVer](https://semver.org/) conventions
...
 assay\_ontology\_term\_id|EFO term
 disease\_ontology\_term\_id|MONDO term or [PATO:0000461](http://bioportal.bioontology.org/ontologies/PATO?p=classes&conceptid=PATO%3A0000461)
 cell\_type\_ontology\_term\_id|CL term
 ethnicity\_ontology\_term\_id|HANCESTRO term, "na" if non-human
 development\_stage\_ontology\_term\_id|HsapDv term if human, child of EFO:0000399 otherwise
 
-The `tissue_ontology_term_id` field must be appended with " (cell culture)" or " (organoid)" if appropriate.
+cellxgene requires ontology terms to enable search, comparison, and integration of data. When no appropriate ontology value is available, then the most
+precise accurate term MUST be used. For example if the `cell_type` field describes a relay interneuron, but the most specific available term in the CL
+ontology is CL:0000099 ("Interneuron"), then the interneuron term can be used to fulfill this requirement, and ensures that users searching for "neuron"
```
pablo-gar, line 112: What about the cases where no appropriate  cell type is available in CL? Should we recommend to use the top of tree in the ontology ([native cell](https://www.ebi.ac.uk/ols/ontologies/cl/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FCL_0000003&viewMode=PreferredRoots&siblings=false))? 

A recent dataset from the human embryo exemplifies the existence of unavailable cell types:
[https://docs.google.com/spreadsheets/d/14NfyiUWGOzgcRg7nUDiLNO9rMHUef6-gM1f7rlR5ymE/edit#gid=1775586196](https://docs.google.com/spreadsheets/d/14NfyiUWGOzgcRg7nUDiLNO9rMHUef6-gM1f7rlR5ymE/edit#gid=1775586196)

docs/corpora_schema.md
pablo-gar, line 71: If are removing the raw matrix requirement for ATAC-seq, [this](https://github.com/chanzuckerberg/single-cell-curation/issues/27) should be resolved ASAP. 

> One of `layer_description` must be "raw" (not "Raw", "Raw counts", etc). Otherwise there [will be an error](https://github.com/chanzuckerberg/corpora-data-portal/blob/ffca067b9e4aea237fa2bd7c7a9cbc5813ebd449/backend/corpora/dataset_processing/process.py#L257-L258) when processing the data in the corpora data portal.

docs/corpora_schema.md
```diff
@@ -69,44 +84,44 @@ corpora_encoding_version|Follows [SemVer](https://semver.org/) conventions
...
+tissue\_ontology\_term\_id|UBERON term. This field SHOULD be appended with " (cell culture)" or " (organoid)" if appropriate.
 assay\_ontology\_term\_id|EFO term
 disease\_ontology\_term\_id|MONDO term or [PATO:0000461](http://bioportal.bioontology.org/ontologies/PATO?p=classes&conceptid=PATO%3A0000461)
 cell\_type\_ontology\_term\_id|CL term
 ethnicity\_ontology\_term\_id|HANCESTRO term, "na" if non-human
 development\_stage\_ontology\_term\_id|HsapDv term if human, child of EFO:0000399 otherwise
 
-The `tissue_ontology_term_id` field must be appended with " (cell culture)" or " (organoid)" if appropriate.
+cellxgene requires ontology terms to enable search, comparison, and integration of data. When no appropriate ontology value is available, then the most
+precise accurate term MUST be used. For example if the `cell_type` field describes a relay interneuron, but the most specific available term in the CL
```
pablo-gar, line 111: What about the cases that there is no cell type available in the ontology? We could suggest the root of the ontology tree, or maybe one of the first children:

-- cell
----- native cell (A cell that is found in a natural setting)
----- cell _in vitro_
----- abnormal cell

A dataset of the human embryo exemplifies the existence of "unassignable" cell types
[https://docs.google.com/spreadsheets/d/14NfyiUWGOzgcRg7nUDiLNO9rMHUef6-gM1f7rlR5ymE/edit#gid=1775586196](https://docs.google.com/spreadsheets/d/14NfyiUWGOzgcRg7nUDiLNO9rMHUef6-gM1f7rlR5ymE/edit#gid=1775586196)

docs/corpora_schema.md
```diff
@@ -185,17 +184,11 @@ The color code at the nth position in the array corresponds to category n in the
...
-The Data Portal requires submitted count matrices and associated metadata to be in one of three formats: AnnData, Loom, or a Seurat v3 RDS. Other formats are rejected.
-Each of these formats has a way to include metadata along with the count data, so a submission can be entirely contained within a single file. 
-
-
 ### **Scanpy/AnnData**
 
-[Anndata](https://anndata.readthedocs.io/en/stable/) is the hdf5-based file format used by scanpy. There is a python library for interacting with it. The
-count data is stored in an attribute `X` of shape (# of cells, # of genes). `X` can either be a numpy.ndarray or a scipy.sparse.spmatrix.
+The Data Portal requires submitted count matrices and associated metadata to be in [Anndata](https://anndata.readthedocs.io/en/stable/) format, the hdf5-based
+file format used by scanpy. There is a python library for interacting with it. The count data is stored in an attribute `X` of shape (# of cells, # of genes).
```
pablo-gar, line 190:
```suggestion
file format used by scanpy — a python library for interacting with it. The count data is stored in an attribute `X` of shape (# of cells, # of genes).
```

docs/corpora_schema.md
```diff
@@ -69,44 +84,44 @@ corpora_encoding_version|Follows [SemVer](https://semver.org/) conventions
...
 gene symbols.
 
-Similarly if the features are mouse genes, then the feature ids must be [MGI](http://www.informatics.jax.org/mgihome/nomen/gene.shtml) gene symbols.
+Similarly if the features are mouse genes, then the feature ids SHOULD be [MGI](http://www.informatics.jax.org/mgihome/nomen/gene.shtml) gene symbols.
 
-Finally, the whole dataset must be annotated with fields that indicate the organism and describe the meanings of the [matrix layers](#Matrix-Layers):
+Finally, the whole dataset MUST be annotated with fields that indicate the organism and describe the meanings of the [matrix layers](#Matrix-Layers):
 
 **Field name**|**Constraints**
 :--|:--
```
pablo-gar, line 127: :--|:--
```suggestion
organism\_ontology\_term\_id|NCBITaxon term
layer\_descriptions|A dictionary with keys as layer names and values as free text description of the layers
version|A dictionary with a key `corpora_schema_version` and its value MUST be the schema version
```

docs/corpora_schema.md
```diff
@@ -69,44 +84,44 @@ corpora_encoding_version|Follows [SemVer](https://semver.org/) conventions
...
-field is not empty, then it must be an [OBO-format ID](http://www.obofoundry.org/id-policy.html), meaning it is a CURIE
+Ontology terms MUST use [OBO-format ID](http://www.obofoundry.org/id-policy.html), meaning they are a CURIE
 where the prefix identifies the ontology. For example `EFO:0000001` is a term in the `EFO` ontology.
 
-If the features of the dataset are human genes, then the feature ids must be [HGNC](https://www.genenames.org/about/guidelines/#!/#tocAnchor-1-7) approved
+If the features of the dataset are human genes, then the feature ids MUST be [HGNC](https://www.genenames.org/about/guidelines/#!/#tocAnchor-1-7) approved
 gene symbols.
 
-Similarly if the features are mouse genes, then the feature ids must be [MGI](http://www.informatics.jax.org/mgihome/nomen/gene.shtml) gene symbols.
+Similarly if the features are mouse genes, then the feature ids SHOULD be [MGI](http://www.informatics.jax.org/mgihome/nomen/gene.shtml) gene symbols.
```
pablo-gar, line 123:
```suggestion
Each layer MUST have a free text description of how it was produced, e.g. "counts per million".
```

docs/corpora_schema.md
```diff
@@ -69,44 +84,44 @@ corpora_encoding_version|Follows [SemVer](https://semver.org/) conventions
...
 
-Finally, the whole dataset must be annotated with fields that indicate the organism and describe the meanings of the [matrix layers](#Matrix-Layers):
+Finally, the whole dataset MUST be annotated with fields that indicate the organism and describe the meanings of the [matrix layers](#Matrix-Layers):
```
pablo-gar, line 124:
```suggestion
The whole dataset MUST be annotated with fields that indicate the organism and describe the meanings of the [matrix layers](#Matrix-Layers).
```

docs/corpora_schema.md
```diff
@@ -69,44 +84,44 @@ corpora_encoding_version|Follows [SemVer](https://semver.org/) conventions
...
```
pablo-gar, line 125:
```suggestion

Finally, the version of this schema MUST be indicated:
```

pablo-gar (commented):
As I was reading it, I thought it could be good to have the "enconding" document that it was once deleted. It is the perfect complement to this document as it provides a succinct descriptiom of how the metadata is encoded in the AnnData. 

Or maybe we could add an "Encoding" section here. @brianraymor any thoughts?

docs/corpora_schema.md
ambrosejcarr: It was our decision that while there may be other forms of data suitable for recording "raw" accessibility data, that we cannot specify a form for raw (observations, genes) data. I agree that #27 should be resolved soon.

docs/corpora_schema.md
ambrosejcarr: This belongs in the layers section, does it not?

docs/corpora_schema.md
ambrosejcarr: Should I split this sentence in two and move the matrix layers portion to the layers section? Could we be more specific about how this is implemented? What is the code string? (similar to e.g. `AnnData.layers["raw"]`)

docs/corpora_schema.md
ambrosejcarr: Present on line 76. Are you suggesting it be moved down?

ambrosejcarr:
> As I was reading it, I thought it could be good to have the "enconding" document that it was once deleted. It is the perfect complement to this document as it provides a succinct descriptiom of how the metadata is encoded in the AnnData.
> 
> Or maybe we could add an "Encoding" section here. @brianraymor any thoughts?

See [above discussion](https://github.com/chanzuckerberg/single-cell-curation/pull/46#issuecomment-819148256). Brian and I prefer integrating the encoding document. What is your opinion?

ambrosejcarr:
@pablo-gar I found it was a bit hard initially to understand the purpose of your final 4 comments. If you could indicate your rationale when making suggestions, that would help me integrate them. 

I think I grok your purpose now - you're making suggestions to condense "dataset" metadata in the final section of the integration metadata section. 

I think you're on to something with those suggestions, but there were pieces littered here and there throughout the schema document that needed to be moved or removed in addition to your suggestions. I've added a commit that makes these changes.

docs/corpora_schema.md
ambrosejcarr: Scanpy is not properly a python library for interacting with AnnData, so I think this edit would be misleading. The text in this paragraph is a bit awkward, but it wasn't written in this PR and it'll be changed if we integrate the encoding, so I'd prefer to leave editorial modifications for the next pass.

docs/corpora_schema.md
ambrosejcarr: Great suggestion, thanks. Please review 7d079fa for how I integrated.

docs/corpora_schema.md
ambrosejcarr: See 60378ec

docs/corpora_schema.md
ambrosejcarr: See 60378ec

docs/corpora_schema.md
ambrosejcarr: See 60378ec

docs/corpora_schema.md
ambrosejcarr: See 60378ec

pablo-gar:
> > As I was reading it, I thought it could be good to have the "enconding" document that it was once deleted. It is the perfect complement to this document as it provides a succinct descriptiom of how the metadata is encoded in the AnnData.
> > Or maybe we could add an "Encoding" section here. @brianraymor any thoughts?
> 
> @pablo-gar See [above discussion](https://github.com/chanzuckerberg/single-cell-curation/pull/46#issuecomment-819148256). Brian and I prefer integrating the encoding document. What is your opinion?

I agree with integration

pablo-gar:
> @pablo-gar I found it was a bit hard initially to understand the purpose of your final 4 comments. If you could indicate your rationale when making suggestions, that would help me integrate them.
> 
> I think I grok your purpose now - you're making suggestions to condense "dataset" metadata in the final section of the integration metadata section.
> 
> I think you're on to something with those suggestions, I went through the document and cleaned up other suggestions that belong in the dataset section (e.g. schema version). I've added a commit [60378ec](https://github.com/chanzuckerberg/single-cell-curation/commit/60378ec3894b217da3e036278d999a01fb9def8b) that makes these changes. Please review whether I've captured your suggestions.

Yeah, the order of the suggestions was off and I should have added a comment, but you got where I was getting to. Thanks!

MaximilianLombardo:
> > Originally, `AnnData` was one of three possible ingestion formats. Therefore, it made sense to have a section at the end discussing how the schema was modeled in various formats.
> > At this point, it seems doubtful that additional formats will be supported. With that in mind, I'd recommend merging `AnnData` sections together.
> 
> I think it would make sense to implement this change at the same time as integrating the encoding document, if we're going that route. My vote is to integrate the encoding document, but I'd like to hear @brianraymor @pablo-gar and @MaximilianLombardo 's opinions before making a decision.

Sorry, I'm a bit late to this thread. It seems like the encoding document is essentially what I had described in the curation tutorial. It goes over metadata defined by the schema and defines where in the anndata object it should go. I guess I prefer to keep that information in the tutorial. Can go either way.

brianraymor:
I recommend that you add to explain all those CAPITALS ...

`The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED" "MAY", and "OPTIONAL" in this document are to be interpreted as described in BCP 14, [RFC2119](https://www.rfc-editor.org/rfc/rfc2119.txt), and [RFC8174](https://www.rfc-editor.org/rfc/rfc8174.txt) when, and only when, they appear in all capitals, as shown here.`

docs/corpora_schema.md
```diff
@@ -11,19 +11,20 @@ Date Last Modified: 2020-12-11
...
 
-Note that the requirements in the schema are just the minimum required information. We expect that datasets will have additional metadata, and this
-will be preserved in datasets submitted to the Data Portal.
+Note that the requirements in the schema are just the minimum required information. Datasets often have additional metadata, which
+is preserved in datasets submitted to the Data Portal.
 
 ## Curation and Validation
 
-When a submitter is preparing a dataset for the cellxgene Data Portal, they will be able to use a command line tool to apply changes to their dataset so that it
+When a submitter is preparing a dataset for the cellxgene Data Portal, they can use the `cellxgene-schema apply` command line tool to apply changes to their dataset so that it
```
brianraymor, line 27: Mmm. Isn't it a two step process and wouldn't it be better to simply reference the CLI documentation:

You can use the config file to augment a dataset with  the schema using:

```
cellxgene-schema apply --source-h5ad original.h5ad --remix-config config.yml --output-filename remixed.h5ad
```

And then verify that the schema was properly added with:

```
cellxgene-schema validate remixed.h5ad
```

docs/corpora_schema.md
```diff
@@ -34,100 +35,110 @@ schema. If it does not, it will reject the dataset with an appropriate error mes
...
 ethnicity|string, "na" if non-human, "unknown" if not available
 development\_stage|string, "unknown" if not available
 
-The `tissue` field must be appended with " (cell culture)" or " (organoid)" if appropriate. Also, if the source of cells is cell culture or organiod, the
-`cell_type` field can be left empty.
-In addition to these free text fields (except sex), each cell must also have ontology values:
+In addition to these free text fields (except sex), each cell MUST also have ontology annotations:
+
+Ontology terms MUST use [OBO-format ID](http://www.obofoundry.org/id-policy.html), meaning they are a CURIE
+where the prefix identifies the ontology. For example `EFO:0000001` is a term in the `EFO` ontology.
```
brianraymor, line 94: Minor editorial. Both there's a hanging `:` and the sentences can be merged? 

Also - don't you want the label to match the ontology? (["blood" versus "Blood" for example](https://chanzuckerbergteam.slack.com/archives/C018B64J3HN/p1617144334028900)) These are really not "free text".

docs/corpora_schema.md
```diff
@@ -34,100 +35,110 @@ schema. If it does not, it will reject the dataset with an appropriate error mes
...
+tissue\_ontology\_term\_id|UBERON term. This field SHOULD be appended with " (cell culture)" or " (organoid)" if appropriate.
 assay\_ontology\_term\_id|EFO term
 disease\_ontology\_term\_id|MONDO term or [PATO:0000461](http://bioportal.bioontology.org/ontologies/PATO?p=classes&conceptid=PATO%3A0000461)
 cell\_type\_ontology\_term\_id|CL term
 ethnicity\_ontology\_term\_id|HANCESTRO term, "na" if non-human
 development\_stage\_ontology\_term\_id|HsapDv term if human, child of EFO:0000399 otherwise
 
-The `tissue_ontology_term_id` field must be appended with " (cell culture)" or " (organoid)" if appropriate.
+cellxgene requires ontology terms to enable search, comparison, and integration of data. When no appropriate ontology value is available, then the most
+precise accurate term MUST be used. For example if the `cell_type` field describes a relay interneuron, but the most specific available term in the CL
```
brianraymor, line 105: I see a line break between "most" and "precise" with the markdown is rendered.

docs/corpora_schema.md
```diff
@@ -34,100 +35,110 @@ schema. If it does not, it will reject the dataset with an appropriate error mes
...
-field is not empty, then it must be an [OBO-format ID](http://www.obofoundry.org/id-policy.html), meaning it is a CURIE
-where the prefix identifies the ontology. For example `EFO:0000001` is a term in the `EFO` ontology.
+#### Gene Metadata
 
-If the features of the dataset are human genes, then the feature ids must be [HGNC](https://www.genenames.org/about/guidelines/#!/#tocAnchor-1-7) approved
-gene symbols.
+Cellxgene uses standard gene symbols to ensure that all datasets it stores measure the same features and can therefore be integrated.
+If the samples being measured are human, then the feature ids MUST be [HGNC](https://www.genenames.org/about/guidelines/#!/#tocAnchor-1-7) approved
+gene symbols. If the features are mouse genes, then the feature ids SHOULD be [MGI](http://www.informatics.jax.org/mgihome/nomen/gene.shtml) gene symbols.
+For other organisms, gene symbols SHOULD be the accepted standard human readable symbols for that organism.
```
brianraymor, line 118: I'm not sure what _accepted standard human readable symbols for that organism_ means or entails for validation.

ambrosejcarr: As a SHOULD, it means nothing for validation. This is guidance for users submitting data from organisms which we don't maintain stronger recommendations.

docs/corpora_schema.md
ambrosejcarr: Addressed in #52

docs/corpora_schema.md
ambrosejcarr: Great catch. Addressed in #52

docs/corpora_schema.md
ambrosejcarr: I removed this from the document when refactoring in #52 to remove the encoding. It belongs in `schema_guide.md` (or the tutorial that will replace it). 

I also added an explanation of when to read this document so that users seeking curation advice are quickly sent to the schema_guide.

pablo-gar:
This version has some changes in the schema not present in v 1.1.0. (e.g. we allow non-RNA assays to not have a raw matrix) A change log may be helpful to keep track of the changes and inform schema updates of current datasets, in case we are doing so. 

And since the version is different than the previous one we should change its version number.

brianraymor:
The `PATO:0000461` link (https://bioportal.bioontology.org/ontologies/PATO?p=classes&conceptid=PATO%253A0000461) is returning a 404.

brianraymor:
RE [change log](https://github.com/chanzuckerberg/single-cell-curation/pull/46#issuecomment-825089894), I like that idea as an appendix. It's an approach that IETF employs when iterating on drafts. For example:

https://tools.ietf.org/html/draft-ietf-httpbis-http2-16#appendix-B

docs/corpora_schema.md
```diff
@@ -34,100 +35,110 @@ schema. If it does not, it will reject the dataset with an appropriate error mes
...
-The `ontology_term_id` fields may be empty strings when no appropriate ontology value is available, for example if the
-`cell_type` field describes a cell type for which no ontology term exists or if the `ethnicity` is unknown. If the
-field is not empty, then it must be an [OBO-format ID](http://www.obofoundry.org/id-policy.html), meaning it is a CURIE
-where the prefix identifies the ontology. For example `EFO:0000001` is a term in the `EFO` ontology.
+#### Gene Metadata
 
-If the features of the dataset are human genes, then the feature ids must be [HGNC](https://www.genenames.org/about/guidelines/#!/#tocAnchor-1-7) approved
-gene symbols.
+Cellxgene uses standard gene symbols to ensure that all datasets it stores measure the same features and can therefore be integrated.
+If the samples being measured are human, then the feature ids MUST be [HGNC](https://www.genenames.org/about/guidelines/#!/#tocAnchor-1-7) approved
```
jychien, line 116: A couple questions in regards to gene symbols:
- Is there a specific version of HGNC that you are validating against? For example, can we have EMR2 (a previous symbol) as a gene index rather than the current ADGRE2?
- Is there additional information on how to handle gene symbols that are redundant? Can there be a number appended as a suffix?

docs/corpora_schema.md
```diff
@@ -34,100 +35,110 @@ schema. If it does not, it will reject the dataset with an appropriate error mes
...
-well as the version of the particular encoding used. The encoding is documented
-[elsewhere](https://github.com/chanzuckerberg/corpora-data-portal/blob/main/backend/schema/corpora_schema_h5ad_implementation.md) and describes techincal details
-of how the schema should be serialized in a particular file format.
-
-**Field name**|**Constraints**
-:--|:--
-corpora_schema_version|Follows [SemVer](https://semver.org/) conventions
-corpora_encoding_version|Follows [SemVer](https://semver.org/) conventions
+cellxgene's data requirements are tailored to optimize data reuse. Because each assay has different characteristics, our requirements differ by assay type. In general,
+cellxgene requires submission of "raw" data suitable for computational reuse, when a standard form exists, and strongly suggests that a "final" matrix suitable for
```
jychien, line 50: If a contributor does not provide a "final" matrix, what does cxg use for the visualization? Does cxg create the "final" matrix from the "raw" matrix normalized using a standard process?

docs/corpora_schema.md
```diff
@@ -34,100 +35,110 @@ schema. If it does not, it will reject the dataset with an appropriate error mes
...
+    retain all barcodes in raw and final matrices, cells MUST be filtered from both. By contrast, those wishing to reuse datasets require access to raw gene expression values,
+    so genes MUST be present in both datasets. Summarizing, any cell barcodes that are removed from the data MUST be filtered from both raw and final matrices.
+    By contrast, genes MUST NOT be filtered from the raw matrix. Any genes that publishers wish to filter from the final matrix MAY have their values replaced by np.nan,
+    which will mask them from exploration.
+
+In addition to these general requirements, the following table describes the matrix data and layers requirements that are assay-specific. If cellxgene does not support
+an assay you would like to publish, please post an issue on this repository to start a conversation about extending the schema. If an entry in the table is empty,
+the cellxgene schema does not have any other requirements on data in those layers beyond the ones listed above. This is usually the case when there are many ways to
+produce the matrix layer in question.
+
```
brianraymor, line 67: I think that this table would be slightly easier to read with the following column order:

| Assay | "raw" Required? | "raw" requirements | Final ("X") Required? | Final ("X") Requirements | Other Layers |
|-------|-----------------|--------------------|-----------------------|--------------------------|--------------|

docs/corpora_schema.md
brianraymor: RE HGNC version - see https://github.com/chanzuckerberg/single-cell-curation/issues/22

brianraymor: Also see https://github.com/chanzuckerberg/single-cell-curation/blob/f54de6f710fbe7d66de76bec0249931b87e4930f/cellxgene_schema_cli/cellxgene_schema/gene_symbol.py#L183 - 

 

>  Also, seurat and scanpy append ".1" or "-1" to duplicated gene names, and these altered
>     names persist throughout the life of the object. They won't match against the HGNC database
>     and we want to merge them, so we need to strip off the suffix and try matching again. 

> This function takes a symbol and returns the symbol with the fixed case and also with the
>     seurat/scanpy suffix stripped off.

docs/corpora_schema.md
ambrosejcarr: > Does cxg create the "final" matrix from the "raw" matrix normalized using a standard process?

No - we would use the raw data that is submitted. This is why "final" is strongly recommended.

docs/corpora_schema.md
ambrosejcarr: Addressed in d04303b

ambrosejcarr:
After merging #50, #52, and #53, this PR contains the set of changes proposed for `v1.1.1` and several draft changes proposed for `v1.2.0`. 

I am going to *close* this PR and open two separate PRs for `v1.1.1` and `draft-v.1.2.0` that correspond to [this feature split](https://docs.google.com/document/d/1yCq3yJM649f6z4OmqbykZAEQk7LW9Wme9_ZPYnJoKtk/edit?usp=sharing).

PR #47, opened 2021-04-13
CITE-seq proposal

ambrosejcarr:
This PR adds a suggestion for how to store CITE-seq datasets that conforms to our [discussions](https://cziscience.slack.com/archives/C01L65J7PLP/p1614869857013900) in cziscience/single-cell-data-representation. 

In brief, AnnData are contemplating adding support for multiple assays, similar to bioconductor's [MultiAssayExperiment](https://bioconductor.org/packages/release/bioc/html/MultiAssayExperiment.html), which would be the preferred way to store these data. However, until AnnData add this support, the suggested approach is to: 

1. Create a separate namespace for protein features. I suggest `:AB` for antibody, but any affix could be used. 
2. Attach the affix to protein features. A suffix will enable users typing gene names into the explorer to discover both RNA and protein data, if they know the gene name. A prefix will enable users to search for all the protein features, if they know what the prefix is. The suffix matches established search patterns, so I suggest a suffix because I believe it will be more discoverable. This has the added benefit of making features adhere to the cellxgene requirement for unique `var` names. 
3. Store both protein and RNA features in the main matrix layers. This has some down-sides. Namely, users downloading the files will need to carefully remove the protein or RNA features before running computational analyses. However, alternatives (separate files, or protein features in `.uns`) were determined to be worse during discussion.

This PR does not make one suggested change: 
* Add a column to `.var` which declares the `feature_type`, e.g. `RNA` or `protein`. This information is not required at this time, as it is redundant to the `:AB` suffix. 

This representation requires updates to `cellxgene-schema validate`. 
* Features MUST be HGNC symbols OR contain an `:AB` suffix. 

This representation requires no changes to the data portal or cellxgene explorer infrastructure or UX, and will enable us to accumulate data while learning about how to best support it. 

**Reviewer Note:** The only substantive change created by this PR was to add a CITE-seq line to the assay table. However, this changes the white space padding in other table entries, and so the full table appears modified in the diff.


ambrosejcarr:
Some review has occurred in [single-cell-data-representation](https://cziscience.slack.com/archives/C01L65J7PLP/p1618279287040500). Changes were not requested.

pablo-gar (commented):
This looks great to me.

ambrosejcarr:
Merging branches into the `ambrosecarr/schema-update` branch prior to revision and splitting according to our [discussion of versioning](https://docs.google.com/document/d/1yCq3yJM649f6z4OmqbykZAEQk7LW9Wme9_ZPYnJoKtk/edit).

PR #53, opened 2021-04-22
RFC: Add mandatory batch field to schema

ambrosejcarr:
This PR describes a proposed schema change and encoding implementation of adding "batch" information to the schema. 

In brief, it is designed to enable a user to specify an arbitrary set of cell metadata fields that jointly describe the batch of each cell. These fields may or may not be part of the cellxgene schema. 

It addresses #2 .

### DIFF
--- a/docs/corpora_schema.md
+++ b/docs/corpora_schema.md
@@ -4,7 +4,7 @@ Contact: acarr@chanzuckerberg.com
 
 Document Status: _Approved_
 
-Version: 1.1.0
+Version: 1.2.0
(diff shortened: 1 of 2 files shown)


ambrosejcarr:
Merging branches into the `ambrosecarr/schema-update` branch prior to revision and splitting according to our [discussion of versioning](https://docs.google.com/document/d/1yCq3yJM649f6z4OmqbykZAEQk7LW9Wme9_ZPYnJoKtk/edit).

PR #62, opened 2021-05-28
curation tutorial review edits

ambrosejcarr:
I was able to successfully follow the tutorial :+1: 

This PR contains primarily: 
* Editorial suggestions
* Spelling fixes
* Changes to code that ensure the python and bash code runs successfully. 
* Added language syntax to code blocks

I thought it was easier just to make the fixes that I felt would be non-controversial than to make suggestions on your PR. 

Some higher level requested changes not included in this PR: 
* please run a spelling checker on the markdown file. I use pycharm to do this, but there are many options
* Appendix 2 mostly duplicates the schema.md and anndata_encoding files. I suggest removing it. 
* Appendix 1 has some out of date references to other sections that are no longer in order. Please check the text in the appendix. 

Questions: 
* Do you want to have the user create a folder somewhere, cd into that folder, and download files into it? Right now they're going to just dump everything into their working directory. If so, the curl links could be updated to use the -o parameter
* Some of the comment text in the code blocks causes the blocks to scroll horizontally. I find that annoying. Can it be avoided by moving the comments to new lines?
* For the example yaml, i suggest creating a file that doesn't have all the values filled to encourage the user to follow along, then provide a link to a "correct" yaml later in the tutorial. right now they are given the correct file up front.
* Appendix 1 is at the end and has some very useful information - I don't see the value of hiding it under the collapsed heading (particularly if removing appendix 2). Do you think it would make sense to unfurl it?

### DIFF
--- a/docs/curation-proto-tut.md
+++ b/docs/curation-proto-tut.md
@@ -4,16 +4,21 @@
 
 <!--- ## Overview --->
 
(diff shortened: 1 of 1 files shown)


ambrosejcarr:
Stale, edits a deprecated document.

PR #103, opened 2021-07-22
introduce `X_approximate_distribution`.

ambrosejcarr:
Introduces the optional `approximate_distribution` field to `uns`. 

Assumptions made by inserting this field in `uns` instead of `obs`: 
* Data in X are homogeneously distributed. This assumption currently makes sense since X holds a single data type. The data formats created by major toolchains recommend that different data types be stored in different matrix layers, so this will likely be the correct solution moving forward.

Assumptions made by creating a `str` typed field instead of `dict`: 
* cellxgene computes on only one matrix layer, and that data type is in X. If cellxgene expands to multi-modal data, we will need to change this field to enable users to describe each data modality. However a, `str` typed field is currently the correct choice. 

See https://github.com/chanzuckerberg/cellxgene/pull/2318 for application requirements that motivate the addition of this field.

### DIFF
--- a/schema/2.0.0/corpora_schema.md
+++ b/schema/2.0.0/corpora_schema.md
@@ -794,12 +794,12 @@ See also `default_embedding` in `uns`.
 </tbody></table>
 <br>
 
-### X_approx_distribution
+### approximate_distribution
 
 <table><tbody>
     <tr>
       <th>Key</th>
-      <td>approx_distribution</td>
+      <td>approximate_distribution</td>
     </tr>
     <tr>
       <th>Annotator</th>
@@ -851,6 +851,7 @@ schema v2.0.0 substantially *remodeled* schema v1.1.0:
 
 * uns
   * Added `batch_condition`
+  * Added `approximate_distribution`
   * Replaced `layer_descriptions` with `normalization`
   * Replaced `version` which included `corpora_schema_version` and `corpora_encoding_version` with `schema_version`
   * Deprecated `tags` and `default_field` presentation metadata
--- a/schema/2.0.0/corpora_schema.md
+++ b/schema/2.0.0/corpora_schema.md
@@ -37,11 +37,11 @@ This document is organized by:
 
 * **AnnData** - The canonical data format for the cellxgene Data Portal is HDF5-backed [AnnData](https://anndata.readthedocs.io/en/latest) as written by version 0.7 of the anndata library.  Part of the rationale for selecting this format is to allow cellxgene to access both the data and metadata within a single file. The schema requirements and definitions for the AnnData `X`, `obs`, `var`, `obsm`, and `uns` attributes are described below.
 
-* **Organisms**. Data MUST be either Human or Mouse data. No other organisms are accepted by the cellxgene Data Portal. 
+* **Organisms**. Data MUST be either Human or Mouse data. No other organisms are accepted by the cellxgene Data Portal.
 
 * **Reserved Names**. The names of the metadata keys specified by the cellxgene schema are reserved and MUST be unique. For example, duplicate `"feature_biotype"` keys in AnnData `var` are not allowed.
 
-* **Redundant Metadata**. It is STRONGLY RECOMMENDED to avoid multiple metadata fields containing identical or similar information. 
+* **Redundant Metadata**. It is STRONGLY RECOMMENDED to avoid multiple metadata fields containing identical or similar information.
 
 *   **No PII**. Curators agree to this requirement as part of the data submission policy.
     However, it is not strictly enforced in our validation tooling because it is difficult for software to predict what is and is not PII.
@@ -83,7 +83,7 @@ cellxgene requires ontology terms to enable search, comparison, and integration 
 Ontology terms for cell metadata MUST use [OBO-format identifiers](http://www.obofoundry.org/id-policy.html), meaning a CURIE (prefixed identifier) of the form **Ontology:Identifier**.
 For example, [EFO:0000001](https://www.ebi.ac.uk/ols/ontologies/efo/terms?short_form=EFO_0000001) is a term in the Experimental Factor Ontology (EFO).
 
-When no appropriate ontology value is available, then the most accurate term MUST be used. 
+When no appropriate ontology value is available, then the most accurate term MUST be used.
 
 For example if `cell_type_ontology_term_id` describes a relay interneuron, but the most specific available term in the CL ontology is [CL:0000099](https://www.ebi.ac.uk/ols/ontologies/cl/terms?obo_id=CL:0000099) for *interneuron*, then the interneuron term can be used to fulfill this requirement and ensures that users searching for "neuron" are able to find these data. If no appropriate high-level term can be found or the cell type is unknown, then the most accurate term is [CL:0000003](https://www.ebi.ac.uk/ols/ontologies/cl/terms?obo_id=CL:0000003) for *native cell*.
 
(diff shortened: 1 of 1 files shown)


bkmartinjr (changes requested):
Two suggestions:
* this does not indicate if the field is optional or required, and if optional, what semantics are associated with the missing value.   I recommend specifically noting it is optional (ie, this is a minor, backward compat change), and that no default value is implied by omitting the field.
* Did you intend to change the name to `approx_distribution`?  If so, also need to change the section title.  However, I am wondering if this is the _only_ distribution we will ever want to specify or add to the schema?  I (slightly) prefer the more explicit name, ie, X_...

brianraymor:
> this does not indicate if the field is optional 

It's part of the section that starts with:

​Curators MAY also annotate the following optional keys and values in uns. If the key is present, then its value MUST NOT be empty.​

brianraymor:
Remember to update the changelog.

bkmartinjr:
Thank you @brianraymor!   

So the only nit is clarification on the name, with the note that we _just_ added all the code to the desktop explorer using the more explicit name.  We can of course change that, but I'd want to do it before we ship, as consistency is material in this case.   I"m approving contingent upon that.

@ambrosejcarr - I'll approve contingent on resolving this.  Would be helpful to have that finalized in the next few days.

bkmartinjr (approved):
approved contingent on resolving the editorial nit around naming, AND that we resolve the naming conflict between this schema and the existing Explorer use of this field.

brianraymor:
> I (slightly) prefer the more explicit name, ie, X_...

I provided feedback to Ambrose during a call. Based on the table in [X](https://github.com/chanzuckerberg/single-cell-curation/blob/827cf9b940e3eb44a0c29ce110b4538acb83b8f6/schema/2.0.0/corpora_schema.md#x-matrix-layers), `X` may be "raw" or "final". I suggested dropping the "X_ prefix and outlining the relationship between X values and this field.

Also, I agree with your earlier point that the schema should document that no default value is assumed. (Otherwise, you wouldn't need that default to be available as a value).

bkmartinjr:
I think we may be mixing up our AnnData (H5AD) terminology and making this more confusing than it needs to be.

AnnData has two fields:    `X` and `raw.X`:
* `X` - the data matrix ([docs](https://anndata.readthedocs.io/en/latest/anndata.AnnData.X.html#anndata.AnnData.X))
* `raw.X` - the "raw" version of `X` ([docs](https://anndata.readthedocs.io/en/latest/anndata.AnnData.raw.html#anndata.AnnData.raw))

There is no specific presumption that either of these have a given distribution, just that one may be processed in some manner (sometimes filtering, sometimes more).  They definitely can not be assumed to have the same distribution in cases where the processing changes that distribution.

In the future, if we wanted to capture the distribution for `raw.X`, or some other continuous (meta) data, the name would be ambiguous without the `X_`.

ambrosejcarr:
@bkmartinjr The reason I elected to drop `X_` was: 
> **cellxgene computes on only one matrix layer, and that data type is in X.** If cellxgene expands to multi-modal data, we will need to change this field to enable users to describe each data modality. However a, str typed field is currently the correct choice.

If, as you say, we decide to capture information about an additional matrix layer (raw, or another modality, e.g. `X.layers["atac"]`), we will need to change the schema to be more specific. There are at least two ways this could be done: by adding `X_` and `atac_` or by changing `str` fields to `dict`s. Currently, other fields also implicitly refer to X, such as `batch_condition`, `normalization`, and `default_embedding` so `approximate_distribution` felt consistent with our current principles and application capabilities. I was planning to add specificity when there is a driving use case. 

> Also, I agree with your earlier point that the schema should document that no default value is assumed. (Otherwise, you wouldn't need that default to be available as a value).

@brianraymor could you clarify what change you're suggesting? I'm not confident I understand clearly enough to write an update that would address this request. 

Note also that I've adjusted the name to `approximate_distribution`, I hadn't captured Brian's feedback fully on the call. Eliminating abbreviations when possible is a good principle. Unless we decide the above logic re `X_` is incorrect, I don't expect to change the name again.

bkmartinjr:
The assumption: 
> cellxgene computes on only one matrix layer

is obvious to us, but may not be obvious to all data submitters or cellxgene users.  I think the explicit name makes the semantics far clearer, at the expense of a few extra keystrokes.

brianraymor:
I was referencing Bruce's earlier comments:

> what semantics are associated with the missing value ... no default value is implied by omitting the field.

If the curator does not explicitly provide a value, then our implementation does not assume an implicit default value that would be overridden by the curator's choice.

ambrosejcarr:
**Summary:** I agree with @bkmartinjr  that metadata that refer explicitly to a matrix layer should be annotated to capture that information and propose `X_normalization` and `X_approximate_distribution`. 

#### Thinking out loud a bit about the implementation:

I do not currently foresee wanting to capture information about `.raw.X` but we are planning to add protein data to the schema, likely in Q1 2022. 

It is highly probable that protein data will be added as an additional layer named either after the modality (`.layers["protein"]`) or the assay type (`.layers["cite-seq"]`), since separate layers for modalities is the recommendation of the R and Python toolchain ecosystems. This is also the most feasible option for RNA/ATAC multimodal datasets, which are likely coming soon (currently commercially available from 10x). 

#### Implementation possibilities for `normalization` and `approximate_distribution`:

1. Optimize for the current state. Specify the layer in the key name: `.uns["X_normalization"] = Str` and `.uns["X_approximate_distribution"] = Str`
2. Optimize for the future state. Specify the layer as a dictionary key: `.uns["normalization"] = Dict[Enum, Str]` and `.uns["approximate_distribution"] = Dict[Enum, Enum]`

The normalization implementation proposed above is probably correct for the future state (option 2). However, because we don't understand if or how we will enable computation on these future data layers, and because it's confusing to have a dict with only one valid key, I don't think optimization for the future state is viable at this time, so this PR proposes changes to implement option 1.

jahilton (approved):

jychien:
It would be helpful to know more about the context of how this field is used so that it can help inform us when this field needs to be wrangled. For the more "basic" datasets, we anticipate getting either the LogNormalized or scTransform matrix.  Based on this thread, though, I am anticipating that for these datasets, the `X_approximate_distribution` field will be omitted.

Is there any circumstance where you see that this field _should_ be filled out, and that omitting this field would result in false assumptions in subsequence cellxgene computations?

bkmartinjr:
> Is there any circumstance where you see that this field _should_ be filled out, and that omitting this field would result in false assumptions in subsequent cellxgene computations?

For the soon-to-ship Explorer, the minimalist answer is "anytime the explorer heuristic determination of the distribution" fails.  Because that is impractical to validate on a per-dataset basis, I suggest setting it whenever you have scRNAseq data with raw counts in the X matrix.  I think you can safely omit it if the data is known to be normally distributed (eg, logged counts).

For other assays, I think the same judgement applies - if Gaussian, OK to omit.  Otherwise, define it.

@ambrosejcarr - curious if you think this approach is sufficiently future-proof.  It does somewhat tie our hands to "normal" being the default fall-back in the future.

ambrosejcarr:
> I think you can safely omit it if the data is known to be normally distributed (eg, logged counts).

This is the approach I would suggest. 

> curious if you think this approach is sufficiently future-proof. It does somewhat tie our hands to "normal" being the default fall-back in the future.

TL;DR: No, but I think it's the right path right now. 

We STRONGLY RECOMMEND "normalized" data matrices. In the majority of cases, data have been transformed to have the same scale (e.g. divide by total counts) and to be approximately normally distributed (log transformation of a count distribution). 

_Some_ approaches (e.g. DESeq2) that model count distributions only transform the data to have equal scale. however they're not commonly used for sc-RNAseq. The proposed approach should usually be sufficient. I do not see a more efficient way to do better on "normalized" matrices than a schema field + curator, since there is significant variation in the transformations that are executed.

If, in the future, we want to be more accurate, we will need to adopt a modeling framework like scVI-tools that models the raw count data. This is venturing into a challenging area that is causing the DCP significant problems.

ambrosejcarr:
@brianraymor gave verbal agreement to merge this PR, so I have approval from all reviewers.

PR #131, opened 2021-09-01
Update requirements for tissue_ontology_term_id

brianraymor:
Per this [comment thread](https://github.com/chanzuckerberg/single-cell-curation/pull/116#discussion_r699680657), clarifying the requirements for `tissue_ontology_term_id` to ensure that validation can correctly enforce.

### DIFF
--- a/schema/2.0.0/corpora_schema.md
+++ b/schema/2.0.0/corpora_schema.md
@@ -698,7 +698,7 @@ When a dataset is uploaded, cellxgene Data Portal MUST automatically add the mat
     </tr>
     <tr>
       <th>Value</th>
-        <td><code>str</code>. If the <code>feature_biotype</code> is <code>"gene"</code> then this MUST be the human-readable ENSEMBL gene name assigned to the <code>feature_id</code>. If the <code>feature_biotype</code> is <code>"spike-in"</code> then this MUST be the ERCC Spike-In identifier appended with <code>" spike-in control"</code>.
+        <td><code>str</code>. If the <code>feature_biotype</code> is <code>"gene"</code> then this MUST be the human-readable ENSEMBL gene name assigned to the <code>feature_id</code>. If the <code>feature_biotype</code> is <code>"spike-in"</code> then this MUST be the ERCC Spike-In identifier appended with <code>" (spike-in control)"</code>.
         </td>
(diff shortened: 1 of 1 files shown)


schema/2.0.0/corpora_schema.md
```diff
@@ -441,7 +441,7 @@ Curators MUST annotate the following columns in the `obs` dataframe:
           <tbody>
             <tr>
               <td>Tissue</td>
-              <td>MUST be an UBERON term<br>(e.g. <a href="http://purl.obolibrary.org/obo/UBERON_0008930"><code>"UBERON:0008930"</code></a> for a <i>sematosensory cortex</i> tissue sample)</td>
+              <td>STRONGLY RECOMMENDED to be an UBERON term<br>(e.g. <a href="http://purl.obolibrary.org/obo/UBERON_0008930"><code>"UBERON:0008930"</code></a> for a <i>sematosensory cortex</i> tissue sample)</td>
```
jahilton, line 444: I'm lacking the context for this change. Why is it no longer a MUST?

brianraymor: Per the linked comment thread, we now support both UBERON and CL and cannot programmatically discern between _Tissue_ and  _Enriched, Sorted, or Isolated Cells from a Tissue_ to enforce (MUST) UBERON for Tissue.

pablo-gar (approved):
LGTM

schema/2.0.0/corpora_schema.md
jahilton: I see (sorry!).
Kicking around another option in my head (spoiler alert: I vote 'nay' on it) - add an appendage to the bottom case of `Enriched, Sorted,or Isolated Cells from a Tissue`. You could then enforce that the lack of an appendage MUST be an UBERON term.
This could also help flag for the user that data from the CD3+ kidney cells examples should not be taken in the same light as unenriched kidney tissue data (currently, each would be tissue:kidney).

However, the tissue field is one where we do not see much issue with. The proposed changes still allow for a submitter to choose the most appropriate term, whether it be at a cellular (CL) or multi-cellular (UBERON) level

jahilton (approved):

PR #132, opened 2021-09-07
Fix markdown formatting issues

rcannood:


ambrosejcarr:
Hi Robrecht, thanks for your interest in this repository. We're deprecating this version of the encoding, and the way this PR was structured doesn't provide a durable solution to the problem (whitespace in markdown). As a result, we're going to close this PR. 

We welcome conversations via issues about potential contributions - it's possible that at some point in the future we'd consider setting up automatic linting, which would provide a more comprehensive fix.

PR #141, opened 2021-09-13
Create workflow for unittests of cellxgene-schema

pablo-gar:
This PR adds a "python application" workflow that will automate running lint and unittests on `cellxgene-schema`

### DIFF
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -1,0 +1,42 @@
+# This workflow will install Python dependencies, run tests and lint with a single version of Python
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
+
+name: Python application
+
+on:
+  push:
+    branches:
+      - main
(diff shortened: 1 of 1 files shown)


.github/workflows/python-app.yml
```diff
@@ -0,0 +1,42 @@
+# This workflow will install Python dependencies, run tests and lint with a single version of Python
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
+
+name: Python application
+
+on:
```
atolopko-czi, line 6: should we follow [this pattern](https://github.com/chanzuckerberg/single-cell-explorer/blob/main/.github/workflows/push_tests.yml#L3-L7)?

atolopko-czi (approved):
Thanks for adding this! One comment for your consideration, but ok to merge otherwise.

.github/workflows/python-app.yml
pablo-gar: Good idea! Thanks

PR #153, opened 2021-10-06
Dunitz/experimental

MDunitz:
Updates to the single-cell-curation package to use logger instead of print. 
Some refactoring.
Small change to the validator function to return the list of errors so the function can be called directly in the batch processing pipeline to allow for easier debugging
See corresponding changes in the [data portal processing pipeline](https://github.com/chanzuckerberg/single-cell-data-portal/pull/1493)


cellxgene_schema_cli/setup.py
```diff
@@ -5,7 +5,7 @@
 
 setup(
     name="cellxgene-schema",
-    version="2.0.1",
+    version="2.0.0",
```
atolopko-czi, line 8: oh, interesting, was this already incremented on `main`?

cellxgene_schema_cli/cellxgene_schema/cli.py
```diff
@@ -34,7 +34,8 @@ def schema_cli():
     type=click.Path(exists=False, dir_okay=False, writable=True),
 )
 def schema_validate(h5ad_file, add_labels_file):
-    if validate.validate(h5ad_file, add_labels_file):
+    is_valid, errors = validate(h5ad_file, add_labels_file)
```
atolopko-czi, line 37: nit: 
```suggestion
    is_valid, _ = validate(h5ad_file, add_labels_file)
```

cellxgene_schema_cli/cellxgene_schema/__init__.py
```diff
@@ -0,0 +1,33 @@
...
+    if not validator.is_valid:
+        return False, validator.errors
+
+    if add_labels_file:
+        writer = AnnDataLabelAppender(validator)
+        writer.write_labels(add_labels_file)
+
+        return validator.is_valid & writer.was_writing_successful, validator.errors
+
+    return True
```
atolopko-czi, line 33: return `True, []` ?

cellxgene_schema_cli/cellxgene_schema/cli.py
atolopko-czi: Though, the code for printing errors to stdout might be might better moved here, since it's CLI-specific. That would leave the underlying validate() call to just generate and maintain the errors for use by the caller.  Now that we have a non-CLI usage, that seems reasonable. Pretty minor!

MDunitz: We might want to run that past @pablo-gar I'm not sure if he prefers/expects the errors to print in a particular order

pablo-gar (changes requested):
We need to update version number and the chagelog as well

cellxgene_schema_cli/tests/test_schema_compliance.py
```diff
@@ -1,8 +1,12 @@
...
-import pandas as pd
+
 import numpy
-from cellxgene_schema import validate
-import fixtures.examples_validate as examples
+import pandas as pd
+from cellxgene_schema.validate import Validator
+
+from cellxgene_schema.write_labels import AnnDataLabelAppender
+from fixtures.examples_validate import adata, h5ad_valid, h5ad_invalid, adata_non_raw, adata_with_labels
```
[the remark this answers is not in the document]
MDunitz: ohh thats fair. I thought that this made it more clear where they were coming from/is a format that I'm more comfortable with. Also this type of import just makes it easier to use the code editor to do refactors. Ime pycharm gets confused by those types of imports and screws it up when moving code to new files. 

That said I can revert the change if you prefer

cellxgene_schema_cli/cellxgene_schema/__init__.py
```diff
@@ -0,0 +1,33 @@
+import os
```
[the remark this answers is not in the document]
atolopko-czi: Yeah, also not entirely sure about this practice. It seems code that goes in __init__.py is usually related to package loading behavior. So maybe move this code to a normal py file, though should work as is.

> In the simplest case, __init__.py can just be an empty file, but it can also execute initialization code for the package or set the __all__ variable, described later

[https://docs.python.org/3/tutorial/modules.html#packages]

@blrnw3 may have stronger opinions

cellxgene_schema_cli/setup.py
atolopko-czi: Oh right, could use test pypi for testing this with the DP repo changes!  The official process for releasing is not buttoned down (had [started on that](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-curation/134), but never finished. In any case, following [similar process to the cellxgene desktop](https://github.com/chanzuckerberg/cellxgene/blob/main/dev_docs/release_process.md#releasing-a-major-or-minor-version-of-cellxgene) is the goal.  For test.pypi.org release, the important thing is to use release candidate version numbers, which have `-rc.#` suffixes. If 2.0.1 is not released to official pypi.org yet, I would use `2.0.1-rc.1`, and if 2.01 is officially release, I would bump to `2.0.2-rc.1`.

cellxgene_schema_cli/cellxgene_schema/cli.py
pablo-gar: Doesn't it help to print the errors even if not called from the CLI? Specially now that during upload the CLI is not used.

cellxgene_schema_cli/cellxgene_schema/__init__.py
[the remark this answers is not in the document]
pablo-gar: I prefer just adding it as a separate method in `validate.py`. Sfaria is using these libraries and I'm afraid we will break with the refactoring

cellxgene_schema_cli/CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to the python package `cellxgene-schema` are documented in t
 The format of this changelog is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [2.0.3] - 2021-10-08
+### Changed
+- Replace print with logging
+- Return the list of errors from validate so they can be used by callers
```
pablo-gar, line 10: - Replace print with logging in `Validator` and `AnnDataLabelAppender`.
- Return the list of errors from `validate` so they can be used by callers.
- Add verbose option to CLI, it prints a progress log.

PR #154, opened 2021-10-06
Fix performance issue of raw layer validation

pablo-gar:
This is a performance fix for when the validator does an integer check on the expression matrix. With large datasets the validator would take hours/days running without finishing the check. This is because a copy of all non-zero values from the expression matrix was made.

The issue has been fixed and only a copy of the first 5000 values is made (an arbitrarily chosen number).

### DIFF
--- a/cellxgene_schema_cli/cellxgene_schema/validate.py
+++ b/cellxgene_schema_cli/cellxgene_schema/validate.py
@@ -935,7 +935,10 @@ class Validator:
             # Get array without zeros
             non_zeroes_index = x.nonzero()
 
-            x_non_zeroes = x[non_zeroes_index]
+            if max_values_to_check > len(non_zeroes_index[0]):
+                max_values_to_check = len(non_zeroes_index[0])
+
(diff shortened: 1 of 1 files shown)


MDunitz (approved):
lgtm

cellxgene_schema_cli/CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ All notable changes to the python package `cellxgene-schema` are documented in t
 The format of this changelog is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [2.0.1] - 2021-10-06
+
+### Changed
+
+- A performance fix was implemented. Improves the performance of `Validator._is_raw()` in `validate.py`, and addresses timeouts when validating datasets with high number of non-zero values in the expression matrix.
```
brianraymor, line 11: ~A performance fix was implemented.~ Improves the performance of `Validator._is_raw()` in `validate.py` and addresses timeouts when validating datasets with a high number of non-zero values in the expression matrix.

PR #157, opened 2021-10-07
Expand list of assays for which raw validation is bypassed

pablo-gar:
Adds EFO:0008939 to list of assays for which raw validation is bypassed in `cellxgene-schema validate`. Changelog was modified accordingly and version updated to v2.0.2


MDunitz (approved):
lgtm

PR #165, opened 2021-11-01
Add warning during raw data validation for when raw.X has non-integer values

pablo-gar:
The validator was missing one check when validating the raw data:

- When both `adata.X` and `adata.raw.X` are present but `adata.raw.X` contains non-integer values, then a warning should be shown suggesting that data in `adata.raw.X` may not be raw.

A warning is preferred as opposed to an error as @jychien states:

> For the Pott gut demo dataset, they actually used STARsolo for their pipeline. I am not as familiar with data from STARsolo, and it looks like it handles multi-gene reads in a way different from cellranger, and possibly could result in non-integers depending on the parameter used (documentation). I think a warning would be good, in case we need to be more flexible and don't want to have to change the validator for different pipelines or assays. But I think we've come across enough matrices where raw is not really raw that a warning would be helpful.

### DIFF
--- a/cellxgene_schema_cli/cellxgene_schema/validate.py
+++ b/cellxgene_schema_cli/cellxgene_schema/validate.py
@@ -1030,6 +1030,15 @@ class Validator:
 
             normalization = self.adata.uns["X_normalization"]
 
+            # If both "raw.X" and "X" exist but neither are raw
+            if (
+                not self._is_raw()
+                and self._get_raw_x_loc() == "raw.X"
+            ):
+                self.warnings.append(
+                    "Raw data may be missing: data in 'raw.X' contains non-integer values."
+                )
+
             # If raw data is missing: no "raw.X", and "X_normalization" is not "none"
             if (
                 not self._is_raw()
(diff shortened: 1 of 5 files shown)


cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -1030,6 +1030,15 @@ def _validate_raw(self):
 
             normalization = self.adata.uns["X_normalization"]
 
+            # If both "raw.X" and "X" exist but neither are raw
+            if (
+                not self._is_raw()
+                and self._get_raw_x_loc() == "raw.X"
+            ):
+                self.warnings.append(
```
seve, line 1038: Should this be a warning or an error

seve (commented):
~One Q~ nvm!

cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -1030,6 +1030,15 @@ def _validate_raw(self):
 
             normalization = self.adata.uns["X_normalization"]
 
+            # If both "raw.X" and "X" exist but neither are raw
+            if (
```
metakuni, line 1034: Could we add a unit test for this?

metakuni (commented):
I'm new to this repo, so pardon the naive questions. 😄 

Should we add a unit test for the new warning?

cellxgene_schema_cli/cellxgene_schema/validate.py
pablo-gar: yes I will do that

pablo-gar: @metakuni Done =D

metakuni (approved):

PR #167, opened 2021-11-02
Change feature_id to index

### LINKED ISSUE #166: Remove feature_id from schema
The schema indicates that var.index & raw.var.index should be named `feature_id` (https://github.com/chanzuckerberg/single-cell-curation/blob/main/schema/2.0.0/schema.md#feature_id)
But this is not validated and not an actual requirement so should be removed from the schema.

brianraymor:
Closes #166

Do we also need more nuanced text related to when `var.index` and `var.raw.index` are present or not? For example, `var.raw.index` is NOT REQUIRED to be present in ATAC-seq datasets like _GABAergic neurons — An Atlas of Gene Regulatory Elements in Adult Mouse Cerebrum_.

### DIFF
--- a/schema/2.0.0/schema.md
+++ b/schema/2.0.0/schema.md
@@ -38,6 +38,8 @@ This document is organized by:
 
 * **AnnData** - The canonical data format for the cellxgene Data Portal is HDF5-backed [AnnData](https://anndata.readthedocs.io/en/latest) as written by version 0.7 of the anndata library.  Part of the rationale for selecting this format is to allow cellxgene to access both the data and metadata within a single file. The schema requirements and definitions for the AnnData `X`, `obs`, `var`, `raw.var`, `obsm`, and `uns` attributes are described below.
 
+  All data submitted to the cellxgene Data Portal is automatically converted to a Seurat V3 object that can be loaded by the R package Seurat. See the [Seurat encoding](https://github.com/chanzuckerberg/single-cell-curation/blob/main/schema/2.0.0/seurat_encoding.md) for further information.
+
 * **Organisms**. Data MUST be from a Metazoan organism or SARS-COV-2 and defined in the NCBI organismal classification. For data that is neither Human, Mouse, nor SARS-COV-2, features MUST be translated into orthologous genes from the pinned Human and Mouse gene annotations.
 
 * **Reserved Names**. The names of the metadata keys specified by the schema are reserved and MUST be unique. For example, duplicate `"feature_biotype"` keys in AnnData `var` are not allowed.
--- a/schema/2.0.0/schema.md
+++ b/schema/2.0.0/schema.md
@@ -620,7 +620,7 @@ When a dataset is uploaded, the cellxgene Data Portal MUST automatically add the
 
 `var` and `raw.var` are both of type [`pandas.DataFrame`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.html).
 
-Curators MUST annotate the following columns in the `var` and `raw.var` dataframes:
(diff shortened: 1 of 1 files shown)


schema/2.0.0/schema.md
```diff
@@ -643,20 +641,20 @@ Curators MUST annotate the following columns in the `var` and `raw.var` datafram
...
+      <td>index</td>
     </tr>
     <tr>
       <th>Annotator</th>
       <td>Curator</td>
     </tr>
     <tr>
       <th>Value</th>
-        <td><code>str</code>. If the <code>feature_biotype</code> is <code>"gene"</code> then this MUST be an ENSEMBL term. If the <code>feature_biotype</code> is <code>"spike-in"</code> then this MUST be an ERCC Spike-In identifier. </td>
+        <td><code>str</code>. If the <code>feature_biotype</code> is <code>"gene"</code> then this MUST be an ENSEMBL term. If the <code>feature_biotype</code> is <code>"spike-in"</code> then this MUST be an ERCC Spike-In identifier.<br><br> <code>var.index</code> MUST contain unique identifiers for features. <code>raw.var.index</code> MUST be identical to <code>var.index</code>.<br><br></td>
```
jahilton, line 657: var.index MUST contain unique identifiers for features. _If AnnData.raw is present_, raw.var.index MUST be identical to var.index (?)

jahilton:
>Do we also need more nuanced text related to when var.index and var.raw.index are present or not?

Do you feel like the per Assay table addressing AnnData.raw.X & AnnData.X is insufficient? Or are you thinking that the var dataframes aren't directly referenced in that table?

schema/2.0.0/schema.md
jahilton: ...or maybe just one note at the intro to var and raw.var (Gene Metadata)...
"The requirements only apply to raw.var if AnnData.raw is present"

schema/2.0.0/schema.md
```diff
@@ -643,20 +641,20 @@ Curators MUST annotate the following columns in the `var` and `raw.var` datafram
...
 <br>
 
-### feature_id
+### index
 
 <table><tbody>
     <tr>
       <th>Key</th>
-      <td>feature_id (<code>var.index</code>/<code>raw.var.index</code>)</td>
+      <td>index</td>
```
jahilton, line 649: I'm afraid there is still an implied requirement that the index must be named `index`. It might be worth including a note
index (no specific name is required)

brianraymor: I meant `index` in the sense of `var.index` and `raw.var.index` which are the keys even if `adata.var.index.name` points to `feature_id`.

Do you feel it would be more precise as *the pandas.DataFrame index for `var` or `raw.var`*

brianraymor:
> Do you feel like the per Assay table addressing AnnData.raw.X & AnnData.X is insufficient? Or are you thinking that the var dataframes aren't directly referenced in that table?

I feel that it is sufficient for close readers. But there is a subtlety of MUST be set IF present that's implicit.

schema/2.0.0/schema.md
brianraymor: I think that I prefer the second note.

pablo-gar:
LGTM

pablo-gar (approved):

schema/2.0.0/schema.md
jahilton: Yes, I think that would help distinguish it from the other schema properties.
_I_ understood what you meant, but I'm worried that some of the R folks might see all of the key:field_name and take them to mean column name, then treat this one the same.

brianraymor: Speaking of R, we should also link the R Encoding document into the schema.

jahilton (approved):

PR #177, opened 2021-11-16
Fix shell error in Makefile

danieljhegeman:
The create-release-candidate make target was making use of the 'shell'
functionality in the Makefile language; the expansion for this
functionality was resulting in evaluation of the shell script
immediately at the beginning of the execution of the target's recipe,
which produced a logging bug and a bug in the naming of the rc branch

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ pydist: clean
 	cd cellxgene_schema_cli && python setup.py sdist bdist_wheel
 
 # For displaying the current version
-current_version := $(shell awk '/current_version =/ { print $$3 }' .bumpversion.cfg)
+current_version = $$(awk '/current_version =/ { print $$3 }' .bumpversion.cfg)
 
 # Show the current version
 show-current-version:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -30,7 +30,7 @@ show-current-version:
 # Set PART={major,minor,patch} as param to make the version bump happen automatically.
 create-release-candidate:
 	bumpversion --config-file .bumpversion.cfg $(PART) --allow-dirty
-	git checkout -b release-version-$(shell awk '/current_version =/ { split($$3,v,"-"); print v[1] }' .bumpversion.cfg)
+	git checkout -b release-version-$$(awk '/current_version =/ { split($$3,v,"-"); print v[1] }' .bumpversion.cfg)
```
danieljhegeman, line 33: the command passed to `shell` was evaluating *before* the execution of ln 32

metakuni (approved):
Awesome, thanks for the fix!

PR #189, opened 2022-01-21
Add ontology ancestor mapping notebook

ebezzi:
Notebook used to compute ancestors for nodes in human, mouse and UBERON (only a subset) ontologies.

### DIFF
--- a/notebooks/compute_ancestor_mapping.ipynb
+++ b/notebooks/compute_ancestor_mapping.ipynb
@@ -1,8 +1,24 @@
 {
  "cells": [
   {
+   "cell_type": "markdown",
+   "id": "89ea87d1-2459-4c09-b482-786db8cd3031",
+   "metadata": {},
+   "source": [
+    "# Ontology ancestors mapping\n",
+    "This notebook can be used to compute mappings between nodes in an ontology files and their ancestors.\n",
+    "It will:\n",
+    "1. Load the required ontology files (human, mouse, UBERON)\n",
+    "1. For `human` and `mouse`, start from each node available in the ontology\n",
+    "1. for `UBERON`, only start from a defined, hardcoded subset of leaf nodes which is relevant for our datasets\n",
+    "1. Recursively compute ancestors using the `is_a` relationship and `part_of` restriction (defined as `BFO_0000050`)\n",
+    "1. Create mappings from each node to its ancestors\n",
+    "1. Write a json file with these mappings (a dictionary of lists)"
+   ]
+  },
+  {
    "cell_type": "code",
-   "execution_count": 361,
+   "execution_count": null,
    "id": "309f0981-0d2b-4392-90e2-3c60bf4e23ae",
    "metadata": {},
    "outputs": [],
@@ -13,7 +29,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 362,
+   "execution_count": null,
    "id": "d9d21563-dc35-4f1d-9c95-3d1730f419e1",
    "metadata": {},
    "outputs": [],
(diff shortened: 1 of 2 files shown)


atolopko-czi (changes requested):
* Suggest mentioning this in this [doc](https://github.com/chanzuckerberg/single-cell-curation/blob/ebezzi/add-ontology-ancestor-notebook/cellxgene_schema_cli/scripts/readme.md).
* And what if we put this notebook/script somewhere under [cellxgene_schema_cli/cellxgene_schema](https://github.com/chanzuckerberg/single-cell-curation/tree/ebezzi/add-ontology-ancestor-notebook/cellxgene_schema_cli/cellxgene_schema/) to keep it closer to related files?

ebezzi:
> * Suggest mentioning this in this [doc](https://github.com/chanzuckerberg/single-cell-curation/blob/ebezzi/add-ontology-ancestor-notebook/cellxgene_schema_cli/scripts/readme.md).
> * And what if we put this notebook/script somewhere under [cellxgene_schema_cli/cellxgene_schema](https://github.com/chanzuckerberg/single-cell-curation/tree/ebezzi/add-ontology-ancestor-notebook/cellxgene_schema_cli/cellxgene_schema/) to keep it closer to related files?

Adding to the doc for sure, not that sure about adding it to `cellxgene_schema` - this is more of a static generation file and it's not quite related to the validator. @pablo-gar since you mostly work on this repo, what are your thoughts?

pablo-gar:
I agree that for the moment we should not include under `cellxgene_schema_cli/cellxgene_schema`. However if at any point we need to put extra constraints in `obs["development_stage"]` for the validator, then we need to make sure to include ancestors in the reference static ontology files.

The work Emanuale did here is because ancestors and descendants cannot be extracted as usual with owlready2, thus that info is not included in the reference files at `cellxgene_schema_cli/cellxgene_schema/ontology_files/all_ontology.json.gz`. 

I may recommend we put this under `docs` to not clutter the root of the repo, @brianraymor any suggestions?

pablo-gar (commented):
@ebezzi are you planning to use this for any sort of production stages? 

If so we have to make sure that we are pulling information from the pinned version of the ontologies as described in the schema v2.0.0. I believe your code pulls the latest version available in obo library.

pablo-gar:
@ebezzi can you also please add a brief description of what the notebook does, maybe at the beginning of the notebook

atolopko-czi (approved):
Approving now, with my suggestions having been considered. Thanks!

ebezzi:
@pablo-gar 

1. What about a `notebooks` folder?
2. Agree, if we increase the ontology constraints (or add new species) we will need to regenerate the files. I think this would be infrequent enough not to justify automating the process
3. This will be used in production. I think @brianraymor mentioned that we're not particularly concerned about having to regenerate the mappings often. When I ran the script, I downloaded the ontology files from the link you sent me and ran it from my machine, so they're the latest versions.
4. I will add a description, thanks for the suggestion!

brianraymor:
Who is the audience for this notebook? I like Pablo's suggestion to use `notebooks`. 

The ontologies must match the [pinned versions](https://github.com/chanzuckerberg/single-cell-curation/blob/main/schema/2.0.0/schema.md#required-ontologies) in schema 2.0.0. HSAPDV and MMUSDV do not change much, but UBERON does.

ebezzi:
@brianraymor The notebook is mostly meant for developers who want/need to regenerate the mapping files. Regenerating it won't be a big deal, but it should be a sporadic activity (say once per year), otherwise we might have to create some sort of automation for this.

pablo-gar (approved):

PR #204, opened 2022-03-26
Add support to generate metrics about testing coverage in schema CLI and upload to Codecov

maniarathi:
### Additions

- Add Codecov badge to the top level repo README.
- Upload coverage report to Codecov in push_test Github Action.

### Modifications

- Alphabetize `requirements.txt`
- Run `coverage` instead of `pytest` to execute unit tests so that coverage stats are generated.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ install:
 	cd cellxgene_schema_cli/ && pip install -r requirements.txt && pip install .
 
 unit-test:
-	cd cellxgene_schema_cli && pytest
+	cd cellxgene_schema_cli && coverage run --source=cellxgene_schema -m unittest discover --start-directory=tests/ --verbose && coverage xml
 
 clean:
 	rm -rf cellxgene_schema_cli/build cellxgene_schema_cli/dist cellxgene_schema_cli/cellxgene_schema.egg-info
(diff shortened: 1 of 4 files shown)


cellxgene_schema_cli/requirements.txt
```diff
@@ -1,9 +1,10 @@
-wheel==0.37.0
+anndata==0.7.6
 click==8.0.1
+coverage==6.3.2
```
seve, line 3: Only substantial change in file I believe

Makefile
```diff
@@ -9,7 +9,7 @@ install:
 	cd cellxgene_schema_cli/ && pip install -r requirements.txt && pip install .
 
 unit-test:
-	cd cellxgene_schema_cli && pytest
+	cd cellxgene_schema_cli && coverage run --source=cellxgene_schema -m unittest discover --start-directory=tests/ --verbose && coverage xml
```
seve, line 12: Do we not need a CODECOV token like we do in dataportal?

maniarathi: Nope. It is a public repository so we don't need a token.

cellxgene_schema_cli/requirements.txt
seve: was just commenting so Pablo didn't have to manually diff too!

Issue #188, opened 2022-01-12
In gene reference files, some gene names are ENSEMBL ids that are not the same as the actual gene id


pablo-gar:
Reference: [single-cell-data-representation](https://cziscience.slack.com/archives/C01L65J7PLP/p1642006704002800).

The gene reference file `cellxgene_schema_cli/cellxgene_schema/ontology_files/genes_homo_sapiens.csv.gz` contains genes whose gene name is an ENSEBML id and it is not the same as the actual gene id.

For example:
`ENSG00000276312	ENSG10010139884.1`

There are 620 of such genes.

Further investigation indicates that this discrepancy already exists in the ENCODE reference GTF file used to create `genes_homo_sapiens.csv.gz`

**Potential solutions**
1. Do nothing, maintaining consistency with ENCODE.
2. Change the gene name to the original ENSEMBL id.

jahilton:
Pulled directly from ENSEMBL - http://ftp.ensembl.org/pub/release-105/gtf/homo_sapiens/
`$  grep "ENSG00000276312" Homo_sapiens.GRCh38.105.*`

None of the results show any gene name for this gene.
`Homo_sapiens.GRCh38.105.chr_patch_hapl_scaff.gtf:KI270733.1	ensembl	gene	167218	167307	.	+	.	gene_id "ENSG00000276312"; gene_version "1"; gene_source "ensembl"; gene_biotype "misc_RNA";
Homo_sapiens.GRCh38.105.chr_patch_hapl_scaff.gtf:KI270733.1	ensembl	transcript	167218	167307	.	+	.	gene_id "ENSG00000276312"; gene_version "1"; transcript_id "ENST00000615130"; transcript_version "1"; gene_source "ensembl"; gene_biotype "misc_RNA"; transcript_source "ensembl"; transcript_biotype "misc_RNA"; tag "basic"; transcript_support_level "NA";
Homo_sapiens.GRCh38.105.chr_patch_hapl_scaff.gtf:KI270733.1	ensembl	exon	167218	167307	.	+	.	gene_id "ENSG00000276312"; gene_version "1"; transcript_id "ENST00000615130"; transcript_version "1"; exon_number "1"; gene_source "ensembl"; gene_biotype "misc_RNA"; transcript_source "ensembl"; transcript_biotype "misc_RNA"; exon_id "ENSE00003727949"; exon_version "1"; tag "basic"; transcript_support_level "NA";
[...]

PR #207, opened 2022-03-30
feat: filter PAR_Y genes

seve:
Adds filter to remove genes with id's ending in PAR_Y.

Note: Committed changes to all .csv's, should I remove non-homo sapiens?

#187

PR #208, opened 2022-03-30
feat: use gene_id as gene_name if gene_name is incorrect id

seve:
Replaces gene_name with gene_id if gene_name starts with "ENSG", but is mismatched from the actual gene_id

#188


pablo-gar (approved):
LGTM but changes in reference csv files need to be incorporated from https://github.com/chanzuckerberg/single-cell-curation/pull/207

PR #220, opened 2022-05-17
fix: schema link in readme

jahilton:
The schema link in the readme was outdated


metakuni:
Thanks for the PR, @jahilton ! I should have looked before doing the same in #222 . 😄

metakuni:
Thanks for the PR, @jahilton ! I should have looked before doing the same in #222 . 😄

PR #234, opened 2022-06-06
chore: add notebooks for POST /collections

danieljhegeman:
Functional: @Bento007 
For awareness: @brianraymor 

Add: 
- Python notebook with sample code for `POST /collections`
- R notebook with sample code for `POST /collections`


notebooks/curation_api/create_collection_dev_R.ipynb
```diff
@@ -0,0 +1,213 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "bea213f4",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "domain_name <- \"cellxgene.dev.single-cell.czi.technology\"\n",
```
Bento007, line 83: This should point to prod by default. We don't want to advertise the dev environment to the curators.

Bento007 (changes requested):
Since these are user-facing notebooks, do we want to have dev in the file names?

notebooks/curation_api/create_collection_dev_R.ipynb
Bento007: This also applies to the python note book.

Bento007: or we need to tell the users where to find the domain name they need. It's not obvious that they can use the same domain from the frontend.

danieljhegeman: agreed, will change

danieljhegeman: done

Bento007 (approved):

PR #265, opened 2022-07-03
chore: add python notebooks for remaining endpoints

danieljhegeman:
Also includes a few small improvements/mods to previously-added notebooks as well as some minor refactoring of existing code.


notebooks/curation_api/python/delete_dataset.ipynb
```diff
@@ -0,0 +1,166 @@
...
+   "source": [
+    "set_api_access_config(api_key_file_path)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "6085d322",
+   "metadata": {},
+   "source": [
+    "### Create new private Collection"
```
nayib-jose-gloria, line 131: Rename to Delete Dataset

notebooks/curation_api/python/update_collection.ipynb
```diff
@@ -0,0 +1,181 @@
...
+   "source": [
+    "set_api_access_config(api_key_file_path)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "6085d322",
+   "metadata": {},
+   "source": [
+    "### Fetch Collection"
```
nayib-jose-gloria, line 146: Rename to Update Collection

notebooks/curation_api/python/update_curator_tag.ipynb
```diff
@@ -0,0 +1,187 @@
...
+   "source": [
+    "set_api_access_config(api_key_file_path)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "6085d322",
+   "metadata": {},
+   "source": [
+    "### Create new private Collection"
```
nayib-jose-gloria, line 152: Rename to Update Curator Tag

notebooks/curation_api/python/src/dataset.py
```diff
@@ -1,18 +1,103 @@
...
+            if dataset_uuid := matches.get("dataset_uuid"):
+                identifier_type = "dataset_uuid"
+                identifier_value = dataset_uuid
+            else:
+                identifier_type = "curator_tag"
+
+    if not identifier_type:
+        raise Exception(f"The identifier '{identifier}' must either 1) include a '.h5ad' suffix OR 2) be a uuid")
+
+    return identifier_value, identifier_type
```
nayib-jose-gloria, line 46: super nit: I would prefer returning (type, value) instead to match the function name order

notebooks/curation_api/python/src/dataset.py
```diff
@@ -1,18 +1,103 @@
...
 
 logger = get_custom_logger()
 
+UUID_REGEX = r"[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}"
+DATASET_ID_REGEX = f"(?P<dataset_uuid>{UUID_REGEX})"
+CURATOR_TAG_PREFIX_REGEX = r"(?P<tag_prefix>.*)"
+EXTENSION_REGEX = r"(?P<extension>h5ad)"
+
+
+def get_identifier_type_and_value(identifier: str) -> Tuple[str, str]:
```
nayib-jose-gloria, line 26: What do you think of adding a parameter to both delete and patch dataset asking the user to input whether the current identifier they're providing is a curator_tag or dataset_uuid? Presumably, they know the difference and can save us from doing this work on the backend

nayib-jose-gloria: we can add a quick check that their input matches either of the available options, else raise an exception asking for a proper input

nayib-jose-gloria (changes requested):
Requested a couple of doc heading changes + a comment considering a refactor for the newly added dataset functions

notebooks/curation_api/python/src/dataset.py
danieljhegeman: >Presumably, they know the difference and can save us from doing this work on the backend

We can _determine_ the difference and can therefore save _them_ from the hassle on the client side 😉 

But seriously, I think it's better this way, just like when they pass us a Collection uuid we don't want to make them also tell us whether it's a private or public Collection. It's unnecessary work for them, and it'll lead to accidental errors when they change an identifier and forget to change the type that they would need to pass in as well.

brianraymor: > We can determine the difference and can therefore save them from the hassle on the client side

+1. Curator time is valuable.

notebooks/curation_api/python/src/collection.py
```diff
@@ -82,3 +82,18 @@ def get_collections(visibility: str = "PUBLIC") -> list:
...
     res.raise_for_status()
     return res.json().get("collections")
+
+
+def update_collection(collection_uuid: str, collection_form_metadata: dict) -> None:
+
+    url = url_builder(f"/collections/{collection_uuid}")
+    headers = get_headers()
+    try:
+        res = requests.put(url, headers=headers, data=json.dumps(collection_form_metadata))
```
nayib-jose-gloria, line 92: probably already on your radar, but just flagging that this is a `patch` request now with Trent's latest merged PR!

Bento007 (changes requested):

notebooks/curation_api/python/create_dataset_from_upload_link.ipynb
```diff
@@ -0,0 +1,166 @@
...
+   "source": [
+    "set_api_access_config(api_key_file_path)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "6085d322",
+   "metadata": {},
+   "source": [
+    "### Delete Dataset"
```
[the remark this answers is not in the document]
danieljhegeman: Ah, missed another change there. Good catch!

danieljhegeman:
@Bento007 do you have any further comments on this or can I merge

Bento007 (approved):

PR #275, opened 2022-07-15
Added further recommendations for donor_id design

### LINKED ISSUE #164: Add donor id field to obs
The schema does not currently record biosample_id (also known as donor_id). Should we add support for this concept in the schema?

### Rationale

A blood dataset is submitted. That dataset comprises blood contributed by Ambrose and Jason. 

_**Story: A Data consumer would like to select blood cells that came from Jason.**_ 

_**Stor 2: A data consumer who finds two populations of T cells would like to know: Are those T cells present in both donors, or are either population specific to Ambrose or Jason?**_

brianraymor:
Closes #164

### DIFF
--- a/schema/3.0.0/schema.md
+++ b/schema/3.0.0/schema.md
@@ -440,7 +440,9 @@ Curators MUST annotate the following columns in the `obs` dataframe:
     </tr>
     <tr>
       <th>Value</th>
-        <td>categorical with <code>str</code> categories. If unavailable, this MUST be <code>"unknown"</code>.<br><br>This MUST be free-text that identifies a unique individual that data were derived from. It is STRONGLY RECOMMENDED that this identifier be unique to a given individual within the collection of datasets that includes this dataset. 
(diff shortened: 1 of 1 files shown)


schema/3.0.0/schema.md
```diff
@@ -440,7 +440,9 @@ Curators MUST annotate the following columns in the `obs` dataframe:
     </tr>
     <tr>
       <th>Value</th>
-        <td>categorical with <code>str</code> categories. If unavailable, this MUST be <code>"unknown"</code>.<br><br>This MUST be free-text that identifies a unique individual that data were derived from. It is STRONGLY RECOMMENDED that this identifier be unique to a given individual within the collection of datasets that includes this dataset.
```
jahilton, line 443: In what scenario(s) would we expect 'unknown'?

brianraymor: Are there not existing scenarios in the data corpus where there is no donor field or its equivalent?

brianraymor: Discussed in slack. Addressed in a7dd8e9616d84f8e8ade1985b49c5514157cc473.

PR #276, opened 2022-07-18
Suspension type

### LINKED ISSUE #227: capture single-cell vs single-nucleus
Researchers need to be aware of which data are single-cell vs single-nucleus so that they can take special care to ensure proper integration.
Most contributors don't provide it automatically because it usually isn't a variable in their study (they either have all sc or all sn). However, when a study does include sn + sc data, it is of significant discussion...
https://doi.org/10.1038/s41598-020-58327-6
https://doi.org/10.1371/journal.pone.0209648
https://doi.org/10.1002/hep4.1854

This was also supported during a recent data integration workshop. Data integrators were given datasets in schema 2.0.0 and one of the first tasks they did was to curate each dataset as single-cell or single-nucleus.

**Proposal**
Lattice submits data as obs.suspension_type with values either `cell` or `nucleus`
Spatial data will need to be considered for when multiple cells are likely captured (`spot` is common terminology for Visium - [ref](https://kb.10xgenomics.com/hc/en-us/articles/360035487952-How-many-cells-are-captured-in-a-single-spot-), while `bead` is common for Slide-seq - [ref](https://doi.org/10.1038/s41587-020-0739-1))

brianraymor:
Closes #227 

The _spatial proteomics_ link is not live yet because EFO released this morning.

### DIFF
--- a/schema/3.0.0/schema.md
+++ b/schema/3.0.0/schema.md
@@ -550,7 +550,7 @@ Curators MUST annotate the following columns in the `obs` dataframe:
           </thead>
           <tbody>
             <tr>
-              <td><i>10x transcription profiling</i> [<a href="http://www.ebi.ac.uk/efo/EFO_0030080"><code>EFO:0030080</code></a>]  and its children</td>
+              <td><i>10x transcription profiling</i> [<a href="http://www.ebi.ac.uk/efo/EFO_0030080"><code>EFO:0030080</code></a>] and its children</td>
               <td><code>"cell"</code> or <code>"nucleus"</code></td>
            </tr> 
             <tr>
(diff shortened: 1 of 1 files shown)


jahilton (commented):
I feel like the assays that aren't on the table should be addressed.

This MUST be the correct type for the corresponding assay. If the assay is not addressed by the table, then select the most appropriate value and inform the curation team during submission so that the table can be updated.:

brianraymor:
I'll make another editorial pass.

> If the assay is not addressed by the table

I outlined my CLI intentions [earlier.](https://github.com/chanzuckerberg/single-cell-curation/issues/227#issuecomment-1185976294) which was to warn when an assay is not in the table, so we can create a tracking issue, but we can do _belts and suspenders_.

brianraymor:
Addressed in 9d129f3163b96a4f3a27bf02ae40603d1c2b75cb

jahilton (approved):

PR #281, opened 2022-07-20
feat: add --ignore-labels flag

ebezzi:
Add an `--ignore-labels` flag. If true, the validator will not check for reserved column names that would be added by the `--add-labels` option.

This is useful if you want to validate a dataset that has already been uploaded to the cellxgene data portal.

### DIFF
--- a/cellxgene_schema_cli/cellxgene_schema/cli.py
+++ b/cellxgene_schema_cli/cellxgene_schema/cli.py
@@ -30,12 +30,18 @@ def schema_cli():
     type=click.Path(exists=False, dir_okay=False, writable=True),
 )
 @click.option(
+    "-i",
+    "--ignore-labels",
+    help="Ignore ontology labels when validating",
+    is_flag=True
+)
+@click.option(
     "-v",
     "--verbose",
     help="When present will set logging level to debug",
     is_flag=True
 )
-def schema_validate(h5ad_file, add_labels_file, verbose):
+def schema_validate(h5ad_file, add_labels_file, ignore_labels, verbose):
     # Imports are very slow so we defer loading until Click arg validation has passed
 
     print("Loading dependencies")
@@ -47,7 +53,7 @@ def schema_validate(h5ad_file, add_labels_file, verbose):
     print("Loading validator modules")
     from .validate import validate
 
-    is_valid, _, _ = validate(h5ad_file, add_labels_file, verbose=verbose)
+    is_valid, _, _ = validate(h5ad_file, add_labels_file, ignore_labels, verbose=verbose)
     if is_valid:
         sys.exit(0)
     else:
--- a/cellxgene_schema_cli/cellxgene_schema/cli.py
+++ b/cellxgene_schema_cli/cellxgene_schema/cli.py
@@ -53,7 +53,7 @@ def schema_validate(h5ad_file, add_labels_file, ignore_labels, verbose):
     print("Loading validator modules")
     from .validate import validate
 
-    is_valid, _, _ = validate(h5ad_file, add_labels_file, ignore_labels, verbose=verbose)
+    is_valid, _, _ = validate(h5ad_file, add_labels_file, ignore_labels=ignore_labels, verbose=verbose)
     if is_valid:
         sys.exit(0)
     else:
--- a/cellxgene_schema_cli/cellxgene_schema/validate.py
+++ b/cellxgene_schema_cli/cellxgene_schema/validate.py
@@ -28,7 +28,7 @@ class Validator:
 
     schema_definitions_dir = env.SCHEMA_DEFINITIONS_DIR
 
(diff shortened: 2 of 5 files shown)


cellxgene_schema_cli/tests/fixtures/examples_validate.py
```diff
@@ -203,7 +203,7 @@
 adata_with_labels = anndata.AnnData(
     X=sparse.csr_matrix(X),
     obs=pd.concat([good_obs, obs_expected], axis=1),
-    var=pd.concat([good_var, var_expected], axis=1),
+    var=var_expected,
```
ebezzi, line 206: Using `pd.concat` here corrupts the resulting dataframe which leads to a failed validation. `var_expected` is a superset of `good_var` so this is also fine.

cellxgene_schema_cli/tests/test_validate.py
```diff
@@ -204,6 +204,32 @@ def test_get_dictionary_mapping_curie(self):
...
+        is_valid = validator.validate_adata()
+
+        self.assertFalse(is_valid)
+
+    def test_validating_labeled_h5ad_should_pass_if_flag_set(self):
+
+        validator = Validator(ignore_labels=True)
+        import copy
+        validator.adata = copy.deepcopy(self.test_adata_with_labels)
+        validator.adata.uns["X_normalization"] = 'none'
```
ebezzi, line 227: Won't pass validation if `X_normalization` isn't `none`.

brianraymor: Also remember that we're deprecating `X_normalization` in 3.0.0.

ebezzi: Good point. I feel even less guilty about this hack now.

cellxgene_schema_cli/tests/fixtures/examples_validate.py
MDunitz: just curious -- why isnt that also an issue for obs?

cellxgene_schema_cli/tests/test_validate.py
```diff
@@ -204,6 +203,32 @@ def test_get_dictionary_mapping_curie(self):
...
+        validator = Validator()
+        validator.adata = self.test_adata_with_labels
+        is_valid = validator.validate_adata()
+
+        self.assertFalse(is_valid)
+
+    def test_validating_labeled_h5ad_should_pass_if_flag_set(self):
+
+        validator = Validator(ignore_labels=True)
+        import copy
```
MDunitz, line 224: is there a reason for putting the import here?

cellxgene_schema_cli/tests/fixtures/examples_validate.py
ebezzi: I think the issue is that duplicate columns are merged in a weird way (if you select the column, you get a `Dataframe` and not a `Series`). The two obs do not have duplicate columns.

cellxgene_schema_cli/tests/test_validate.py
ebezzi: Not necessarily. I will move it to the top.

cellxgene_schema_cli/cellxgene_schema/cli.py
```diff
@@ -47,7 +53,7 @@ def schema_validate(h5ad_file, add_labels_file, verbose):
     print("Loading validator modules")
     from .validate import validate
 
-    is_valid, _, _ = validate(h5ad_file, add_labels_file, verbose=verbose)
+    is_valid, _, _ = validate(h5ad_file, add_labels_file, ignore_labels, verbose=verbose)
```
pablo-gar, line 56: Shouldn't it be? 
```suggestion
    is_valid, _, _ = validate(h5ad_file, add_labels_file, ignore_labels=ignore_labels, verbose=verbose)
```

cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -1312,7 +1318,7 @@ def validate(
         logging.basicConfig(level=logging.DEBUG)
     else:
         logging.basicConfig(level=logging.INFO, format="%(message)s")
-    validator = Validator()
+    validator = Validator(ignore_labels)
```
pablo-gar, line 1321:
```suggestion
    validator = Validator(ignore_labels=ignore_labels)
```

pablo-gar (changes requested):
minor changes

cellxgene_schema_cli/cellxgene_schema/cli.py
pablo-gar: better to explicitly name them in case more args get added in the future

pablo-gar (approved):

PR #296, opened 2022-07-27
feat: deprecate X_normalization

ebezzi:
Removes `X_normalization` from the code. 

Also replaces the `2_0_0.yml` schema definition with `3_0_0.yml` and updates the relevant test h5ad. Note that `3_0_0.yml` is - for now - just a copy of the old `2_0_0.yml` - it will be updated as we go with more PRs.

Note: we won't enforce the absence of `X_normalization` in this PR, but we will do it as part of https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-curation/264


cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -1076,47 +1033,20 @@ def _validate_raw(self):
         # If all checks passed then proceed with validation
         if all(checks):
 
-            normalization = self.adata.uns["X_normalization"]
-
             # If both "raw.X" and "X" exist but neither are raw
             if not self._is_raw() and self._get_raw_x_loc() == "raw.X":
                 self.warnings.append(
                     "Raw data may be missing: data in 'raw.X' contains non-integer values."
                 )
```
pablo-gar, line 1041:
```suggestion
            # Only "X" exists but it's not raw
            if (
                not self._is_raw()
                and self._get_raw_x_loc() == "X"
            ):
               self.errors.append(
                       "Only `X` was found and it doesn't contain raw values.
                )
```

cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -1076,47 +1033,20 @@ def _validate_raw(self):
...
-
             # If raw data is in X and there is nothing in raw.X (i.e. normalized values are not provided), then
             # add a warning because normalized data for RNA data is STRONGLY RECOMMENDED
             if (
                 self._is_raw()
                 and self._get_raw_x_loc() == "X"
-                and normalization == "none"
             ):
-
                 self.warnings.append(
```
pablo-gar, line 1048: Should this be an error? @brianraymor 

```suggestion
                self.error.append(
```

cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -1076,47 +1033,20 @@ def _validate_raw(self):
         # If all checks passed then proceed with validation
         if all(checks):
 
-            normalization = self.adata.uns["X_normalization"]
-
             # If both "raw.X" and "X" exist but neither are raw
             if not self._is_raw() and self._get_raw_x_loc() == "raw.X":
                 self.warnings.append(
```
pablo-gar, line 1038: @brianraymor should this be an error?
```suggestion
                self.error.append(
```

pablo-gar (changes requested):

pablo-gar (approved):
LGTM

PR #298, opened 2022-07-28
Pin, download and process latest ontology versions for cellxgene schema 3.0.0

nayib-jose-gloria:
Issue: https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-curation/254

Changes:
- Updated owl_info.yaml to pin to latest ontology versions, and ran `make update-references` (as instructed in readme) to update files accordingly


pablo-gar (approved):
LGTM

ebezzi (approved):

brianraymor:
Why are ontologies being copied? There's been an ongoing discussion about ensuring that long term all the ontology processing occurs in one places; otherwise, there's the potential for brittle dependencies.

ebezzi:
@brianraymor The original idea was to pull them from this repo but there were some extra complexities (due to dependencies, IIRC) and we ended up copying them. I agree it's not ideal, we should probably go back to our original plan but that will take a while.

PR #307, opened 2022-08-03
feat: add validation for deprecated columns

ebezzi:


cellxgene_schema_cli/cellxgene_schema/schema_definitions/3_0_0.yaml
```diff
@@ -105,6 +117,9 @@ components:
         required: null # Means it's required
         index:
             unique: true
+        # deprecated_columns:
```
ebezzi, line 120: Needs to be uncommented when https://github.com/chanzuckerberg/single-cell-curation/pull/305 is merged

cellxgene_schema_cli/tests/test_schema_compliance.py
```diff
@@ -597,6 +597,19 @@ def test_categories_with_zero_values_warn(self):
             ],
         )
 
+    def _test_deprecated_fields(self):
```
ebezzi, line 600: Remove the initial `_` when https://github.com/chanzuckerberg/single-cell-curation/pull/305 is merged

cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -1151,6 +1167,11 @@ def _deep_check(self):
                 "adata is loaded in backed mode, validation requires anndata to be loaded in memory"
             )
 
+        # Checks for deprecated columns
+        self._check_deprecated_columns()
+        if self.errors:
+            return
```
nayib-jose-gloria, line 1173: I don't think we want this check + return here, the current pattern is collecting the various validation errors in one pass and printing all the error msgs in self.errors at the top-level call

ebezzi: Good point, this was actually something I added temporarily. Removing it.

nayib-jose-gloria (approved):
looks good!

Issue #253, opened 2022-06-29
cellxgene-schema must not allow clarifying text for assay_ontology_term_id


brianraymor:
The schema has removed guidance in `assay_ontology_term_id` that allowed clarifying text enclosed in parentheses if there was not an exact match for an assay:

`assay_ontology_term_id`

 If there is not an exact match for the assay, clarifying text MAY be enclosed in parentheses and appended to the most accurate term. For example, the sci-plex assay could be curated as <code>"EFO:0010183 (sci-plex)"</code>

If present, then validation MUST fail and display a warning.

---

The following must be removed from the cellxgene-schema label application step:

`assay`

Any clarifying text enclosed in parentheses and appended to <code>assay_ontology_term_id</code> MUST be appended to <code>assay</code>.<br><br> For example, if the sci-plex assay was curated as <code>"EFO:0010183 (sci-plex)"</code>, then the value would be <code>"single-cell library construction (sci-plex)"</code>.

jychien:
Tested the following cases. It looks good, but logging could be slightly improved to be more precise:
| Test Case                 | Expected Result<img width=200/> | Obtained Result                                                                                                                                                                                                                                                                  | Test File  <img width=200/> | Needs attention<img width=100/> |
| --------------------------------------------------------------------------------- | --------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------------------ | --------------- |
| assay\_ontology\_term\_id appended with ' (Slide-seqV2)'                          | Fail validation | Failed validation<br>Logging: ERROR: 'EFO:0009899 (Slide-seqV2)' in 'assay\_ontology\_term\_id' is not a valid ontology term id of 'EFO'.<br>ERROR: 'EFO:0009899 (Slide-seqV2)' in 'assay\_ontology\_term\_id' is not a child term id of '\[\['EFO:0002772', 'EFO:0010183'\]\]'. | adata\_assay\_text.h5ad  |                 |
| valid efo assay\_ontology\_term\_id but not a child of EFO:0002772 or EFO:0010183 | Fail validation | Failed validation<br>Logging: ERROR: 'EFO:0006571' in 'assay\_ontology\_term\_id' is not a child term id of '\[\['EFO:0002772', 'EFO:0010183'\]\]'.                                                                                                                              | adata\_assay\_child.h5ad |                 |

[...]

PR #310, opened 2022-08-05
chore(error logging): remove superfluous validation steps and error logging

nayib-jose-gloria:
Changes: 
- If term_id does not belong to an allowed ontology, skip follow-up validation steps that will always fail as a result (i.e. checks that its a valid child term of specified ancestors, or a specified allowed term from within an allowed ontology type)
- Based on feedback from Jenny [here](https://github.com/chanzuckerberg/single-cell-curation/issues/253#issuecomment-1205866042) for more precise error logging 
- Updated tests


pablo-gar (approved):

nayib-jose-gloria:
duplicate goal to https://github.com/chanzuckerberg/single-cell-curation/pull/319

Issue #257, opened 2022-06-29
cellxgene-schema must annotate feature_biotype


brianraymor:
**Note: Clarifying a [dependency](https://github.com/chanzuckerberg/single-cell-curation/issues/236#issuecomment-1170511153)**.

Previously, `feature_biotype` was [set by Curators](https://github.com/chanzuckerberg/single-cell-curation/issues/236). It now needs to be annotated by the Data Portal.

When a dataset is uploaded, cellxgene Data Portal MUST automatically add the matching human-readable name for the corresponding feature biotype, identifier, and the NCBITaxon term for the reference organism to the `var` and `raw.var` dataframes. Curators MUST NOT annotate the following columns:

### feature_biotype

<table><tbody>
    <tr>
      <th>Key</th>
      <td>feature_biotype</td>
    </tr>
    <tr>
      <th>Annotator</th>
      <td>Data Portal</td>
    </tr>
    <tr>
      <th>Value</th>
        <td>This MUST be <code>"gene"</code> or <code>"spike-in"</code>.  
        </td>
    </tr>
</tbody></table>
<br>

jychien:
I've tested genes and spike-ins for datasets with only raw matrix and datasets with both raw and normalized matrix. Everything looks good!
| Test Case                                            | Expected Result                                             | Obtained Result                                                              | Test File                        | Needs attention |
| ---------------------------------------------------- | ----------------------------------------------------------- | ---------------------------------------------------------------------------- | -------------------------------- | --------------- |
| scRNA-seq with both var and raw.var gene features    | Pass validation, and add feature\_biotype with --add-labels | Passed validation and added feature\_biotype='gene' for var and raw.var      | adata\_biotype\_genes\_both.h5ad |                 |
| scRNA-seq with both var and raw.var spike-ins        | Pass validation, and add feature\_biotype with --add-labels | Passed validation and added feature\_biotype='spike-ins' for var and raw.var | adata\_biotype\_ercc\_both.h5ad  |                 |
| scRNA-seq with only var and no raw.var gene features | Pass validation, and add feature\_biotype with --add-labels | Passed validation and added feature\_biotype='gene' for var                  | adata\_biotype\_genes.h5ad       |                 |
| scRNA-seq with only var and no raw.var spike-ins     | Pass validation, and add feature\_biotype with --add-labels | Passed validation and added feature\_biotype='spike-ins' for var and raw.var | adata\_biotype\_ercc.h5ad        |                 |

PR #311, opened 2022-08-05
feat: automate adding feature_biotype column to var and raw.var after schema validation

nayib-jose-gloria:
#257 

Changes:
- feature_biotype is now automatically set in var and raw.var based on the feature ID, and thus is now a reserved column name in the schema 
- feature_id's starting with ERCC -> "spike-in". All else, "gene" (the validator already runs validations that the feature_id is either a valid ENSEMBL or ERCC ID, so I believe this condition is sufficient for setting the feature_biotype)
- raw.var no longer has required columns, so some minor refactors account for the change in our schema yaml (i.e. checks for whether components have a "columns" entry) 
- updated tests + removed feature_biotype column from valid example h5ad test fixture


cellxgene_schema_cli/cellxgene_schema/write_labels.py
```diff
@@ -202,6 +202,25 @@ def _get_mapping_dict_feature_reference(
...
+        :return a mapping dictionary: {id: "gene", id: "spike-in", ...}
+        :rtype dict
+        """
+        mapping_dict = {}
+
+        for i in ids:
+            if i.startswith('ERCC'):
+                mapping_dict[i] = "spike-in"
+            else:
+                mapping_dict[i] = "gene"
```
pablo-gar, line 220: In the future we are likely to accept "protein" as `feature_type` so let's not make it just an else for gene, raising an error ensures that future developers won't let this slip

```suggestion
            elif i.startswith('ENS'):
                 mapping_dict[i] = "gene"
            else:
               raise ValueError (f"{i} is not a recognized `feature_name` and cannot be assigned a `feature_type`)
```

pablo-gar (changes requested):

pablo-gar (approved):

Issue #215, opened 2022-04-26
cellxgene-schema must enforce the AnnData encoding version required by the schema


brianraymor:
### AnnData Encoding

See the related conversation in [**#single-cell-data-wrangling**](https://czi-sci.slack.com/archives/C024HCSH9PT/p1648255752093579).

`cellxgene-schema` CLI must enforce the encoding requirements in the schema:

_The canonical data format for the cellxgene Data Portal is HDF5-backed [AnnData](https://anndata.readthedocs.io/en/latest) as written by version 0.8 of the anndata library. Part of the rationale for selecting this format is to allow cellxgene to access both the data and metadata within a single file. The schema requirements and definitions for the AnnData X, obs, var, raw.var, obsm, and uns attributes are described below._

brianraymor:
@jahilton - for visibility since he reported the issue that resulted in the requirement to validate encoding.

@ebezzi raised an issue in the connected [PR](https://github.com/chanzuckerberg/single-cell-curation/pull/317#issue-1333334668):
> I don't think we should enforce that the AnnData version is less than 0.8 as it should be backward compatible. It would be great to enforce AnnData <= 0.8 though (since future versions of AnnData will likely fail with unpredictable errors), but unfortunately, it doesn't look like AnnData adds the version to the file metadata, so that's not possible. This could be a good addition to AnnData, if we're ok with submitting an issue. Thoughts @brianraymor @maniarathi?

There is a note in a [comment](https://github.com/scverse/anndata/issues/698#issuecomment-1028201630) that claims:

> The file format version is something that's new this version!

I have not tracked this down yet.

Is there also a possibility, that there are different encoding versions used for data types for the [on disk format](https://github.com/scverse/anndata/blob/74b63d41a9494d1d5bb78915cab9fb23169a287f/docs/fileformat-prose.rst) that could provide a heuristic? 

@pablo-gar @ambrosejcarr - any thoughts?

brianraymor:
RE _the [on disk format](https://github.com/scverse/anndata/blob/74b63d41a9494d1d5bb78915cab9fb23169a287f/docs/fileformat-prose.rst) that could provide a heuristic?_ 

```
import h5py
f = h5py.File("anndata8.h5ad", "r")
dict(f["obs"].attrs)
```

returns an encoding version of:

```
 'encoding-type': 'dataframe',
 'encoding-version': '0.2.0'}
```

compared to inspecting an anndata 7 dataset that returns:

```
[...]

PR #317, opened 2022-08-09
chore: update to anndata 0.8

ebezzi:
Update to AnnData 0.8. Notes:

- I don't think we should enforce that the AnnData version is less than 0.8 as it should be backward compatible. It would be great to enforce AnnData <= 0.8 though (since future versions of AnnData will likely fail with unpredictable errors), but unfortunately, it doesn't look like AnnData adds the version to the file metadata, so that's not possible. This could be a good addition to AnnData, if we're ok with submitting an issue. Thoughts @brianraymor @maniarathi?
- The 3.0.0 schema enforcing is already in the code.


brianraymor:
Moving this requirements conversation to the [connected product issue](https://github.com/chanzuckerberg/single-cell-curation/issues/215).

pablo-gar (approved):

PR #318, opened 2022-08-09
feat: warn for low gene count

ebezzi:
Adds a warning for low gene counts. 

Note: since this warning will raise by default on all tests (since they only have four genes), I have added a mocking solution that allows overriding the schema configuration settings.


cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -136,7 +136,8 @@ def _set_schema_def(self):
                     f"Validation cannot be performed."
                 )
             else:
-                self.schema_def = schema.get_schema_definition(version)
+                if not self.schema_def:
```
ebezzi, line 139: Required so that `_set_schema_def` can be called externally (by tests) to enable mocking. Without this line, it would be overwritten each time `validate_adata()` is called

nayib-jose-gloria (approved):
wait on user opinions re: the warning wording before merging, but the code refactor looks good!

cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -700,6 +701,14 @@ def _validate_dataframe(self, df_name: str):
                 self._get_column_def(df_name, "index"),
             )
 
+        low_rows_threshold = self._get_component_def(df_name).get("warn_if_less_than_rows")
+        if low_rows_threshold is not None:
+            num_rows = df.shape[0]
+            if num_rows < low_rows_threshold:
+                self.warnings.append(
+                    f"Dataframe '{df_name}' only has {num_rows} rows. This might indicate an error in the dataset."
```
ebezzi, line 709:
```suggestion
                    f"Dataframe '{df_name}' only has {num_rows} rows. Features SHOULD NOT be filtered from expression matrix."
```

pablo-gar (approved):

PR #319, opened 2022-08-09
chore: remove ancestry-related errors if a term is in wrong ontology

ebezzi:
Note: this solves issue number 2 from the ticket.


brianraymor:
@ebezzi - I think that the `#2` reference above just linked to a closed issue and not what you actually wanted to reference?

pablo-gar (approved):

Issue #273, opened 2022-07-15
cellxgene-schema must validate suspension_type


brianraymor:
See [schema reference](https://github.com/chanzuckerberg/single-cell-curation/blob/main/schema/3.0.0/schema.md#suspension_type).

1. `cellxgene-schema `MUST warn and fail if the corresponding assay is present in the table, but the value of the `suspension_type` does not match the required value(s) in the table.
2. `cellxgene-schema` MUST warn if the corresponding assay is not present in the table. 

---

Curators MUST annotate the following column in the `obs` dataframe:

### suspension_type

<table><tbody>
    <tr>
      <th>Key</th>
      <td>suspension_type</td>
    </tr>
    <tr>
      <th>Annotator</th>
      <td>Curator</td>
    </tr>
    <tr>
      <th>Value</th>
        <td>categorical with <code>str</code> categories. This MUST be <code>"cell"</code>, <code>"nucleus"</code>, or <code>"na"</code>.<br>
        <br>This MUST be the correct type for the corresponding assay:
          <br><br>
          <table>
          <thead>
          <tr>
          <th>For Assay</th>
          <th>MUST Use</th>
          </tr>
          </thead>
          <tbody>
            <tr>
              <td><i>10x transcription profiling</i> [<a href="http://www.ebi.ac.uk/efo/EFO_0030080"><code>EFO:0030080</code></a>] and its children</td>
              <td><code>"cell"</code> or <code>"nucleus"</code></td>
           </tr> 
            <tr>
              <td><i>ATAC-seq</i> [<a href="http://www.ebi.ac.uk/efo/EFO_0007045"><code>EFO:0007045</code></a>] and its children</td>
              <td><code>"nucleus"</code></td>
           </tr>
            <tr>
              <td><i>CEL-seq2</i> [<a href="http://www.ebi.ac.uk/efo/EFO_0010010"><code>EFO:0010010</code></a>]</td>
              <td><code>"cell"</code> or <code>"nucleus"</code></td>
           </tr>
            <tr>
              <td><i>CITE-seq</i> [<a href="http://www.ebi.ac.uk/efo/EFO_0009294"><code>EFO:0009294</code></a>] and its children</td>
              <td><code>"cell"</code></td>
           </tr>
            <tr>
              <td><i>DroNc-seq</i> [<a href="http://www.ebi.ac.uk/efo/EFO_0008720"><code>EFO:0008720</code></a>]</td>
[...]

PR #322, opened 2022-08-10
feat: add suspension_type

nayib-jose-gloria:
#273 

Notes for Reviewers: 
- Please opine on warning message 
- Please opine on newly introduced yaml headings (warning_message, complex_rule and its subfields, etc.)
- I separated all the rules to be able to provide specific error messages, but let me know if you think it would be a cleaner pattern to group the 'cell'-only, 'nucleus'-only, 'na' only, 'cell' or 'nucleus' rules into groups and give more generic messages (i.e. something like 'ERROR: assay in [list of rule group assays] found to have invalid suspension type')


cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -515,9 +521,32 @@ def _validate_column_dependencies(
...
+                    ontology_keys = []
+                    ancestor_list = []
+                    for key, val in ancestor_map.items():
+                        ontology_keys.append(key)
+                        ancestor_list.append(val)
+                    query_exp = f"@filter_df_by_ancestor_match({rule_def['column']}, {ontology_keys}, {ancestor_list})"
+            elif "rule" in dependency_def:
+                query_exp = dependency_def["rule"]
+            else:
+                continue
```
nayib-jose-gloria, line 545: I added this else: continue with the idea of not breaking if we add another subcase in the future but I can remove if its too confusing

cellxgene_schema_cli/tests/fixtures/examples_validate.py
```diff
@@ -46,10 +46,11 @@
             "HANCESTRO:0575",
             "HsapDv:0000003",
             "donor_1",
+            "nucleus",
         ],
         [
             "CL:0000192",
-            "EFO:0010183",
+            "EFO:0009918",
```
nayib-jose-gloria, line 53: changed because previous example was not a valid EFO for suspension_type, would add warnings to all the test outputs

ebezzi (approved):

cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -540,9 +546,32 @@ def _validate_column_dependencies(
...
+                        allowed_ancestors = dict(zip(ontologies, ancestors))
+                        return validate_curie_ancestors_vectorized(term_id, "", allowed_ancestors, inclusive, False)
+
+                    # hack: pandas dataframe query doesn't support Dict inputs
+                    ontology_keys = []
+                    ancestor_list = []
+                    for key, val in ancestor_map.items():
+                        ontology_keys.append(key)
+                        ancestor_list.append(val)
+                    query_exp = f"@filter_df_by_ancestor_match({rule_def['column']}, {ontology_keys}, {ancestor_list})"
```
pablo-gar, line 566: Can you encapsulate this in a separate function for readability?

pablo-gar (changes requested):

Issue #262, opened 2022-07-01
cellxgene-schema must perform basic 2.0.0 -> 3.0.0 migration


brianraymor:
To make it easier for Lattice and/or the Data Portal to migrate datasets from 2.0.0 to 3.0.0:

The `cellxgene-schema` CLI should have an option to update the datasets prior to re-validation:

- Rename `ethnicity_term_ontology_id` to `self_reported_ethnicity_term_ontology_id`
- Rename `ethnicity` to `self_reported_ethnicity`
- Delete all deprecated fields in the table below:

<table>
<thead>
  <tr>
    <th>Reserved Name</th>
    <th>AnnData</th>
    <th>Deprecated in</th>
  </tr>
</thead>
<tbody>
  <tr>
    <td>X_normalization</td>
    <td>uns</td>
    <td>3.0.0</td>
  </tr>
  <tr>
    <td>default_field</td>
    <td>uns</td>
    <td>2.0.0</td>
  </tr>
  <tr>
    <td>layer_descriptions</td>
    <td>uns</td>
    <td>2.0.0</td>
  </tr>
  <tr>
    <td>tags</td>
    <td>uns</td>
    <td>2.0.0</td>
  </tr>
   <tr>
    <td>version</td>
    <td>uns</td>
    <td>2.0.0</td>
  </tr>
  <tr>
    <td>contributors</td>
    <td>uns</td>
    <td>1.1.0</td>
  </tr>
  <tr>
    <td>preprint_doi</td>
    <td>uns</td>
    <td>1.1.0</td>
  </tr>
  <tr>
    <td>project_description</td>
    <td>uns</td>
    <td>1.1.0</td>
  </tr>
  <tr>
    <td>project_links</td>
    <td>uns</td>
    <td>1.1.0</td>
  </tr>
  <tr>
    <td>project_name</td>
    <td>uns</td>
    <td>1.1.0</td>
  </tr>
  <tr>
    <td>publication_doi</td>
    <td>uns</td>
    <td>1.1.0</td>
  </tr>
  <tr>
    <td></td>
    <td></td>
    <td></td>
  </tr>
</tbody>
</table>

- Set `suspension_type` based on the corresponding `assay_ontology_term_id` if possible - some assays support multiple choices.
- Update `schema_version` to "3.0.0"

jahilton:
I think 3. Delete X_normalization should be updated to
3. Delete any deprecated field (#264 )

brianraymor:
[...]

PR #323, opened 2022-08-11
feat: 2to3 conversion

ebezzi:
#262

Should we print a warning for a non-automated suspension type, or is setting to NaN sufficient? Should it be under a different command?


cellxgene_schema_cli/cellxgene_schema/convert.py
```diff
@@ -0,0 +1,64 @@
...
+    ontology_checker = ontology.OntologyChecker()
+
+    def assign_suspension_type(item):
+        if item in match_assays:
+            return match_assays[item]
+        else:
+            for k, v in match_assays_or_children.items():
+                try:
+                    if ontology_checker.is_descendent_of("EFO", item, k):
+                        return v
```
nayib-jose-gloria, line 55: we should update this to v[0], or update the assay map values to strings or else we'll be assigning lists and the category validation will fail

nayib-jose-gloria: let's just update the assay map to map to a string

cellxgene_schema_cli/cellxgene_schema/convert.py
```diff
@@ -0,0 +1,64 @@
...
+                except Exception:
+                    return np.nan
+        return np.nan
+                
+    dataset.obs["suspension_type"] = dataset.obs.apply(lambda row: assign_suspension_type(row.assay_ontology_term_id), axis = 1)
```
nayib-jose-gloria, line 60: idea, update this to 
```
suspension_type_map = {}
for item in dataset.obs["assay_ontology_term_id"].cat.categories:
    suspension_type_map[item] = assign_suspension_type(item)
dataset.obs["suspension_type"] = dataset.obs.apply(lambda row: suspension_type_map.get(row.assay_ontology_term_id), axis = 1)
```

to only iterate over categories. might also need a try-catch in case of no categories. That way we only call the ontology checker once per distinct assay type vs. every row.

cellxgene_schema_cli/cellxgene_schema/convert.py
```diff
@@ -0,0 +1,104 @@
...
+
+def convert(input_file, output_file):
+    print(f"converting {input_file} into {output_file}")
+    
+    dataset = ad.read_h5ad(input_file)
+
+    # Rename ethnicity_ontology_term_id field
+    dataset.obs.rename(columns={"ethnicity_ontology_term_id": "self_reported_ethnicity_ontology_term_id"}, inplace=True)
+    # For existing datasets with labels already added
+    dataset.obs.rename(columns={"ethnicity": "self_reported_ethnicity"}, inplace=True)
```
nayib-jose-gloria, line 14: helpful for reducing error output noise during mass automated conversion of datasets that already have added labels. Will have no effect on datasets without labels added.

Edge case: If someone erroneously adds an ethnicity column in addition to a self_reported_ethnicity column, could be confusing, but the docs already state that both self_reported_ethnicity and ethnicity are reserved names. So, validation should fail regardless.

nayib-jose-gloria (approved):

PR #328, opened 2022-08-18
Add logging for suspension_type auto-assignment in cellxgene-schema convert command

nayib-jose-gloria:
#262 

Achieves the following during the convert cli command:

If suspension_type column is already in obs, leave it in the object and report what the categories currently are for this column
If suspension_type is not present in obs, add suspension_type for assays that have only one possible value and report the value that was added
Else, if NaN was added for suspension_type, report this as well


jychien:
Looks good to me, thanks!

PR #330, opened 2022-08-23
refactor: throw error on anndata version not matching 0.8.0

nayib-jose-gloria:
#215 

Update from logging a warning on anndata version not matching 0.8.0, to logging an error.

Also, deleted our draft 2to3 notebook now that the conversion command is done.


jychien:
Looks good to me! Here is the error logging for reference:

> ERROR: The h5ad artifact was generated with an AnnData version different from 0.8.0.
Validation complete in 0:00:00.939354 with status is_valid=False

PR #331, opened 2022-08-24
Update python notebooks with the curator query parameter

Bento007:
#2793

### DIFF
--- a/notebooks/curation_api/python/get_collections.ipynb
+++ b/notebooks/curation_api/python/get_collections.ipynb
@@ -33,7 +33,11 @@
   {
    "cell_type": "markdown",
    "id": "2c9a0189",
-   "metadata": {},
+   "metadata": {
+    "pycharm": {
+     "name": "#%% md\n"
+    }
+   },
    "source": [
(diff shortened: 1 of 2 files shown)


notebooks/curation_api/python/get_collections.ipynb
```diff
@@ -89,19 +113,30 @@
...
+   "metadata": {
+    "pycharm": {
+     "name": "#%% md\n"
+    }
+   },
    "source": [
-    "### Fetch Collection"
+    "### Fetch Collection\n",
+    "#### arguments:\n",
+    "visibility: \"PRIVATE\" to retrive private collections\n",
```
nayib-jose-gloria, line 124: not sure why, but the github rendering isn't recognizing this newline. as a result, the args are appearing as part of the same line, which is confusing.

Bento007: it rendered fine on my system, but to be safe I made them bullet points.

nayib-jose-gloria (approved):

PR #333, opened 2022-08-24
fix: remove h5ad requirements

Bento007:
https://github.com/chanzuckerberg/single-cell-data-portal/issues/3102

The h5ad suffix is no longer required.


notebooks/curation_api/python/src/dataset.py
```diff
@@ -31,7 +30,7 @@ def get_identifier_type_and_value(identifier: str) -> Tuple[str, str]:
         identifier_type = "id"
     else:
         # CURATOR_TAG_PREFIX_REGEX is superfluous; leaving in to match lambda handler code; may use later
-        matched = re.match(f"({UUID_REGEX}|{CURATOR_TAG_PREFIX_REGEX})\\.{EXTENSION_REGEX}$", identifier)
+        matched = re.match(f"({UUID_REGEX}|{CURATOR_TAG_PREFIX_REGEX})$", identifier)
```
nayib-jose-gloria, line 33: nit: without the EXTENSION_REGEX check to differentiate it from the earlier check, we don't need to check for the UUID_REGEX again here right?

nayib-jose-gloria: ^ if that is true, we can probably simplify or remove most of the logic in this function, like the groupdict check for id below

nayib-jose-gloria (approved):

PR #336, opened 2022-08-29
chore: add raw API calls in Python

danieljhegeman:
https://github.com/chanzuckerberg/single-cell-data-portal/issues/3102

Port of R notebooks to Python: the user can see how the requests are formulated manually, which the current Python notebooks do not make clear.

### DIFF
--- a/notebooks/curation_api/python_raw/create_dataset_from_local_file.ipynb
+++ b/notebooks/curation_api/python_raw/create_dataset_from_local_file.ipynb
@@ -13,21 +13,46 @@
    "id": "1ff1ec6e",
    "metadata": {},
    "source": [
-    "The script in this notebook performs the upload of a local datafile to a given Collection (as identified by its Collection uuid), where the datafile becomes a Dataset accessible via the Data Portal UI.\n",
+    "_\\*\\*The sample code in this notebook limits s3 upload durations to 12 hours. If you think your large file upload may take longer than that, please make use of the `upload_local_datafile` function, whose underlying code supports unlimited upload duration, as seen in the `python/create_dataset_from_local_file.ipynb` notebook.\\*\\*_\n",
+    "\n",
+    "The script in this notebook performs the upload of a local datafile to a given Collection (as identified by its Collection id), where the datafile becomes a Dataset accessible via the Data Portal UI.\n",
     "\n",
     "In order to use this script, you must...\n",
     "- have a Curation API key (obtained from upper-righthand dropdown in the Data Portal UI after logging in)\n",
-    "- know the id of the Collection to which you wish to upload the datafile (from `/collections/<collection_id>` in url path in Data Portal UI)\n",
+    "- know the id of the Collection to which you wish to upload the datafile (taken from `/collections/<collection_id>` in url path in Data Portal UI when viewing the Collection)\n",
     "\n",
-    "**For new Dataset uploads**:\n",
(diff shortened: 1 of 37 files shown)


notebooks/curation_api/python_raw/create_dataset_from_local_file.ipynb
```diff
@@ -0,0 +1,302 @@
...
+    "\n",
+    "**For new Dataset uploads**:\n",
+    "- You must decide upon a string tag (the `curator_tag`) to use to uniquely identify the resultant Dataset within its Collection. This tag must *NOT* be the tag of an existing Dataset within this Collection (read on below), and it must _NOT_ conform to the uuid format.\n",
+    "\n",
+    "**For replacing/updating existing Datasets**:\n",
+    "- Uploads to a curator tag for which there already exists a Dataset in the given Collection will result in the existing Dataset being replaced by the new Dataset created from the datafile that you are uploading.\n",
+    "- Alternatively, while not necessarily recommended, an existing dataset _may_ be targeted for replacement by using the Dataset's Cellxgene uuid as the tag when writing to S3.\n",
+    "- You can only add/replace Datasets in private Collections or revision Collections.\n",
+    "\n",
+    "For all uploads, the `.h5ad` suffix must be appended to the tag in the S3 write key. See example below."
```
Bento007, line 30: this will soon be dated

notebooks/curation_api/python_raw/create_dataset_from_local_file.ipynb
```diff
@@ -0,0 +1,302 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "80427063",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import boto3\n",
```
Bento007, line 134: why is the dependecy import so low in the script. Move it closer to the top where it is most common

notebooks/curation_api/python_raw/create_dataset_from_local_file.ipynb
```diff
@@ -0,0 +1,302 @@
...
+   "source": [
+    "print(res.status_code)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "a79b1f19",
+   "metadata": {},
+   "source": [
+    "### Retrieve temporary s3 write credentials. These credentials will only work for _this_ Collection."
```
Bento007, line 205: This is not true. The credential work for all collections they own.

Bento007 (changes requested):
We on longer require the h5ad suffix.

notebooks/curation_api/python_raw/create_dataset_from_local_file.ipynb
danieljhegeman: I did that because I view the dependency import as not being in the category of "instructional field fill-in", which is what I wanted to have up-front, after the initial headings. 

Everything the curator really needs to pay attention to should come first. Then the dependencies can be imported.

I agree that this breaks typical code file conventions but this is also notebook code, not for a production service. Different goal

Bento007: If anyone has ever written in python they know imports come as early as possible. If they are looking at these notebooks, then they have written some python. I don't your argument as a compelling enough reason to break convention. Knowing you don't have the libraries required to run the script is helpful to know at the beginning No one will be confused by having import at the top. You can even put the imports right after the initial docstring.

danieljhegeman: Ok I'll make the reorder

Bento007 (commented):
Are you still planning on moving the imports?

Bento007 (approved):

PR #337, opened 2022-08-30
fix: remove h5ad documentation from notebooks

Bento007:
#3102


Bento007:
This work was complete in a different PR.

PR #338, opened 2022-09-02
fix: id regex

danieljhegeman:
- some minor doc updates
- identifier regex was using the wrong named regex

### DIFF
--- a/notebooks/curation_api/python/src/dataset.py
+++ b/notebooks/curation_api/python/src/dataset.py
@@ -100,9 +100,7 @@ def get_dataset(collection_id: str, identifier: str):
 
     params_dict = dict()
     params_dict[identifier_type] = identifier_value
-    print(url)
-    print(params_dict)
-    print(headers)
+
     try:
         res = requests.get(url, headers=headers, params=params_dict)
         res.raise_for_status()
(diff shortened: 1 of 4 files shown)


notebooks/curation_api/python/src/dataset.py
```diff
@@ -23,7 +23,7 @@
 
 
 def get_identifier_type_and_value(identifier_value: str) -> Tuple[str, str]:
-    if matched := re.match(f"({UUID_REGEX}|{CURATOR_TAG_PREFIX_REGEX})$", identifier_value):
+    if matched := re.match(f"({ID_REGEX}|{CURATOR_TAG_PREFIX_REGEX})$", identifier_value):
```
danieljhegeman, line 26: caused id's to not get recognized as such

Bento007 (approved):

notebooks/curation_api/python_raw/create_collection.ipynb
```diff
@@ -209,4 +209,4 @@
  },
  "nbformat": 4,
  "nbformat_minor": 5
-}
+}
```
danieljhegeman, line 212: this makes no sense to me 🤣 

it keeps happening

Issue #343, opened 2022-09-09
cellxgene-schema must validate Drop-seq refinement in suspension_type


brianraymor:
Please see the details in [Drop-seq must use either "cell" or "nucleus" in suspension_type](https://github.com/chanzuckerberg/single-cell-curation/issues/342).

nayib-jose-gloria:
release candidate for cellxgene-schema version with updated validation and conversion accounting for this change: https://drive.google.com/file/d/1PIC7UkqvRsIgeEHffxr87i5iqDvwDbCr/view?usp=sharing

jychien:
Just got the updated cellxgene-schema-3.0.0rc14.tar.gz from you, and it looks good! :) Thanks!

PR #344, opened 2022-09-09
fix: Update suspension_type validation rules for Drop-seq assays

nayib-jose-gloria:
#343 

Changes:
- updated 3.0.0 OpenAPI spec rules to allow obs rows with Drop-seq assays to have either nucleus OR cell as suspension_type
- updated convert script to account for not being able to auto-assign suspension_type to Drop-seq obs rows anymore + updated logged message to more clearly explain workflow where auto-assignment cannot be done.
- updated schema def doc
- update tests

### DIFF
--- a/cellxgene_schema_cli/cellxgene_schema/convert.py
+++ b/cellxgene_schema_cli/cellxgene_schema/convert.py
@@ -69,7 +69,7 @@ def convert(input_file, output_file):
     match_assays = {
         # 'EFO:0010010': ['cell', 'nucleus'], 
         'EFO:0008720': 'nucleus',
-        'EFO:0008722': 'cell',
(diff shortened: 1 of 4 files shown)


cellxgene_schema_cli/cellxgene_schema/convert.py
```diff
@@ -69,7 +69,7 @@ def convert(input_file, output_file):
     match_assays = {
         # 'EFO:0010010': ['cell', 'nucleus'], 
         'EFO:0008720': 'nucleus',
-        'EFO:0008722': 'cell',
+        # 'EFO:0008722': ['cell', 'nucleus'], ,
```
pablo-gar, line 72: Small nit, should we remove the commented out lines. Or maybe explain why they are here?

pablo-gar (approved):

PR #345, opened 2022-09-12
feat: add standalone 'unlabel' command to cli, for removing data processing pipeline added labels

nayib-jose-gloria:
- Added 'remove-labels' command to cli
- Moved overlapping code from convert command to remove_labels and imported into convert


metakuni:
Minor nit: We should probably also keep the `CHANGELOG.md` updated with all the 3.0.0 changes too.

cellxgene_schema_cli/cellxgene_schema/remove_labels.py
```diff
@@ -0,0 +1,44 @@
...
+        "assay",
+        "disease",
+        "ethnicity",
+        "organism",
+        "sex",
+        "tissue",
+        "self_reported_ethnicity",
+        "development_stage",
+    ]
+
```
pablo-gar, line 22: We should not hard-code these, we should use the schema definition to obtain these values.

Overall I believe the code has to be refractor to look something like this:

1. Make all of these functions into a class `AnnDataLabelRemover` similar to [`AnnDataLabelAppender`](https://github.com/chanzuckerberg/single-cell-curation/blob/de4b7544a508b1ecba0d85794e68e00aec8f04ed/cellxgene_schema_cli/cellxgene_schema/write_labels.py#L13)
2. Initialize it similar to `AnnDataLabelAppender` so that you have access to the [schema definition](https://github.com/chanzuckerberg/single-cell-curation/blob/de4b7544a508b1ecba0d85794e68e00aec8f04ed/cellxgene_schema_cli/cellxgene_schema/schema_definitions/3_0_0.yaml)
3. Create methods to retrieve what are the "labeled" fields of the current schema definition (they all have the `[add_labels][to_column]` keys)
4. Create method to remove those fields from AnnData

This would:

- Keep consistency with rest of codebase
- Easier to maintain, as changes in the schema definition would be reflected automatically

---

I would also encourage the use of docstrings and type hints. 

---

I believe the same should be done for the `convert.py` file, while that is very important to do it is probably for a different PR

pablo-gar (changes requested):

pablo-gar (approved):
Looks good, thanks for incorporating all the changes!

PR #346, opened 2022-09-15
feat: deprecate curator tag from and update python_raw notebooks

danieljhegeman:
@Bento007 

My suggested changes, various updates. Mostly fixing areas in headers that lacked specificity or that we missed in previous updates for other issues.

Very few of my changes actually overwrite yours from tsmith/3153. I think only a couple instances of "Dataset id" vs "dataset_id". Anyways, see what you think.

**I've only done them for the python_raw notebooks.** I'll copy over to the other two folders of notebooks after we agree on the changes so that I don't have to fix things 3 separate times.


Bento007 (approved):
Great work

danieljhegeman:
@Bento007 can I get you to re-approve now that I've copied over the changes to `python/` and `R/` as well? 😃

Bento007 (approved):

PR #348, opened 2022-09-15
feat: added tissue and cell type mappings. #3230.

MillenniumFalconMechanic:
Added a notebook for generating artifacts required for new ontology-aware tissue and cell type filters in the Single Cell Data Portal. The notebook is similar in concept to the existing [compute_ancestor_mapping.ipynb](https://github.com/chanzuckerberg/single-cell-curation/blob/main/notebooks/compute_ancestor_mapping.ipynb), which generates a dictionary used by the development stage filter functionality.


pablo-gar (approved):
I'm going to leave the same comment I left you on slack.

The gist of it is that I believe we should move away from generating fixtures using a notebook and make this into a formal class for centralized access to constants and methods.

These constants and methods should be used when enriching metadata (instead of dictionaries created by this notebook) and any other feature that requires so (e.g. Gene Expression aka scExpression aka WMG)

---

Full comment

I’m a little concerned that we are generating fixed dictionaries of descendants and ancestors and then using those to enrich the metadata. 

Was there a reason not to implement  on-the-fly methods that obtain the ascendants and descendants directly from the UBERON/CL ontologies, thus omitting the need to generate and store the dictionaries?

Similar to the work that I implemented on WMG’s [tissue mapper](https://github.com/chanzuckerberg/single-cell-data-portal/blob/main/backend/wmg/data/tissue_mapper.py)? (edited) 
I don’t want to block you on the staging deployment but I think we should move towards that paradigm for the following reasons:

1. No need to regenerate the dictionaries anytime a new cell type or tissue (specially orphans) is added.
2. Centralized place to access these functions
3. Ability to self encode and access the constants, SYSTEMS, ORGANS, etc.

PR #351, opened 2022-09-20
fix: update dataset endpoints to use dataset_id path param

danieljhegeman:
This commit does NOT include the update to `PUT /datasets` which will be added once that change is incorporated into the Curation API.


nayib-jose-gloria (approved):
looks good!

PR #358, opened 2022-09-25
feat: add create_dataset Python notebook

danieljhegeman:
- Add 'POST /datasets' for all 3 types of notebook
- Update header in local file notebook
- Rename the two upload notebooks for consistency
- do NOT introduce `id` response attribute naming refactor yet


Bento007 (approved):

Bento007:
How come we don't want to introduce the `id` refactor yet?

danieljhegeman:
> How come we don't want to introduce the id refactor yet?

It was a separate issue so I figured I should introduce it in a separate PR 🙂 

Keep PR's light

PR #367, opened 2022-10-07
feat: download progress indicator

danieljhegeman:
Add a visual progress indicator when downloading asset files.

For the python notebooks, this is a `%` number that increments.

For the R notebooks, I used the `httr` builtin `progress()` callback which produces a linear progress bar.


notebooks/curation_api/python/src/dataset.py
```diff
@@ -90,12 +90,19 @@ def download_assets(collection_id: str, dataset_id: str):
...
+            print(f"\nDownloading {asset['filetype']} file to {download_filename}... ")
             with requests.get(asset["presigned_url"], stream=True) as res:
                 res.raise_for_status()
+                filesize = int(res.headers["Content-Length"])
                 with open(download_filename, "wb") as df:
+                    total_bytes_received = 0
                     for chunk in res.iter_content(chunk_size=1024 * 1024):
                         df.write(chunk)
+                        total_bytes_received += len(chunk)
+                        percent_of_total_upload = float("{:.1f}".format(total_bytes_received / filesize * 100))
```
Bento007, line 102: what does the `:` mean in `{:.1f}`

Bento007 (approved):

notebooks/curation_api/python/src/dataset.py
danieljhegeman: the `:` is a field separator

the first entry (empty) is `field_name` (for str interpolation); the second (`.1f`) is `format_spec` which is what truncates the float to a single decimal

Issue #159, opened 2021-10-15
Document Lattice guidance for curators


jahilton:
We need to provide guidance for curators that are able to create Collections, submit h5ad datasets, publish, etc.
How to report error on upload or conversion, how to contact Lattice for a pre-Publish QA, etc.

brianraymor:
@jahilton - will you have time to address in Q3?

jahilton:
No, I think Q4 is more realistic for this

jahilton:
@brianraymor thoughts on where to keep these guidelines?
We have [this directory](https://github.com/Lattice-Data/lattice-tools/tree/main/cellxgene_resources) in our repo that we haven't publicized but will be shaping up for curator resources.
Or somewhere in single-cell-curation?

brianraymor:
I think it would be best to keep the guidelines in single-cell-curation with the schema and CLI.

PR #379, opened 2022-11-14
159-curation-guidelines

jahilton:
Addresses #159


BAevermann:
I think it reads well. Are we going to make those collection lists publicly viewable?

BAevermann (approved):

jahilton:
>Are we going to make those collection lists publicly viewable?

@BAevermann the sheet is currently publicly viewable. And I don't see anything like private collection links shared that shouldn't be. But it's also not organized well or particularly useful that we should be actively sharing the link broadly.

guidelines/curation_guidelines.md
```diff
@@ -0,0 +1,30 @@
...
+
```
brianraymor, line 6: RE _currently being curated_ - I would not expect the Datasheet sheet to expose private collections so the comment is confusing. I could imagine that the sheet might include the DOI and curator name for in progress items.

brianraymor, line 6: RE _currently being curated_ - I would not expect the Datasheet sheet to expose private collections so the comment is confusing. I could imagine that the sheet might include the DOI and curator name for in progress items.

jahilton: The Dataset sheet only includes DOI and curator team for in progress items, so this expectation is accurate.
Does this wording clarify that?
>The curator should then check the Dataset sheet for the **DOI of the** primary study and...

brianraymor: Yes.

jahilton: Addressed in 7fc87fd96ddb9d7eddbe34dd1b617d4139549ac6

brianraymor (approved):

PR #392, opened 2023-01-06
fix: update example_workflow curation notebook to fit redesigned, versioned data model

nayib-jose-gloria:
Changes: 

1) Added 2 lines to block under 'Create a revision of the Collection'
- In the new data model, private collections will return a 'version' collection ID as the API response 'id' of a collection, whereas public collections will return a 'canonical' collection ID, which remains constant across all published versions of a collection. You may think of the public-facing canonical collection ID as a permanent pointer to the most recently published version of a collection, whereas private collection ID's point to a specific version. As a result, to start a new revision of a public collection, you must pass the API response collection 'id' that is returned AFTER publish of a collection, as you cannot start revisions off versions. You may fetch the canonical collection ID programmatically by calling GET collections/<collection_id> and passing in the private collection version ID after publish. 

2) Simplified call under 'Replace the Second Dataset'
- When creating a revision, the datasets now retain the same ID they have in the published version of the collection (this is because collection versions simply retain a list of dataset version IDs they point to, so on creation, a new revision just points to the same list its published version points to). Therefore, 'original_id' is no longer a necessary field to match revision datasets to their published versions, and thus has been deprecated. Instead, if a user would like to target a particular dataset for replacement in a revision, they simply have to pass along the collection revision ID + dataset's ID directly.

### DIFF
--- a/notebooks/curation_api/python/example_workflow.ipynb
+++ b/notebooks/curation_api/python/example_workflow.ipynb
@@ -310,8 +310,8 @@
    "execution_count": null,
    "outputs": [],
    "source": [
-    "# After publish, the collection will now have a new permanent, canonical ID across versions.\n",
-    "# Use this ID for revisions.\n",
+    "# After publish, the collection will now have a new permanent, canonical ID that will stay constant across published versions.\n",
+    "# Use this ID to start a revision.\n",
     "collection_info = get_collection(new_collection_id)\n",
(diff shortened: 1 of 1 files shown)


Bento007:
Similar to the UI, the curators will expect the `new_collection_id` to be a `canonical_collection_id`.

Bento007:
I think we should hold off on doing this until we support versions in the API

PR #400, opened 2023-02-20
Use backed mode to read H5AD when validating

### LINKED ISSUE #399: Cannot validate large h5ad files because the entire X matrix is read into memory
## Motivation
This may just be a question (though, depending on the answer, it might be a feature request).

I'm a scientist at the Allen Institute for Brain Science. We have a single cell rnaseq dataset comprised of 4 million cells and 30 thousand genes. Stored as a sparse matrix in an h5ad file, this produces a 215 GB file. We would like to experiment with aligning our files with the cellxgene curation schema. However, running the validation tool is impossible. This line of code

https://github.com/chanzuckerberg/single-cell-curation/blob/main/cellxgene_schema_cli/cellxgene_schema/validate.py#L111-L113

means that the first thing the validation tool tries to do is read the full 215 GB X matrix into memory. I don't know the history of this choice. Is there no work around that will enable the h5ad to be read in with `backed='r'`?

bkmartinjr:
Fixes #399

Primary change: significantly reduce memory use when validating CSR-formatted data. The `X` and `raw.X` are read incrementally, in chunks, dramatically reducing memory usage. Dense data is run in backed mode, and should also see reduced memory use. CSC is still loaded into memory as before, and has no change.

Changes:
* revise Validator() to use backed mode when reading AnnData
* incrementally compute over `X` and `raw.X`
* reformat validate.py to conform with repo Python style configuration
* clean up debt in tests related to Pandas and AnnData

Old and new code was compared:
* there is a slight performance penalty (less than 10%) for using backed mode on small to mid-sized datasets
* large CSR datasets now complete without an OOM failure (or excessive swapping).  This was tested to datasets significantly exceeding the size of main memory (eg., a 500GiB+ H5AD on a 64GiB machine with no swap).

Notes for reviewers:
* some of the logic in sparse determination and nnz counting was potentially a bit off - eg., how NaN is handled, or some of the manual calculation of sparsity differentiating between explicitly and implicitly stored zeros. The PR leaves the logic identical to the previous version (or at least that is my intent).
* the vast majority of AnnData datasets are CSR

### DIFF
--- a/cellxgene_schema_cli/cellxgene_schema/validate.py
+++ b/cellxgene_schema_cli/cellxgene_schema/validate.py
@@ -391,12 +391,6 @@ class Validator:
 
         return
 
-    def _get_matrix_by_name(
-        self, matrix_name: str
-    ) -> Union[sparse.spmatrix, np.ndarray]:
-        matrix_getter = operator.attrgetter(matrix_name)
-        return matrix_getter(self.adata)
-
     def _get_matrix_format(self, matrix: Union[np.ndarray, sparse.spmatrix]) -> str:
         """
         Given a matrix, returns the format as one of: csc, csr, coo, dense
--- a/cellxgene_schema_cli/cellxgene_schema/validate.py
+++ b/cellxgene_schema_cli/cellxgene_schema/validate.py
@@ -1,6 +1,5 @@
 import logging
 import math
-import operator
 import re
 import sys
 from datetime import datetime
--- a/cellxgene_schema_cli/cellxgene_schema/validate.py
+++ b/cellxgene_schema_cli/cellxgene_schema/validate.py
@@ -1,4 +1,6 @@
 import logging
+import math
+import operator
 import re
 import sys
 from datetime import datetime
@@ -6,15 +8,10 @@ from datetime import datetime
 import anndata
 import os
 import pandas as pd
+import numpy as np
 from pandas.core.computation.ops import UndefinedVariableError
-from numpy import ndarray
-from numpy import count_nonzero
-from numpy import cumprod
-from numpy import isnan
-from numpy import nditer
-from numpy import vectorize
 from scipy import sparse
-from typing import List, Dict, Union, Tuple
+from typing import List, Dict, Union, Tuple, Optional
 from . import ontology
 from . import schema
 from . import env
@@ -30,7 +27,6 @@ class Validator:
     schema_definitions_dir = env.SCHEMA_DEFINITIONS_DIR
 
     def __init__(self, ignore_labels=False):
-
         # Set initial state
         self.errors = []
         self.warnings = []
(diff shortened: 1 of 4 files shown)


cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -392,10 +390,82 @@ def _validate_feature_id(self, feature_id: str, df_name: str):
...
+        # doing a slice seemed safer, if less performant.  Using `format_str`, which
+        # currently works, uses private API:
+        #
+        # >>> return getattr(matrix, "format_str", "dense)
+        #
+        format = "unknown"
+        if self.adata.n_obs == 0 or self.adata.n_vars == 0:
+            format = "dense"
+        else:
+            matrix_slice = matrix[0:1, 0:1]
```
ebezzi, line 412: Question: why do we need this slice to infer the type of the matrix?

cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -392,10 +390,82 @@ def _validate_feature_id(self, feature_id: str, df_name: str):
...
+        """
+
+        # Note: the AnnData proxy classes DO support the `format_str` property, but
+        # doing a slice seemed safer, if less performant.  Using `format_str`, which
+        # currently works, uses private API:
+        #
+        # >>> return getattr(matrix, "format_str", "dense)
+        #
+        format = "unknown"
+        if self.adata.n_obs == 0 or self.adata.n_vars == 0:
```
ebezzi, line 409: Why would this imply that the matrix is dense?

cellxgene_schema_cli/cellxgene_schema/validate.py
bkmartinjr: Because the type of X when you open in backed is a private AnnData-internal class which has no public API to introspect it (and the type itself is private). If we hard-wire the private (internal) type system or use other private interfaces, it is subject to breaking.

Using the public __getitem__ is guaranteed to return it in a form that supports public type introspection.

cellxgene_schema_cli/cellxgene_schema/validate.py
bkmartinjr: it implies it doesn't matter, so I randomly choose dense.  I don't think it matters what is returned....

ebezzi: Ah ok, I get it now. I believe such case should be rejected by the validator already, but since the check might happen later in the code, it makes sense 👍

bkmartinjr: From my reading of the code, the validator does not check elsewhere for an empty matrix (X or raw.X).  Caveat: I did a quick skim of the code.

ebezzi (approved):

Bento007 (approved):

atolopko-czi (commented):
Do we need a test update to ensure that the new chunking code is exercised?  (cellxgene_schema_cli/cellxgene_schema/validate.py#L439-L442). Just a few nits beyond that concern, otherwise LGTM.

Other's PR approvals are sufficient if you want to merge as is, so not approving outright.

Bento007:
I was running this in a limited memory machine and ran into memory limitation when writing labels. Did you https://github.com/chanzuckerberg/single-cell-curation/blob/37bef7bf457ebe3cbf2da79985c8d8f4573bc04e/cellxgene_schema_cli/cellxgene_schema/write_labels.py#L34

The machine had 28Gb of memory with a 47Gb dataset.

PR #413, opened 2023-03-22
feat: notebooks for GET /collections/{collection_id}/versions

danieljhegeman:


notebooks/curation_api/python/get_collection_versions.ipynb
```diff
@@ -0,0 +1,130 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "a6aab8e0",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "set_api_access_config(env=\"dev\")"
```
nayib-jose-gloria, line 85: nit: for the other notebooks, we've been keeping prod as the default arg. Not against doing dev instead, but just pointing out the divergence in case we want to switch over for all of them

nayib-jose-gloria (approved):
1 optional nit to consider, but looks good overall.

notebooks/curation_api/python/get_collection_versions.ipynb
danieljhegeman: thanks for catching this!!

PR #417, opened 2023-03-31
Don't copy the original anndata to write labels

### LINKED ISSUE #416: Support adding labels to larger datasets by reducing memory cost
## Motivation

In order to support larger h5ads we need to reduce the amount of memory required. The way AnnDataLabelAppender is currently written, it triples the amount of memory required. The explosion of memory required can be avoided if we eliminated making a copying of the anndata. 

## Definition of Done

Avoid make a copy of the original dataset in memory to process it with the AnnDataLabelAppender.

### Suggestion

Remove the `copy` and `to_memory` of the annData when initializing AnnDataLabelAppender. The validator which is passed in as a parameter to this class already has the anndata opened and loaded as read only. That means it will be read and modified without need to make a full copy. The required parts will be loaded as needed and written to a new file, leaving the original file intact. 
   
https://github.com/chanzuckerberg/single-cell-curation/blob/79a78b41414441d6f79ecf57baa57e3cea0c5eb3/cellxgene_schema_cli/cellxgene_schema/write_labels.py#L33-L36
changed to 
```
self.adata = validator.adata
```

This change should not affect the contents of the new file produced. 

## Tasks
- [ ] remove the copy of anndata into memory
- [ ] verify that both the original h5ad remains unchanged.
- [ ] verify that the labeled h5ad produced is the same file as before the changes were implemented. No regression.

## Notes
- comparing the hash of the h5ads is enough to ensure they are they same.

Bento007:
fixes #416 

- remove the copy of the anndata when writing labels. This is safe because the original dataset is open in read only mode. Any changes made to the datasets are only captured in memory. Those changes are then written out to a different files.
- add the stack trace to error collected when writing labels. This is to help debugging in the future.
- add missing tests
- add a test to verify the label h5ad is generated consistently. This is to verify that present and future changes don't accidentally change the labeled output file. 

# Testing
Ran validation and wrote labels with this [raw.h5ad](https://s3.console.aws.amazon.com/s3/buckets/corpora-data-dev?region=us-west-2&prefix=00a3fb7e-cd16-4ea1-b7b3-5e915061ffe2) and was able to produce a new h5ad that had the has the same sha256 hash as [local.h5ad](https://s3.console.aws.amazon.com/s3/buckets/corpora-data-dev?region=us-west-2&prefix=00a3fb7e-cd16-4ea1-b7b3-5e915061ffe2)

## script
```python
import logging
import hashlib

from cellxgene_schema.validate import validate

logging.basicConfig(level=logging.INFO)

raw_file = "../raw.h5ad"
original_labeled_file = "../local.h5ad"
new_labeled_file = "../new.h5ad"


def hash(file_name):
    with open(file_name, 'rb') as f:
        # Read the contents of the file in chunks
        chunk_size = 1024
        hasher = hashlib.sha256()
        while chunk := f.read(chunk_size):
            hasher.update(chunk)
    return hasher.hexdigest()

if __name__ == "__main__":
    validate(raw_file, new_labeled_file)

    original_hash = hash(original_labeled_file)
    new_hash = hash(new_labeled_file)
    raw_hash = hash(raw_file)

    assert original_hash == new_hash
    assert new_hash != raw_hash

    print(f"raw hash: {raw_hash}")
    print(f"org hash: {original_hash}")
    print(f"new hash: {new_hash}")
```

## output
```
INFO:cellxgene_schema.validate:Starting validation...
WARNING:cellxgene_schema.validate:WARNING: Dataframe 'var' only has 136 rows. Features SHOULD NOT be filtered from expression matrix.
INFO:cellxgene_schema.validate:Validation complete in 0:00:00.293064 with status is_valid=True
INFO:cellxgene_schema.write_labels:Writing labels
INFO:cellxgene_schema.validate:H5AD label writing complete in 0:00:00.297082, was_writing_successful: True
raw hash: 8ff3e4f5b595ef0d2a2fe4351ed4162b2dea65a1c38ea1adfaa0ba61eae9cdde
org hash: 55fbc095218a01cad33390f534d6690af0ecd6593f27d7cd4d26e91072ea8835
new hash: 55fbc095218a01cad33390f534d6690af0ecd6593f27d7cd4d26e91072ea8835
```

### DIFF
--- a/cellxgene_schema_cli/cellxgene_schema/write_labels.py
+++ b/cellxgene_schema_cli/cellxgene_schema/write_labels.py
@@ -382,7 +382,7 @@ class AnnDataLabelAppender:
         # Print errors if any
         if self.errors:
             for e, tb in self.errors:
-                logger.error(e, extra={"exec_info":tb})
+                logger.error(e, extra={"exec_info": tb})
             self.was_writing_successful = False
         else:
             self.was_writing_successful = True
--- a/cellxgene_schema_cli/cellxgene_schema/write_labels.py
+++ b/cellxgene_schema_cli/cellxgene_schema/write_labels.py
@@ -2,6 +2,7 @@ import logging
 import pandas as pd
 
 from typing import List, Dict, Optional
+import traceback
 
 from cellxgene_schema.validate import Validator, ONTOLOGY_CHECKER
 
@@ -30,11 +31,7 @@ class AnnDataLabelAppender:
                 "Validate AnnData first before attempting to write labels"
             )
 
-        if validator.adata.isbacked:
-            self.adata = validator.adata.to_memory()
-        else:
-            self.adata = validator.adata.copy()
-
+        self.adata = validator.adata
         self.validator = validator
         self.schema_def = validator.schema_def
         self.errors = []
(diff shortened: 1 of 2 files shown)


ebezzi (approved):
LGTM. We should add a comment that specifies that it's safe to avoid the copy since the original h5ad is only modified in memory and not on disk. I would also add a comment that this constraints the `add-label` step to be at the end.

PR #421, opened 2023-04-13
feat: update get_dataset_assets with updated workflow for fetching/downloading dataset assets

nayib-jose-gloria:
https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/single-cell-data-portal/4553

### DIFF
--- a/notebooks/curation_api/python/get_dataset_version.ipynb
+++ b/notebooks/curation_api/python/get_dataset_version.ipynb
@@ -47,7 +47,6 @@
     "    # download_assets,\n",
     "    get_dataset_version,\n",
     ")\n",
-    "\n",
     "from src.utils.config import set_api_access_config  # still required for setting api url env vars"
    ],
    "metadata": {


notebooks/compute_ancestor_mapping.ipynb
```diff
@@ -3,7 +3,11 @@
   {
    "cell_type": "markdown",
    "id": "89ea87d1-2459-4c09-b482-786db8cd3031",
-   "metadata": {},
+   "metadata": {
+    "pycharm": {
+     "name": "#%% md\n"
```
nayib-jose-gloria, line 8: GH pipeline was throwing a linting error until I removed an extra newline at the end...

nayib-jose-gloria: Feel free to ignore when reviewing, no real diff

danieljhegeman (commented):
see pr for this pr

notebooks/curation_api/R/get_dataset_assets_R.ipynb
```diff
@@ -3,25 +3,42 @@
...
+    }
+   },
    "source": [
     "The script in this notebook retrieves download links for a Dataset and then uses those links to download all available assets.\n",
     "\n",
-    "Fetching a Dataset requires only the Collection id and the Dataset id; it does not require an API key/access token."
+    "Fetching a Dataset requires only the Collection id and the Dataset id; it does not require an API key/access token.\n",
+    "\n",
+    "Note: This notebook guides download for assets of the current version of a dataset. For public collections, that is the most recently\n",
+    "published version of a dataset. For private collections, that is the most recently successfully processed dataset version. This notebook does not cover downloading previously published versions of a revised dataset.\n",
```
danieljhegeman, line 29: > This notebook does not cover downloading previously published versions...

I'm a little hesitant to have wording like this. The download process is identical for _datasets_ and _dataset_versions_. But this wording suggests that they are markedly different.

danieljhegeman: I think this portion should just be removed; leave the user to port the logic here to the dataset objects returned on the `.._versions/..` endpoints.

notebooks/curation_api/R/get_dataset_assets_R.ipynb
```diff
@@ -3,25 +3,42 @@
...
+   "metadata": {
+    "pycharm": {
+     "name": "#%% md\n"
+    }
+   },
    "source": [
     "The script in this notebook retrieves download links for a Dataset and then uses those links to download all available assets.\n",
     "\n",
-    "Fetching a Dataset requires only the Collection id and the Dataset id; it does not require an API key/access token."
+    "Fetching a Dataset requires only the Collection id and the Dataset id; it does not require an API key/access token.\n",
```
brianraymor, line 26: Fetching a dataset does NOT require a collection id.

notebooks/curation_api/R/get_dataset_assets_R.ipynb
brianraymor: I think that we need to be clear that a dataset url is a permanent url for **a** version of a dataset.

nayib-jose-gloria: I can remove that line, fair enough that it can be confusing. I do want to keep the rest since Brian wanted us to be clear with users that they are downloading a version of a dataset. So I want to be clear about what exactly they are downloading in the example scenario since it does slightly differ depending on whether the collection is public or private. 

> This notebook demonstrates downloading assets of the current version of a particular dataset. For public collections, that refers to the most recently published version of a dataset. For private collections, that refers to the most recently successfully processed dataset version.

danieljhegeman: That's fine!

danieljhegeman: Also agree with Brian's comment about specifying that it is a permalink--for that specific version.

notebooks/curation_api/python/src/dataset.py
```diff
@@ -57,51 +57,31 @@ def delete_dataset(collection_id: str, dataset_id: str):
         success(logger, success_message)
 
 
-def get_download_links_for_assets(collection_id: str, dataset_id: str):
+def download_assets(datasets: list, log_version_id: bool = False):
```
danieljhegeman, line 60: I know this is pedantic but I think this is a strictly worse method name. The method doesn't download the assets. It retrieves urls for the assets.

danieljhegeman: `get_asset_urls`?

danieljhegeman (changes requested):
one nit, then 💚

notebooks/curation_api/python/src/dataset.py
danieljhegeman: whoops

danieljhegeman (approved):
😄

PR #422, opened 2023-04-13
some changes

danieljhegeman:
- simplify target file download name
- remove usages of 'filename' which no longer exists on the asset response object


nayib-jose-gloria (approved):
good catch, ty!

PR #428, opened 2023-04-21
include x-request-id in failure response

### LINKED ISSUE #393: Feat(curation notebooks): display the x-request-id in the error response
When an error occurs especially 500 errors, the only piece of debug information the user has is the x-request-id. Making this value visible to the users will help developer trouble shoot errors seen by the users in the future.

Bento007:
fixes #393

### DIFF
--- a/notebooks/curation_api/python/src/utils/logger.py
+++ b/notebooks/curation_api/python/src/utils/logger.py
@@ -40,6 +40,8 @@ def failure(logger: logging.Logger, e, *messages):
     logger.error("\n\033[1m\033[38;5;9mFAILED\033[0m\n")  # 'FAILED' in bold red
     logger.error(e.response.reason + "\n")
     logger.error(e.response.text + "\n")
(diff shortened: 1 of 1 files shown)


notebooks/curation_api/python/src/utils/logger.py
```diff
@@ -40,6 +40,8 @@ def failure(logger: logging.Logger, e, *messages):
     logger.error("\n\033[1m\033[38;5;9mFAILED\033[0m\n")  # 'FAILED' in bold red
     logger.error(e.response.reason + "\n")
     logger.error(e.response.text + "\n")
+    logger.error("x-request-id: {}".format(e.response.headers.get("x-request-id")))
```
danieljhegeman, line 43:
```suggestion
    logger.error(f"x-request-id: {e.response.headers.get('x-request-id')}")
```

danieljhegeman (changes requested):
nit but let's remain consistent with the (newer) template string code pattern already established in the repo

danieljhegeman (approved):

Issue #373, opened 2022-11-03
post-ontology bump updates (3.0.0+1)


jahilton:
Collection `a48f5033-3438-4550-8574-cdff3263fdfd` - All 3 datasets all **assay: EFO:0700010**

Collection `A Proximal-to-Distal Survey of Healthy Adult Human Small Intestine and Colon Epithelium by Single-Cell Transcriptomics` - update CL:0000677 'gut absorptive cell' to **cell_type: CL:4030026** (@rachadele please review/confirm & update the title to the Collection ID once published)

Collection `edb893ee-4066-4128-9aec-5eb2b03f8287` - Both datasets update EFO:0010183 'single cell library construction' to **assay: EFO:0700011**

jlzamanian:
Collection: d2684035-a36e-458e-96af-8e37930bfdf6
Dataset: Integrated tendon dataset (only dataset)

Update the tissue ontology of all 4 samples to new term UBERON:8480009

[tendon of semitendinosus](https://www.ebi.ac.uk/ols/ontologies/uberon/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FUBERON_8480009&lang=en&viewMode=PreferredRoots&siblings=false) 

The update will go through Lattice.

EMRutherford:
Collection:  https://cellxgene.cziscience.com/collections/1d1c7275-476a-49e2-9022-ad1b1c793594
Dataset:  Amacrine cells of the human fovea and peripheral retina

Replace CL:0000561 (amacrine cell) with CL:4030027 (GABAergic amacrine cell).

EMRutherford:
Collection:  Integration of eQTL and a Single-Cell Atlas in the Human Eye Identifies Causal Genes for Age-Related Macular Degeneration
Dataset:  Amacrine cells of human eye

For the following group IDs in the processed matrix files:
1.0 - Replace CL:0000561 (amacrine cell) with CL:0004219 (A2 amacrine cell)
5.0 - Replace CL:0000561 (amacrine cell) with CL:4030028 (glycinergic amacrine cell)
11.0 - Replace CL:0000561 (amacrine cell) with CL:0004232 (starburst amacrine cell)

EMRutherford:
Collection: Integration of eQTL and a Single-Cell Atlas in the Human Eye Identifies Causal Genes for Age-Related Macular Degeneration

Dataset:  Retinal ganglion cells of human eye

[...]

PR #430, opened 2023-04-25
chore: proposed 'convert' script Template [DO NOT MERGE]

nayib-jose-gloria:
- Proposed template for 'convert' script 
- Concept: During a schema version migration, the 'convert' command will be called against every public dataset in the corpus. It will encompass both general changes that must apply to every dataset in the corpus, and bespoke changes to specific collections/datasets. 
- An automated process will pre-populate a 'convert' command with any deterministic ontology term replacements + some necessary boilerplate and commented instructions. 
- Curators will then add to this PR (if necessary) to include non-deterministic changes like those tracked [here](https://github.com/chanzuckerberg/single-cell-curation/issues/373) or those flagged as needing Curator intervention in the proposed monthly ontology update deprecated terms report. 
- Once ready, Curator + On-call Engineer will approve PR, and the On-Call Engineer will merge + cut a new cellxgene-schema release and kick off an automated corpus migration. 

- The changes to the cli and utils files are one-time changes, not part of the proposed monthly automation
- We can discuss whether there are more useful 'util' functions to add for common dataframe transformation patterns; so far I only included one that I found useful. 

NOT INCLUDED in this demo but part of the proposed workflow: 
- updated ontology pinned files 
- updated downloaded ontologies

### DIFF
--- a/cellxgene_schema_cli/cellxgene_schema/cli.py
+++ b/cellxgene_schema_cli/cellxgene_schema/cli.py
@@ -80,16 +80,18 @@ def remove_labels(input_file, output_file):
 
 @click.command(
     name="convert",
-    short_help="Convert an h5ad from version 2.0.0 to version 3.0.0",
-    help="Convert an h5ad from version 2.0.0 to version 3.0.0. No validation will be performed on either"
+    short_help="Convert an h5ad to the latest schema version.",
+    help="Convert an h5ad to the latest schema version. No validation will be performed on either"
     "the input or the output file.",
 )
 @click.argument("input_file", nargs=1, type=click.Path(exists=True, dir_okay=False))
 @click.argument("output_file", nargs=1, type=click.Path(exists=False, dir_okay=False))
-def convert(input_file, output_file):
+@click.option("--collection_id", default=None, type=str, help="Collection ID, if converting already published dataset")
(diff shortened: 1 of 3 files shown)


jahilton:
This looks good. Super straightforward on how to identify observations that need a field changed, which field is being changed, and what the new value should be
I think my only question is how to dummy-proof it. How can we validate the stuff that we fill in such that we don't cause an error that stops the whole thing, we don't have a typo that results in 0 observations being updated, etc. Or are we just relying on post-migration QA to catch things (which won't be so bad if it's only 1 month's worth of updates)

nayib-jose-gloria:
> This looks good. Super straightforward on how to identify observations that need a field changed, which field is being changed, and what the new value should be I think my only question is how to dummy-proof it. How can we validate the stuff that we fill in such that we don't cause an error that stops the whole thing, we don't have a typo that results in 0 observations being updated, etc. Or are we just relying on post-migration QA to catch things (which won't be so bad if it's only 1 month's worth of updates)

It's a fair concern--I'm going to do some thinking on options for automating test coverage and get back to this

jahilton (approved):
Looks great. Just warrants some consideration of how curators can use the convert script without breaking things

PR #436, opened 2023-05-09
feat(): Add version tracking for Gencode releases

Bento007:
Related to [#4813](https://github.com/chanzuckerberg/single-cell-data-portal/issues/4813)

- Add *gene_info.yml* to track which Gencode release is used
- Add documentation and process on how to update and modify *gene_info.yml*. Relates to https://github.com/chanzuckerberg/single-cell-data-portal/issues/4744
- Remove unused make recipes
- Download and process gene info in parallel. This speeds up the process especially when downloading the files.
- Download gene files to a temporary directory this is deleted automatically
- Add github actions to process genes when *gene_info.yml* changes
- Move ontology-processing github action to it's own file. This is to limit the number of times the ontology files are rebuilt. Before this change the ontology file would be rebuilt every time a push to a branch with change to any of the ontology info files was made. This resulted in rebuilding the file numerous time when no additional changes occurred. Now the ontology files will only be regenerated when there is a commit that changes them. Future commits to the same branch will not result in regenerating the ontology files unless the ontologies sources changed.
- Unit tests are run when ontology files are regenerated.

# Notes
- here you can see the GHA succeeds https://github.com/chanzuckerberg/single-cell-curation/actions/runs/4930521537/jobs/8811531780


cellxgene_schema_cli/cellxgene_schema/ontology_files/gene_info.md
```diff
@@ -0,0 +1,31 @@
+This file describes the expected format of *[gene_info.yml](./gene_info.yml)*. It is used to track is used to track the
```
nayib-jose-gloria, line 1: accidentally duplicated start of sentence

nayib-jose-gloria (changes requested):
It looks like the GHA workflow file to run this job was deleted in the latest commit--was that a mistake?

Bento007:
That we me trying to clean up my git history.

.github/workflows/gene_update.yml
```diff
@@ -0,0 +1,37 @@
...
+        run: |
+          pip install -r cellxgene_schema_cli/requirements.txt
+          make gene-processing -C cellxgene_schema_cli
+      - name: Commit
+        run: |
+          git config user.name github-actions
+          git config user.email github-actions@github.com
+          git add .
+          git commit -m "AUTO: update genes"
+          git push
```
nayib-jose-gloria, line 37: Same note as the other PR--push tests don't run in the PR after this commit, which could hide failing tests

nayib-jose-gloria (commented):
Changes look good! Just missing a step to re-run tests after auto-commit

nayib-jose-gloria (commented):
very nit: I'm thinking we should rename push tests, since we're doing more than just testing. push_automation? push_jobs?

nayib-jose-gloria (approved):

PR #437, opened 2023-05-09
feat: Script to Produce Curator Report of Deprecated Terms in Corpus After Ontology Version Bump

nayib-jose-gloria:
Changes:
- Script to Produce Curator Report of Deprecated Terms in Corpus, pulling from all_ontologies file processing pinned ontologies 
- Updated ontology processing that creates all_ontologies file to include required fields for producing curator report (note: new stored info can also potentially be leveraged to transform terms to their replacement terms automatically, if available--but that is out of scope for this PR)
- Updated current all_ontologies file to pick up new fields ('replaced_by' for terms that are marked as deprecated, and 'considers' + 'comments' if replaced_by is not included)
- GHA to trigger script + upload curator report of deprecated terms to github artifacts

Notes for Reviewer:
- Currently only producing a curator report for public collections. This will require super curator auth privileges in our GHA, which we need to determine if we want to do. Might need to consider a different solution, or reconsider having private collections involved. Thinking I'll break that work into a different PR since it may be considerable work/discovery--thoughts?
- Have yet to update the pinned ontology files, so the workflow should currently produce an empty curator report (since no terms currently in the corpus are deprecated, as we validate before allowing upload). 
          - I have tested locally with new ontology versions pinned in the owl_info, and it works!


schema_bump_dry_run_scripts/ontologies/ontology_bump_dry_run.py
```diff
@@ -0,0 +1,72 @@
...
+    "prod": "https://api.cellxgene.cziscience.com",
+    "staging": "https://api.cellxgene.staging.single-cell.czi.technology",
+    "dev": "https://api.cellxgene.dev.single-cell.czi.technology",
+}
+
+
+def dry_run():
+    # Load processed ontologies file
+    ontologies = "cellxgene_schema_cli/cellxgene_schema/ontology_files/all_ontology.json.gz"
+    with gzip.open(ontologies, "r") as f:
```
Bento007, line 18: If you open in "rt" mode you don't need to decode.

schema_bump_dry_run_scripts/ontologies/ontology_bump_dry_run.py
```diff
@@ -0,0 +1,72 @@
...
+        "development_stage",
+        "disease",
+        "organism",
+        "self_reported_ethnicity",
+        "sex",
+        "tissue",
+    }
+    # cache terms we know are not deprecated to skip processing; init with special-case, non-ontology terms we use
+    non_deprecated_term_set = {"multiethnic", "unknown", "na"}
+    f = open("ontologies-curator-report.txt", "w")
```
Bento007, line 38: is there a reason you aren't using a context manager for this open file? Best practice is to us a context manager in part because it will close the file even if an except occurs within the context block.

schema_bump_dry_run_scripts/ontologies/requirements.txt
```diff
@@ -0,0 +1 @@
+requests
```
Bento007, line 1: pin

Bento007 (changes requested):
Why did the size of all_ontology.json.gz change?

nayib-jose-gloria:
> Why did the size of all_ontology.json.gz change?

I ran the updated ontology processing script to add replaced_by, consider, and comments for deprecated terms, which we need for the dry run script

Bento007 (approved):

PR #438, opened 2023-05-09
feat(): adding GHA to run download-ontologies make recipe when owl_info.yml changes

Bento007:
relates to https://github.com/chanzuckerberg/single-cell-data-portal/issues/4780

Here is the job running https://github.com/chanzuckerberg/single-cell-curation/actions/runs/4930636257/jobs/8811783833

This is what the commit looks like:
Screenshot 2023-05-09 at 2 50 17 PM


Bento007:
The github action took 20min to complete, are we ok with that?

.github/workflows/ontology_update.yml
```diff
@@ -0,0 +1,37 @@
+name: Updated Ontologies
+
+on:
+  pull_request:
+    paths:
+      - '**/cellxgene_schema_cli/cellxgene_schema/ontology_files/owl_info.yml'
+
+jobs:
+  gene-processing:
```
nayib-jose-gloria, line 9: nit: rename to ontologies-processing

nayib-jose-gloria:
> The github action took 20min to complete, are we ok with that?

We should confirm with Brian, but I think that's fine. That said, I think we can get away with a smaller requirements.txt--anndata install can take a while, for instance, and its required for other parts of the cli but not ontology processing.

.github/workflows/ontology_update.yml
```diff
@@ -0,0 +1,37 @@
...
+        run: |
+          pip install -r cellxgene_schema_cli/requirements.txt
+          make download-ontologies -C cellxgene_schema_cli
+      - name: Commit
+        run: |
+          git config user.name github-actions
+          git config user.email github-actions@github.com
+          git add ./cellxgene_schema_cli/cellxgene_schema/ontology_files/all_ontology.json.gz
+          git commit -m "AUTO: update ontologies"
+          git push
```
nayib-jose-gloria, line 37: I noticed the push tests don't run after this commit--I'm not sure why though, the configuration of push_tests looks like it should trigger. We do have some tests around the 'validate' command that we should run after an ontology update

nayib-jose-gloria (changes requested):
A couple of things to look into, but this largely looks great--thanks!

.github/workflows/ontology_update.yml
Bento007: It only runs when a change is detected in the owl_info.yml. If you push another commit to the same branch that doesn't change that file, then the GHA will not run.

cellxgene_schema_cli/cellxgene_schema/ontology_files/all_ontology.json.gz
nayib-jose-gloria, line 1: do you know why the size decreased?

.github/workflows/ontology_update.yml
Bento007: github docs https://docs.github.com/en/actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow

nayib-jose-gloria:
> > The github action took 20min to complete, are we ok with that?
> 
> We should confirm with Brian, but I think that's fine. That said, I think we can get away with a smaller requirements.txt--anndata install can take a while, for instance, and its required for other parts of the cli but not ontology processing.

I ran locally and it takes about 8 minutes though, so it does seem like a weirdly large timing diff that is not accounted for just by installing too many packages. Maybe worth looking into, and timeboxing for an hour? Since we can tolerate 20 min but it would be nicer if we can easily speed it up

nayib-jose-gloria:
> > > The github action took 20min to complete, are we ok with that?
> > 
> > 
> > We should confirm with Brian, but I think that's fine. That said, I think we can get away with a smaller requirements.txt--anndata install can take a while, for instance, and its required for other parts of the cli but not ontology processing.
> 
> I ran locally and it takes about 8 minutes though, so it does seem like a weirdly large timing diff that is not accounted for just by installing too many packages. Maybe worth looking into, and timeboxing for an hour? Since we can tolerate 20 min but it would be nicer if we can easily speed it up

or just make a low pri issue to follow up on it after if we have time

nayib-jose-gloria (commented):
Changes look good! Just missing a step to re-run tests after auto-commit + explaining why the all_ontology json size is changing

Bento007:
So i changed up the push-test to check for ontology updated before running. Updated are applied if needed and the tests are run against the new ontology changes.

In this [GHA](https://github.com/chanzuckerberg/single-cell-curation/actions/runs/4939914652/jobs/8831100585) under `ontology-processing` you can see that we update the ontology files and commit the changes `9a76728..fa97dfc  tsmith/owl-4780 -> tsmith/owl-4780`. Then in `unit-test` and `lint` it pulls the new changes `fa97dfc` under checkout.

nayib-jose-gloria (approved):
thanks!

PR #439, opened 2023-05-10
Remove duplicate jobs from GHA

Bento007:
The test run in python-app.yml are already run in push-test.yml. There is no reason to duplicate work.


nayib-jose-gloria (approved):
good catch

PR #441, opened 2023-05-11
feat(ontology upgrade dry-run): Report Deprecated Terms in Private Datasets + Report Collections With Open Revisions

nayib-jose-gloria:
https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/single-cell-data-portal/4825

Changes:
- Added AWS auth to github actions workflow for dry-runs, fetch auth0 secrets from AWS
- auth as super curator machine account, and fetch private datasets to check for deprecated terms 
- also report public collections with open revisions, since they will not be auto-migrated
- Some refactoring into helper functions in dry-run script

Notes for Reviewer:
- Currently, implemented private dataset fetching by fetching private Collections index endpoint --> making N follow up requests for private datasets. This is because there is no private dataset index endpoint, and the private Collections index endpoint response does not contain all the dataset ontology terms. Future optimization to avoid the N additional API Calls would be to introduce either the missing metadata to the private Collection index endpoint, or create an API path for fetching all private datasets 

See dry-run [output](https://github.com/chanzuckerberg/single-cell-curation/actions/runs/4951157319?pr=441) (no diff besides reporting public collection IDs with open revisions because pinned ontologies have not been updated)


.github/workflows/ontology_bump_dry_run.yml
```diff
@@ -2,9 +2,17 @@ name: Ontology Bump Dry Run
 
 on:
   push:
-    branches: main
+    branches:
+      - main
+      - nayib/dry-run-with-private-collections # TODO: temp while testing, remove before merge
     paths:
       - '**/cellxgene_schema_cli/cellxgene_schema/ontology_files/all_ontology.json.gz'
+      - '**/.github/workflows/ontology_bump_dry_run.yml' # TODO: temp while testing, remove before merge
```
Bento007, line 10: I should have done this with my GHA PR. It would have made testing easier.

schema_bump_dry_run_scripts/ontologies/ontology_bump_dry_run.py
```diff
@@ -1,72 +1,131 @@
...
+    "self_reported_ethnicity",
+    "sex",
+    "tissue",
 }
 
 
+def get_headers():
+    claims = "openid profile email offline read:collections write:collections"
+    auth0_secrets = json.loads(os.getenv('AUTH0_SECRETS'))
+    response = requests.post(
```
Bento007, line 30: This auth flow is blocked on prod to prevent password brute force attacks. You need to generated an API key and use the curation API.

Bento007 (changes requested):
Please link the terraform changes to this PR as well. 

There is a lot of if logic. It would be helpful to have some unit-test to verify all of the different paths.

schema_bump_dry_run_scripts/ontologies/ontology_bump_dry_run.py
nayib-jose-gloria: What's the best way to do that programmatically? I'm assuming the API Keys eventually expire

Bento007: Looks good

.github/workflows/push_tests.yml
```diff
@@ -41,7 +41,39 @@ jobs:
...
+        run: |
+          pip install -r schema_bump_dry_run_scripts/ontologies/requirements.txt
+          pip install pytest
+      - name: ontology dry run changes
+        uses: dorny/paths-filter@v2
+        id: filter
+        with:
+          filters: |
+            ontology_dry_run:
+              - 'schema_bump_dry_run_scripts/**'
```
Bento007, line 75: Since this job only needs to run when these files change, I recommend putting this job in it's own GHA file, that way it only runs when those files change. Otherwise we will run through the whole setup for every PR.

Bento007 (changes requested):

.github/workflows/ontology_bump_dry_run.yml
```diff
@@ -2,10 +2,15 @@ name: Ontology Bump Dry Run
 
 on:
   push:
-    branches: main
+    branches:
+      - main
```
Bento007, line 6: should this also run on branches?

Bento007 (approved):

PR #447, opened 2023-05-15
feat: generate diff between gencode releases when processing new GENCODE files

nayib-jose-gloria:
https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/single-cell-data-portal/4814

- Output diff file will live where processed ontology files live. The 'migrate' script that will automate the 'Lattice Algorithm' for removing deprecated terms will be part of the cellxgene_schema_cli package as well.
- Only accounts for gene IDs that were 'deprecated' (i.e. ID appeared previously, no longer appears in new reference files). It does not detect IDs that changed

### DIFF
--- a/cellxgene_schema_cli/scripts/gene_processing.py
+++ b/cellxgene_schema_cli/scripts/gene_processing.py
@@ -1,3 +1,4 @@
+import csv
 import gzip
 import os
 import sys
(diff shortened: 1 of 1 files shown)


cellxgene_schema_cli/scripts/gene_processing.py
```diff
@@ -169,11 +170,43 @@ def process_gene_info(gene_info: dict) -> None:
...
+            gene_id = row[0]
+            if gene_id not in new_ref_gene_ids:
+                removed_gene_ids.append(gene_id)
+
+    diff_filepath = os.path.join(env.ONTOLOGY_DIR, f"{output_filename}_diff.txt")
+    with open(diff_filepath, "w") as f:
+        for gene_id in removed_gene_ids:
+            f.write(f"{gene_id}\n")
+
+    os.remove(previous_ref_filepath)
```
Bento007, line 207: This should live in `process_gene_info`. Deleting the file at the same level as creating the file make more sense than hiding it in a nested function. You could even us a temp file with a context manager in `process_gene_info`. That will ensure things get cleaned up nicely.

Bento007 (changes requested):

cellxgene_schema_cli/scripts/gene_processing.py
```diff
@@ -169,11 +170,43 @@ def process_gene_info(gene_info: dict) -> None:
     """
     print("download", gene_info["description"])
     temp_file_path, _ = urllib.request.urlretrieve(gene_info["url"])
+    previous_ref_filepath = os.path.join(env.ONTOLOGY_DIR, f"previous_{gene_info['description']}.csv.gz")
+    os.rename(gene_info["new_file"], previous_ref_filepath)
```
Bento007, line 174: can you add a `try-except` block to restore the old file is something fails during this process. I'm imaging something going wrong and how much work it would be to go back and rebuild so we can run the diff again.

cellxgene_schema_cli/scripts/gene_processing.py
```diff
@@ -169,11 +170,50 @@ def process_gene_info(gene_info: dict) -> None:
     """
     print("download", gene_info["description"])
     temp_file_path, _ = urllib.request.urlretrieve(gene_info["url"])
-    print("process", gene_info["description"])
-    gene_info["processor"](temp_file_path, gene_info["new_file"])
+    previous_ref_filepath = os.path.join(env.ONTOLOGY_DIR, f"previous_{gene_info['description']}.csv.gz")
+    try:
+        # temporarily backup previous processed csv
+        os.rename(gene_info["new_file"], previous_ref_filepath)
+        print("process", gene_info["description"])
```
Bento007, line 177: is this poor naming going to come back to haunt us?

nayib-jose-gloria: What poor naming? I'm not opposed to renaming anything here for clarity

PR #448, opened 2023-05-16
feat: update curation report to group reporting deprecated terms in datasets from the same collection

nayib-jose-gloria:
- Decompose parsing into 1) a mapping function that runs the logic to detect deprecated terms + maps reporting data by collection ID and deprecated term, and 2) a function that iterates over the map and writes to the output file in the desired format
- Update writing logic to remove Dataset ID + replace with count of affected datasets in Collection for the detected deprecated term


Bento007 (changes requested):
Since these files will not likely be updated frequently we should add good documentations to ease future changes/debugging.

Bento007 (changes requested):
Since these files will not likely be updated frequently we should add good documentations to ease future changes/debugging.

Bento007 (approved):

PR #449, opened 2023-05-18
feat(genes): generate dry-run report for depreciated genes

Bento007:
related to https://github.com/chanzuckerberg/single-cell-data-portal/pull/issues/4817

## Changes
- Add GHA to run gene dry run and send results to slack
- Add GHA to run test on gene dry run code when it changes
- combine GHA for ontology and gene dry run tests
- Add make recipe to run gene dry run tests. This makes it easier to run consistently locally and in GHA.
- Add code to generate a report with deprecated genes.


schema_bump_dry_run_scripts/genes/gene_bump_dry_run.py
```diff
@@ -0,0 +1,212 @@
...
+    files = os.listdir(ONTOLOGY_DIR)
+    for file in files:
+        if file.endswith(suffix):
+            with open(f"{ONTOLOGY_DIR}/{file}") as fp:
+                organism = getattr(SupportedOrganisms, file.removesuffix(suffix).upper()).value
+                diff_map[organism] = fp.read().splitlines()
+    return diff_map
+
+
+def fetch_private_datasets(base_url) -> Tuple[List[dict], Optional[str]]:
```
Bento007, line 94: @nayib-jose-gloria

schema_bump_dry_run_scripts/genes/gene_bump_dry_run.py
```diff
@@ -0,0 +1,212 @@
...
+{{ collection }}
+{% endfor %}
+"""
+
+    j2_template = Template(report, trim_blocks=True, lstrip_blocks=True)
+    report = j2_template.render(data)
+    return report
+
+
+def get_genes(dataset: dict, stage: str) -> List[str]:
```
nayib-jose-gloria, line 59: this is the logic came up with to be compatible with the upcoming cxg processing change, FYI:

```
    with tiledb.open(s3_path, "r") as var:
        var_df = var.df[:]
        suffix = 0
        while f"name_{suffix}" not in var_df.columns:
            suffix += 1
        index_name = "name_{suffix}"
        stored_genes = var_df[index_name].to_list()
```

.github/workflows/gene_bump_dry_run.yml
```diff
@@ -0,0 +1,55 @@
...
+        uses: aws-actions/aws-secretsmanager-get-secrets@v1
+        with:
+          # TODO: Change to prod and make separate, manual dev workflow once ready for deploy
+          secret-ids:
+            AUTH0_SECRETS, corpora/backend/dev/auth0-secret
+      - name: Run Dry-Run Script
+        run: python3 -m schema_bump_dry_run_scripts.genes.gene_bump_dry_run
+        env:
+          corpus_env: "dev" # TODO: Change to prod and make separate, manual dev workflow once ready for deploy
+      # TODO: temp upload while waiting to integrate with Slack
```
nayib-jose-gloria, line 50: nit: can remove this TODO now

.github/workflows/schema_bump_dry_run_tests.yml
```diff
@@ -0,0 +1,56 @@
...
+      - name: Python cache
+        uses: actions/cache@v1
+        with:
+          path: ~/.cache/pip
+          key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements*.txt') }}
+          restore-keys: |
+            ${{ runner.os }}-pip-
+      - name: Install dependencies
+        run: |
+          pip install -r schema_bump_dry_run_scripts/genes/requirements.txt
```
nayib-jose-gloria, line 30: change to `schema_bump_dry_run_scripts/ontologies/requirements.txt`

.github/workflows/schema_bump_dry_run_tests.yml
```diff
@@ -0,0 +1,56 @@
...
+
+  genes-dry-run-tests:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Set up Python 3.8
+        uses: actions/setup-python@v1
+        with:
+          python-version: 3.8
```
nayib-jose-gloria, line 43: removesuffix, used in the dry run, was introduced in Python 3.9 :)

.github/workflows/gene_bump_dry_run.yml
```diff
@@ -0,0 +1,55 @@
...
+jobs:
+  dry-run:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Set up Python 3.8
+        uses: actions/setup-python@v1
+        with:
+          python-version: 3.8
```
nayib-jose-gloria, line 23: duplicate note: removesuffix, used in the dry run, was introduced in Python 3.9

schema_bump_dry_run_scripts/genes/gene_bump_dry_run.py
```diff
@@ -0,0 +1,221 @@
...
+def get_diff_map() -> Dict[str, List[str]]:
+    # list all of the files ending with diff.txt in the cellxgene_schema/ontology_files directory
+    # for each file, open it and read the contents into a dictionary
+    diff_map = {}
+    suffix = "_diff.txt"
+    files = os.listdir(ONTOLOGY_DIR)
+    for file in files:
+        if file.endswith(suffix):
+            with open(f"{ONTOLOGY_DIR}/{file}") as fp:
+                organism = getattr(SupportedOrganisms, file.removesuffix(suffix).upper()).value
```
nayib-jose-gloria, line 75: I think we need to remove the "gene_" prefix as well

.github/workflows/gene_bump_dry_run.yml
Bento007: I don't understand this comment

schema_bump_dry_run_scripts/genes/gene_bump_dry_run.py
```diff
@@ -0,0 +1,223 @@
...
```
nayib-jose-gloria, line 76: This is the comment I'm referring to--the push tests are failing because GHA is set to python 3.8, but this function was introduced in python 3.9

schema_bump_dry_run_scripts/genes/gene_bump_dry_run.py
```diff
@@ -0,0 +1,224 @@
...
+    deprecated_genes_in_dataset = set()
+    for organism in organisms:
+        deprecated_genes_source = diff_map.get(organism, set())
+        intersection_genes = dataset_genes_to_compare.intersection(deprecated_genes_source)
+        if intersection_genes:
+            deprecated_genes_in_dataset.update(intersection_genes)
+            is_deprecated_genes_found = True
+
+    if is_deprecated_genes_found:
+        dataset_group = (*sorted(deprecated_genes_in_dataset), len(dataset_genes_to_compare))
```
nayib-jose-gloria, line 131: Mind adding a comment or adding to the docstring explaining the purpose of this block?

schema_bump_dry_run_scripts/genes/gene_bump_dry_run.py
```diff
@@ -0,0 +1,227 @@
...
+        collection = deprecated_datasets[collection_id]
+        if dataset_group_key not in collection["dataset_groups"]:
+            collection["dataset_groups"][dataset_group_key] = {
+                "datasets": [],
+                "num_datasets": 0,
+                "deprecated_genes": deprecated_genes_in_dataset,
+                "num_genes": len(dataset_genes_to_compare),
+            }
+        group = collection["dataset_groups"][dataset_group_key]
+        group["num_datasets"] += 1
```
nayib-jose-gloria, line 145: nit: do we need this if we can call len(datasets)?

schema_bump_dry_run_scripts/genes/gene_bump_dry_run.py
```diff
@@ -0,0 +1,227 @@
...
+        # dataset_group_key is a unique key used to identify a group of datasets that have the same set of deprecated
+        # genes and the same number of genes. This is used to group datasets together in the report.
+        if collection_id not in deprecated_datasets:
+            deprecated_datasets[collection_id] = {"dataset_groups": {}}
+        collection = deprecated_datasets[collection_id]
+        if dataset_group_key not in collection["dataset_groups"]:
+            collection["dataset_groups"][dataset_group_key] = {
+                "datasets": [],
+                "num_datasets": 0,
+                "deprecated_genes": deprecated_genes_in_dataset,
```
nayib-jose-gloria, line 141: nit: do we need these if we can derive them from the dataset_group_key?

nayib-jose-gloria: these == 'deprecated genes' and 'num_genes'

schema_bump_dry_run_scripts/genes/gene_bump_dry_run.py
```diff
@@ -0,0 +1,227 @@
...
+        if intersection_genes:
+            deprecated_genes_in_dataset.update(intersection_genes)
+            is_deprecated_genes_found = True
+
+    if is_deprecated_genes_found:
+        dataset_group_key = (*sorted(deprecated_genes_in_dataset), len(dataset_genes_to_compare))
+        # dataset_group_key is a unique key used to identify a group of datasets that have the same set of deprecated
+        # genes and the same number of genes. This is used to group datasets together in the report.
+        if collection_id not in deprecated_datasets:
+            deprecated_datasets[collection_id] = {"dataset_groups": {}}
```
nayib-jose-gloria, line 135: nit: if 'dataset_groups' is the only key at this level, can we remove it so we can just reference `deprecated_datasets[collection_id][dataset_group_key]` directly?

schema_bump_dry_run_scripts/genes/gene_bump_dry_run.py
```diff
@@ -0,0 +1,227 @@
...
+    :param base_url: The base URL for fetching private datasets.
+    :type base_url: str
+    :param diff_map: The map of organisms and their deprecated genes.
+    :type diff_map: dict
+    :return: A tuple containing the dictionary of collections with datasets containing deprecated genes and the list
+        of non-auto-migrated public collections.
+    :rtype: tuple
+    """
+    private_deprecated = dict()
+    non_auto_migrated = []
```
nayib-jose-gloria, line 189: nit: make set() for constant time lookup of `revision_of not in non_auto_migrated `

nayib-jose-gloria (commented):
A few nits that you can optionally address, but this looks good overall--thanks!

schema_bump_dry_run_scripts/genes/gene_bump_dry_run.py
Bento007: Calling len(dataset) for every dataset would be more expensive than incrementing this counter, but if we eliminate the need for a dictionary look up it could be faster, but these are all minor optimizations. I would prefer to keep as much of the logic for creating the dataset group in this functions to simplify testing.

schema_bump_dry_run_scripts/genes/gene_bump_dry_run.py
Bento007: > nit: do we need these if we can derive them from the dataset_group_key?
yes we could, but for readability I treated them as different things. If we did use the dataset_group_key we would need to parse the tuple as `dataset_group_key[:-1]` which makes less sense than using `deprecated_genes_in_dataset` which has already be calculated and stored. This would al require moving more processing logic into the parent function which would be more code duplication.

Bento007: > these == 'deprecated genes' and 'num_genes'
`num_genes` is the number of genes in a dataset, not just deprecated. This was a specific request from to track these values.

schema_bump_dry_run_scripts/genes/gene_bump_dry_run.py
Bento007: It's like this to accommodate the public and private dataset code paths for `gene_compare`. I could put a wrapper around it to obfuscate that details, but that would only be for readability.

nayib-jose-gloria (approved):

PR #462, opened 2023-05-30
feat: release cellxgene-schema to test pypi

nayib-jose-gloria:
- Set up to release new minor version to Test PyPI and PyPI when there are ontology/GENCODE reference file bumps in main
- For now, setting it to release as a release candidate for testing, and only to test PyPI. Will change once ready to bump in prod
- Added Test PyPI / PyPI API keys to Github secrets
- Test PyPI 3.1.0rc0 published release: https://test.pypi.org/project/cellxgene-schema/3.1.0rc0/
- Action triggering release above: https://github.com/chanzuckerberg/single-cell-curation/actions/runs/5125586931

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 3.0.2-rc.0
+current_version = 3.0.3-rc.0
 commit = True
 parse = (?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)(?:-(?P<prerel>rc)\.(?P<prerelversion>\d+))?
 serialize = 
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 3.3.0-rc.0
+current_version = 3.3.0
 commit = True
 parse = (?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)(?:-(?P<prerel>rc)\.(?P<prerelversion>\d+))?
 serialize = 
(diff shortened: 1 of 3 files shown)


.bumpversion.cfg
```diff
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 3.0.2-rc.0
+current_version = 3.0.0-rc.1
```
Bento007, line 2: Why did this reset back to 3.0.0?

.github/workflows/publish-to-pypi.yml
```diff
@@ -0,0 +1,46 @@
...
+      - name: Build dist
+        run: >-
+          make pydist
+      - name: Publish distribution to Test PyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          password: ${{ secrets.TEST_PYPI_API_TOKEN }}
+          repository-url: https://test.pypi.org/legacy/
+          packages-dir: cellxgene_schema_cli/dist/
+# TODO: Uncomment once ready to release to prod PyPI
```
Bento007, line 40: It feels odd to release to both pypi and testpypi at the same time. We shouldn't test that the RC works before release to pypi. We should install it using testpypi and run a few tests to confirm it works.

.bumpversion.cfg
nayib-jose-gloria: This was an artifact of testing, I'll fix

.github/workflows/publish-to-pypi.yml
nayib-jose-gloria: Unit tests already run against a local build from the same dist files we push. What other tests do you have in mind?

Bento007: download from testpipy and make sure it runs correctly.

Bento007 (approved):

Issue #440, opened 2023-05-10
Portal fills in schema_version


jahilton:
# Design

## `uns` (Dataset Metadata)

`uns` is a ordered dictionary with a `str` key.

...

When a dataset is uploaded, CELLxGENE Discover MUST automatically add the `schema_version` key and its value to `.uns`. If `schema_version` is already defined, then its value MUST be overwritten. 

### schema_version

<table><tbody>
    <tr>
      <th>Key</th>
      <td>schema_version</td>
    </tr>
    <tr>
      <th>Annotator</th>
      <td>CELLxGENE Discover</td>
    </tr>
    <tr>
      <th>Value</th>
        <td>
          This MUST be <code>"3.1.0"</code>.
        </td>
    </tr>
</tbody></table>
<br>

# Context
Currently, [uns.schema_version](https://github.com/chanzuckerberg/single-cell-curation/blob/main/schema/3.0.0/schema.md#schema_version) is required and it is required to match the current schema version.
The validator will only ever validate a single schema version (the current one) so there is no desire to allow for any value other than the current version.
Upcoming plans to change the schema_version on a ~monthly cadence for ontology bumps will result in curators needing to more readily track the current version. (I foresee a lot of validation failing _just_ on uns.schema_version being a step or two or behind)

# Proposal
uns.schema_version Annotator changed from Curator to Data Portal
Upon submission, uns.schema_version is filled in with the current schema version
If uns.schema_version is already defined and it does not match the current version, a warning is produced specifying that the validator is validating against a different version.
For validation, the defined uns.schema_version is otherwise ignored. Upon submission, the defined uns.schema_version will be overwritten.

PR #463, opened 2023-05-31
feat(schema version): Always use latest version of schema

Bento007:
- raise warning if schema version is not latest
- update the schema_version when writing labels.
- relates  to #440 
- relates to https://github.com/chanzuckerberg/single-cell-data-portal/issues/4786


cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -130,23 +131,17 @@ def _set_schema_def(self):
...
-                    f"Schema version '{version}' is not supported. Current supported versions: '{supported_versions}'. "
-                    f"Validation cannot be performed."
-                )
-            else:
-                if not self.schema_def:
-                    self.schema_def = schema.get_schema_definition(version)
+        version = self.adata.uns.get("schema_version")
+        supported_versions = schema.get_schema_versions_supported()
+        supported_versions.sort()
+        if version and version != supported_versions[-1]:
```
nayib-jose-gloria, line 137: should we update this to only check against major version, as suggested in stand-up?

Bento007: Adding the complexity of comparing strictly the major version does not give us any added value. In practice we only ever have one schema that the CLI supports. So this should only ever have the latest schema.

Bento007: From #440 we will forcefully validate against the lasted scheme version regardless of the what the `schema_version` in the h5ad.

nayib-jose-gloria: True, but it means we'll always print this warning from the first minor schema update onwards. Since the yaml name is 3.0.0 and a new one is only produced for major version updates.

nayib-jose-gloria: It would also produce a confusing warning message, since it'd read:
`"Schema version '3.1.0' is not supported. Current supported versions: [3.0.0]. "`

nayib-jose-gloria: Maybe the change should be in [how we derive the supported versions](https://github.com/chanzuckerberg/single-cell-curation/blob/d18856cc00be4ef078caa379a413403a16b99d04/cellxgene_schema_cli/cellxgene_schema/schema.py#L20)

nayib-jose-gloria (approved):
ty!

PR #467, opened 2023-06-02
Add github actions to generate convert script and open a PR

Bento007:


.github/workflows/conversion_script.yml
```diff
@@ -0,0 +1,52 @@
...
+      - name: Install dependencies
+        run: |
+          pip install -r migration_assistant/requirements.txt
+      - name: Run Conversion Script Generator
+        run: |
+          make generate-conversion-script
+      - name: Get Version
+        id: get_version
+        run: |
+          echo "version=$(make show-current-version)" >> $GITHUB_ENV
```
nayib-jose-gloria, line 43: This will return the schema version before the version bump. We should either:
1) trigger this workflow after the release to PyPI (which contains the step that bumps the version), 
2) bump the version earlier in the workflow in its own step so it's picked up both here and in the PyPI release step, or 
3) Change the PR message to read 'update convert.py from <env.version> to <env.version + 0.1.0>'

.github/workflows/conversion_script.yml
```diff
@@ -0,0 +1,52 @@
...
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v5
+        with:
+          commit-message: 'AUTO: update convert.py to schema_version ${{ env.version }}'
```
nayib-jose-gloria, line 47: convert.py -> migrate.py

.github/workflows/conversion_script.yml
```diff
@@ -0,0 +1,63 @@
...
+          make generate-conversion-script
+      - name: Get Version
+        id: get_version
+        run: |
+          echo "old_version=$(make show-current-version)" >> $GITHUB_ENV
+      - name: setup git
+        run: |
+          git config user.name github-actions
+          git config user.email github-actions@github.com
+      - name: Bump Version
```
nayib-jose-gloria, line 48: We need to change the conditions that trigger this workflow and the publish-to-PyPI workflow. bumpversion is set up to commit, and both that workflow and this workflow call it + have the same triggers. So we'll end up bumping the version twice across both workflows.

nayib-jose-gloria: easiest route is likely to have a bump version workflow triggered by the ontology reference file updates, which in turn triggers both this workflow + publish-minor-to-pypi

Bento007: I move the GHA into publish-to-pypy.yml and running it after the package is published.

PR #472, opened 2023-06-05
fix(gencode report): change term to feature

Bento007:
Request by curators


nayib-jose-gloria (approved):
I'd double check if Deprecated Gene should also be Deprecated Feature, but otherwise looks good!

PR #480, opened 2023-06-09
feat(migration): add gencode changes to migration

Bento007:
- related to [#4798](https://github.com/chanzuckerberg/single-cell-data-portal/issues/4789)
- if there are *_diff.txt* files with deprecated features(genes), then the *migration.py* script will include the `remove_deprecated_features` function.
- removing deprecated features from both `adata.var` and `adata.raw.var` if raw is present. This is part of the lattice alogrithm.
- encapsulate the lattice gencode update logic into a function to simplify generating the migration_template as well as making unit testing easier.
- fix anndata test fixture to copy `anndata` object to raw rather than a pointer. This is how the test fixtures should have been setup originally to avoid producing unintended changes in both `anndata` and `anndata.raw`


cellxgene_schema_cli/cellxgene_schema/utils.py
```diff
@@ -11,3 +16,19 @@ def replace_ontology_term(dataframe, ontology_name, update_map):
...
+
+
+def remove_deprecated_features(adata: ad.AnnData) -> ad.AnnData:
+    deprecated = get_deprecated_feature_ids()
+
+    # Filter out genes that don't appear in the approved annotation
+    var_to_keep = adata.var.index[~adata.var.index.isin(deprecated)].tolist()
+    adata = adata[:, var_to_keep]
+
+    # Repeat much of the same steps for the raw.var, if it exists
```
nayib-jose-gloria, line 28: i think this is right, but are we 100% sure we also want to remove from raw.var? Its not in the epic-level requirements, did we check with Lattice?

scripts/migration_assistant/generate_script.py
```diff
@@ -30,16 +31,15 @@ def get_ontology_term_map(term_map_filepath) -> dict:
         return replaced_by_map
 
 
-def get_genecode_term_map() -> List[str]:
-    # TODO: read in the gencode term map
-    return []
+def migrate_gencode() -> bool:
+    return len(get_deprecated_feature_ids()) > 1
```
nayib-jose-gloria, line 35: I made an [issue to track this separately](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/single-cell-data-portal/4988), but one note--this will always return True because we never re-set the diff.txt's back to blank after a migration. So the repo will retain the list of the previous diff set even if the next migration doesn't update the GENCODE files. Probably not in scope to fix here, unless you think there's an easy place to resolve that as part of this PR

nayib-jose-gloria (commented):
Looks good overall! Just want to sync on a couple of things before merging

cellxgene_schema_cli/cellxgene_schema/utils.py
Bento007: I was looking at the lattice script and that is what they do. I'm not 100% sure one this. We can ask Lattice today.

Bento007:
I think I should embed list of deprecated genes in `migration.py` the same way you are for the ontologies. It increases transparency. Then we can remove the diff files.

nayib-jose-gloria (commented):
changes look good, though seems like an import is messed up somewhere as the unit tests for migration assistant files are failing

PR #489, opened 2023-06-09
fix(migrate): specify files to commit

Bento007:
specify files to commit to avoid committing extra files.

### DIFF
--- a/.github/workflows/minor-schema-version-bump.yml
+++ b/.github/workflows/minor-schema-version-bump.yml
@@ -10,7 +10,7 @@ on:
 
 permissions:
   id-token: write
-  contents: read
+  contents: write
   pull-requests: write
 
 jobs:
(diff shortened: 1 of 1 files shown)


nayib-jose-gloria (approved):

.github/workflows/minor-schema-version-bump.yml
```diff
@@ -215,3 +215,6 @@ jobs:
           branch: auto/update-convert-py-to-${{ env.new_version }}
           base: main
           reviewers: bento007 # TODO assign curators
+          add-paths:
```
nayib-jose-gloria, line 218: It looks like this needs to be slightly different syntax (tried it in the demo branch and it failed)

See https://github.com/peter-evans/create-pull-request#add-specific-paths

nayib-jose-gloria (approved):

PR #507, opened 2023-06-13
fix(gencode): Only update gencode files that have changed

Bento007:
Previously we were producing a new csv.gz file for all organism in gene_info.yml even if they haven't been update. This PR checks if the new version and old version of the gene_info are the same. If they are the same then the original is restored and the new one is discarded.

- Using the hash of gz files to determine if the contents are identical.
- Modify how gz files are stored to make comparing contents possible.
- Stopped modifying `gene_info` in _gene_processing.py_. Moved all values that can be derived from `gene_info` into `process_gene_info`. It makes the `gene_info` dictionary less complicated to understand.
- Only create a _diff.txt file if there are deprecated feature_ids. [#4988](https://github.com/chanzuckerberg/single-cell-data-portal/issues/4988)


cellxgene_schema_cli/scripts/gene_processing.py
```diff
@@ -209,9 +208,51 @@ def generate_gene_ref_diff(output_filename: str, current_ref_filepath: str, prev
...
+        processer = _parse_gtf
+    elif gene_info["url"].endswith("txt"):
+        processer = _process_ercc
+    else:
+        raise TypeError(f"unknown file type: {gene_info['file_type']}")
+
+    # add version to URL if needed
+    url = gene_info["url"].format(version=gene_info["version"]) if gene_info.get("version") else gene_info["url"]
+
+    new_file = os.path.join(env.ONTOLOGY_DIR, f"new_genes_{gene_info['description']}.csv.gz")
```
Bento007, line 237: Rather then renaming the original file before proceeded, the new_file has a different name that does not collide with the original file. If everything goes will we replace the original is replace with the new_file.

nayib-jose-gloria (approved):
i like it! thanks

PR #510, opened 2023-06-14
chore: add tests for migrate script + replace ontology term function 

nayib-jose-gloria:
https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/single-cell-data-portal/4791

- Seed migrate script with a blank template run to better model what form it will take
- Add unit test that runs migrate script against real h5ad fixture, ensures it can validate after
- Add unit tests for replace_ontology_term util function

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 3.0.2
+current_version = 3.1.0-rc.0
 commit = True
 parse = (?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)(?:-(?P<prerel>rc)\.(?P<prerelversion>\d+))?
 serialize = 
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 3.1.0-rc.0
+current_version = 3.0.2
 commit = True
 parse = (?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)(?:-(?P<prerel>rc)\.(?P<prerelversion>\d+))?
 serialize = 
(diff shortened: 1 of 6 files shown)


.bumpversion.cfg
```diff
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 3.0.2
```
Bento007, line 2: is this intentional?

cellxgene_schema_cli/cellxgene_schema/migrate.py
```diff
@@ -1,151 +1,54 @@
...
-        "EFO:0030027": "nucleus",
+    # ONTOLOGY TERMS TO UPDATE ACROSS ALL DATASETS IN CORPUS
+    # Initialization is AUTOMATED for newly deprecated terms that have 'Replaced By' terms in their ontology files
+
+    # Curators should review the monthly 'Curator Report' and add deprecated term replacements to corresponding map if
+    # 'Replaced By' is not available for a deprecated term.
+
+    # If Curators have non-deprecated term changes to apply to all datasets in the corpus where applicable,
+    # add them here.
+    ontology_term_maps = {
```
Bento007, line 19: If we move this out of the function we could mock it for better testings.

Bento007 (commented):
Was the version bump intentional?

.bumpversion.cfg
nayib-jose-gloria: Unintentional, fixing

Bento007 (approved):

cellxgene_schema_cli/cellxgene_schema/migrate.py
nayib-jose-gloria: I like having the unit test set up to run the migrate script as written without mocking. Especially since Lattice may need to modify this map after it's generated. At a minimum, it ensures a test run that would catch any syntax errors leading to an exception before merging. 

That said, still a good suggestion so we can mock it for other additional tests. I'll go ahead and merge for now but keeping that in mind for another pass

PR #512, opened 2023-06-16
Jahilton/data removal guidelines

jahilton:
Changing the guidance for how curators should request published data be deleted per https://github.com/chanzuckerberg/single-cell/issues/320


guidelines/curation_guidelines.md
```diff
@@ -27,4 +27,4 @@ As with Publishing new Collections, the curator should email the [Lattice team](
 Deleting Published Datasets
 ----------------
 Removing Data that have been publicly available should be avoided, but it is possible.
-Please contact the [Lattice team](mailto:lattice-info@lists.stanford.edu) with any Deletion requests along with a thorough justification for the removal.
+Please [create an issue in this repo](https://github.com/chanzuckerberg/single-cell-curation/issues/new) with any Deletion requests along with a thorough justification for the removal. The issue will need to be approved by the Lattice team prior to data removal by CELLxGENE developers.
```
brianraymor, line 30: `
Please [create an issue in single-cell-curation](https://github.com/chanzuckerberg/single-cell-curation/issues/new)
`

The issue must be approved by Jason Hilton prior to data removal by CELLxGENE developers.

---

I was thinking about a template but the expectation is that such requests should be few in nature?

brianraymor (approved):
My comments are only editorial Feel free to take or not. 

You might also label as "curation"?

guidelines/curation_guidelines.md
jahilton: They shouldn't happen frequently enough to warrant a template (nor would I want to give the impression that we anticipate them frequently enough that it warrants a template).
OK if they assign the issue to me so that I get notified? Anyone else they should add as assignee?

brianraymor: It would be best to assign to you.

PR #541, opened 2023-07-07
feat: add owl file url validation before running processing

nayib-jose-gloria:
Changes: 
- Check URLs for 200 response before starting download thread. Fail main thread if non-200 and stop auto-commit in the process

Tested with a couple of intentionally 404 URLs in this branch, reverted back to working order now


Bento007 (commented):
Are there any tests to update with this change?

nayib-jose-gloria:
> Are there any tests to update with this change?

We don't have any existing tests for this script, nor the gene processing script. Worth considering for a separate PR

Bento007 (approved):

PR #543, opened 2023-07-11
feat: upon schema updates, run scripts to generate ontology mappings and push to s3

nayib-jose-gloria:
[#4797](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/single-cell-data-portal/4797)

Changes:
- added step to post-minor schema update workflow that 1) runs scripts that generate ontology mappings and ontology descendant mappings used in data portal (for frontend filters, ancestor fields in portal API, gene_info API) using the newest pushed ontology reference files, 2) pushes these latest generated files to s3, and 3) pushes a webhook that will trigger data-portal to pick up these changes from s3 + commit to main (see https://github.com/chanzuckerberg/single-cell-data-portal/pull/5188) 
- moved scripts/compute_tissue_and_cell_type_mappings.ipynb from data portal to curation repo 
- updated scripts that generate mappings to use pinned URLs and point to correct filepaths where applicable 
- converted python notebooks to python scripts using jupyter cli, committed and using those in workflow. Kept original notebooks for documentation / curator manual use

Testing:
- Tested using a temporary workflow in this branch (see [here](https://github.com/chanzuckerberg/single-cell-curation/actions/runs/5524751653))


scripts/compute_mappings/compute_ancestor_mapping.py
```diff
@@ -0,0 +1,174 @@
+#!/usr/bin/env python
```
Bento007, line 1: Add somewhere what these were converts from and using what for reproducibility.

scripts/compute_mappings/compute_ancestor_mapping.py
```diff
@@ -0,0 +1,174 @@
...
+# 1. For `human` and `mouse`, start from each node available in the ontology
+# 1. for `UBERON`, only start from a defined, hardcoded subset of leaf nodes which is relevant for our datasets
+# 1. Recursively compute ancestors using the `is_a` relationship and `part_of` restriction (defined as `BFO_0000050`)
+# 1. Create mappings from each node to its ancestors
+# 1. Write a json file with these mappings (a dictionary of lists)
+
+# In[1]:
+
+
+from owlready2 import get_ontology  # noqa
```
Bento007, line 18: We should add unit test for this code now that it's not a note book.

scripts/compute_mappings/compute_ancestor_mapping.py
```diff
@@ -0,0 +1,174 @@
...
+import json
+import yaml
+
+
+# In[2]:
+
+
+def get_ancestors(class_name):
+    z = onto.search_one(iri=f"http://purl.obolibrary.org/obo/{class_name}")
```
Bento007, line 27: where does `onto` come from?

Bento007 (changes requested):
The python code generated by jupyter cli is not a well written script and will require some refactoring. To avoid refactoring the generated script, instead include instructions to only modify the notebook and regenerate the script from the source. Also add a banner to the top of the generated script to not modify the script and modify the source instead.

scripts/compute_mappings/compute_ancestor_mapping.py
nayib-jose-gloria: I'll make an issue for this separately

scripts/compute_mappings/compute_ancestor_mapping.py
nayib-jose-gloria: It was taking it from a global declared after this function declaration, which I agree is pretty confusing. I refactored the script to clean this part up

scripts/compute_mappings/compute_ancestor_mapping.py
nayib-jose-gloria: Added a README to the directory

Bento007 (approved):

PR #560, opened 2023-07-21
Change log for 3.0.3

Bento007:
The change in #552 will be reapplied after the release.


.bumpversion.cfg
```diff
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 3.0.2
+current_version = 3.0.3-rc.0
```
nayib-jose-gloria, line 2: bump to 3.0.3 rather than rc

nayib-jose-gloria (approved):

PR #596, opened 2023-08-04
feat(migrate): set migrate reviewers

Bento007:
Now the reviewers that are set when minor-schema-version-bump.yml can be changed using github variables. Before a new PR was required to modify this value. 
Screenshot 2023-08-04 at 1 43 06 PM


nayib-jose-gloria (commented):
the code change is good, but FYI I believe assigning github teams with this github action is not possible unless you use a human user's PAT (see team-reviewer details [in the docs](https://github.com/peter-evans/create-pull-request))

nayib-jose-gloria (approved):

PR #613, opened 2023-08-30
updated requirements for raw matrices

### LINKED ISSUE #612: Update requirements for raw matrices
# Context

See [**single-cell-census**](https://czi-sci.slack.com/archives/C048BN7K6U8/p1693321161006959).

# Design

@bkmartinjr proposal:

**Note**: "positive integers" needs to be clarified since the value is notated as a floating point (1.0 vs 1). 

## `X` (Matrix Layers)

...

The following table describes the matrix data and layers requirements that are **assay-specific**. If an entry in the table is empty, the schema does not have any other requirements on data in those layers beyond the ones listed above.

| Assay | "raw" required? | "raw" location | "normalized" required? | "normalized" location |
|-|-|-|-|-|
| scRNA-seq (UMI, e.g. 10x v3) | REQUIRED. Values MUST be de-duplicated molecule counts. Each cell MUST contain at least one non-zero value. All non-zero values MUST be positive integers.| `AnnData.raw.X` unless no "normalized" is provided, then `AnnData.X` | STRONGLY RECOMMENDED | `AnnData.X` |
| scRNA-seq (non-UMI, e.g. SS2) | REQUIRED. Values MUST be one of read counts (e.g. FeatureCounts) or  estimated fragments (e.g. output of RSEM). Each cell MUST contain at least one non-zero value. All non-zero values MUST be positive integers. | `AnnData.raw.X` unless no "normalized" is provided, then `AnnData.X` | STRONGLY RECOMMENDED | `AnnData.X` |
| Accessibility (e.g. ATAC-seq, mC-seq) | NOT REQUIRED | | REQUIRED | `AnnData.X` | STRONGLY RECOMMENDED |
|||||

brianraymor:
Closes #612

### DIFF
--- a/schema/4.0.0/schema.md
+++ b/schema/4.0.0/schema.md
@@ -180,8 +180,8 @@ The following table describes the matrix data and layers requirements that are *
 
 | Assay | "raw" required? | "raw" location | "normalized" required? | "normalized" location |
 |-|-|-|-|-|
-| scRNA-seq (UMI, e.g. 10x v3) | REQUIRED. Values MUST be de-duplicated molecule counts. Each cell MUST contain at least one non-zero value. All non-zero values MUST be positive integers converted to `float`.| `AnnData.raw.X` unless no "normalized" is provided, then `AnnData.X` | STRONGLY RECOMMENDED | `AnnData.X` |
-| scRNA-seq (non-UMI, e.g. SS2) | REQUIRED. Values MUST be one of read counts (e.g. FeatureCounts) or  estimated fragments (e.g. output of RSEM). Each cell MUST contain at least one non-zero value. All non-zero values MUST be positive integers converted to `float`. | `AnnData.raw.X` unless no "normalized" is provided, then `AnnData.X` | STRONGLY RECOMMENDED | `AnnData.X` |
+| scRNA-seq (UMI, e.g. 10x v3) | REQUIRED. Values MUST be de-duplicated molecule counts. Each cell MUST contain at least one non-zero value. All non-zero values MUST be positive integers stored as `numpy.float32`.| `AnnData.raw.X` unless no "normalized" is provided, then `AnnData.X` | STRONGLY RECOMMENDED | `AnnData.X` |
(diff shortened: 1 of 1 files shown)


schema/4.0.0/schema.md
```diff
@@ -180,8 +180,8 @@ The following table describes the matrix data and layers requirements that are *
 
 | Assay | "raw" required? | "raw" location | "normalized" required? | "normalized" location |
 |-|-|-|-|-|
-| scRNA-seq (UMI, e.g. 10x v3) | REQUIRED. Values MUST be de-duplicated molecule counts. | `AnnData.raw.X` unless no "normalized" is provided, then `AnnData.X` | STRONGLY RECOMMENDED | `AnnData.X` |
-| scRNA-seq (non-UMI, e.g. SS2) | REQUIRED. Values MUST be one of read counts (e.g. FeatureCounts) or  estimated fragments (e.g. output of RSEM). | `AnnData.raw.X` unless no "normalized" is provided, then `AnnData.X` | STRONGLY RECOMMENDED | `AnnData.X` |
+| scRNA-seq (UMI, e.g. 10x v3) | REQUIRED. Values MUST be de-duplicated molecule counts. Each cell MUST contain at least one non-zero value. All non-zero values MUST be positive integers converted to `float`.| `AnnData.raw.X` unless no "normalized" is provided, then `AnnData.X` | STRONGLY RECOMMENDED | `AnnData.X` |
```
bkmartinjr, line 183: nit: "positive integers converted to `float`" - perhaps it would be more clear as "positive integers stored as `float32`"?

bkmartinjr (approved):

schema/4.0.0/schema.md
brianraymor: How about "positive integers stored as `numpy.float32` to be even more precise?

bkmartinjr: No concerns. But I noticed that other parts of the schema are not python type specific, so it would introduce a bit of weirdness.

I suggest "IEEE 754 32-bit float" if you want to be both overly specific and language agnostic :-)

brianraymor: We're already referencing numpy constants for embeddings and:

**_Note on types_**
The types below are python3 types. Note that a python3 str is a sequence of Unicode code points, which is stored null-terminated and UTF-8-encoded by anndata.

Are there areas where we could improve?

Issue #607, opened 2023-08-23
cellxgene-schema CLI must validate schema_version


brianraymor:
# Context

See [**single-cell-four**](https://czi-sci.slack.com/archives/C03PXFMM7ST/p1692813292823499?thread_ts=1692809869.002419&cid=C03PXFMM7ST).

As part of automated migration, the schema transitioned from curators annotating `schema_version` to CELLxGENE Discover annotating `schema_version`.  For 3.1, schema_version is temporarily overwritten as part of the transition:

_When a dataset is uploaded, CELLxGENE Discover MUST automatically add the schema_version key and its value to uns. If schema_version is already defined, then its value MUST be overwritten._

In 4.0, curators must not annotate this field. An error should be thrown blocking validation if it is provided, rather than the overwriting behavior it currently uses.

nayib-jose-gloria:
@jahilton Ready for testing! For all testing, you can pull the latest release candidate version of 4.0.0 by running:
`pip install git+https://github.com/chanzuckerberg/single-cell-curation/@main#subdirectory=cellxgene_schema_cli`

jahilton:
Unexpected behavior: the file will pass validation if uns['schema_version'] = `None`
Expected to fail because 'schema_version' is present in uns.

Sample adata was set by `adata.uns['schema_version'] = None` & writing to .h5ad

jahilton:
You are correct. It drops when saving so the tested .h5ad didn't actually have uns.schema_version.
In that case, all clear!

PR #617, opened 2023-09-12
feat(schema 4): update schema version validation for 4.0.0 + refactor schema definition loading

nayib-jose-gloria:
#607

Changes:
- Throw a validation error when an h5ad has a schema_version in 'uns' (now a reserved column to be annotated by add-labels only)
- Updated h5ad test fixtures to remove schema_version from 'uns'
- Refactor how we load schema definition yaml to be agnostic to schema version (previously dependent on the latest major version number and required updating the file name each major schema update, but we only ever support one major schema version at a time and have no desire to do otherwise)
- Simplified or removed functions that assumed support of multiple schema versions at once
- Bump version to 4.0.0 release candidate 
- Updated unit tests--removed tests that assumed schema_version in 'uns' and added test to ensure proper validation of reserved columns in 'uns'


Bento007 (approved):
:+1:

PR #618, opened 2023-09-13
feat: improve validation to X_{suffix} embeddings

joyceyan:
https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/single-cell-curation/590

not sure what the best way to test this is

### DIFF
--- a/cellxgene_schema_cli/cellxgene_schema/validate.py
+++ b/cellxgene_schema_cli/cellxgene_schema/validate.py
@@ -842,13 +842,9 @@ class Validator:
                         "float, integer, or unsigned integer of any precision (8, 16, 32, or 64 bits)."
                     )
                 if np.isinf(value).any() or np.isnan(value).any():
-                    self.errors.append(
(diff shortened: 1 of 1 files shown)


seve (commented):
Do we want an if/elif chain, or do we just want if's so we can output all possible validation errors instead of just breaking on the first one?

joyceyan:
> Do we want an if/elif chain, or do we just want if's so we can output all possible validation errors instead of just breaking on the first one?

ah that's a good point - i think using if's is a bit better so that we can see all the validation errors

seve (approved):

nayib-jose-gloria (commented):
Would you mind a follow-up PR adding unit tests for this in [this](https://github.com/chanzuckerberg/single-cell-curation/blob/main/cellxgene_schema_cli/tests/test_schema_compliance.py#L1212) test subclass? ty!

PR #619, opened 2023-09-14
feat: validate raw matrix does not have more genes than raw var

danieljhegeman:
I factored out the dimension attribute access so that testing could make use of `unittest.mock.patch`; the AnnData object appears to be somewhat resistant to my attempts to edit the actual `.raw.X` and `.raw.var`.

### DIFF
--- a/cellxgene_schema_cli/cellxgene_schema/utils.py
+++ b/cellxgene_schema_cli/cellxgene_schema/utils.py
@@ -50,6 +50,14 @@ def remove_deprecated_features(adata: ad.AnnData, deprecated: List[str]) -> ad.A
     return adata
 
 
+def get_num_vars_in_raw_matrix(adata: ad.AnnData) -> int:
+    return adata.raw.X.shape[1]
+
+
+def get_num_vars_in_raw_var(adata: ad.AnnData) -> int:
+    return adata.raw.var.shape[0]
+
+
 def get_matrix_format(adata: ad.AnnData, matrix: Union[np.ndarray, sparse.spmatrix]) -> str:
     """
     Given a matrix, returns the format as one of: csc, csr, coo, dense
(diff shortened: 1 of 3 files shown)


cellxgene_schema_cli/cellxgene_schema/utils.py
```diff
@@ -50,6 +50,14 @@ def remove_deprecated_features(adata: ad.AnnData, deprecated: List[str]) -> ad.A
     return adata
 
 
+def get_num_vars_in_raw_matrix(adata: ad.AnnData) -> int:
```
nayib-jose-gloria, line 53: nit: I don't think its worth a util function for 1 line, same effect to just have this line + a comment in validate.py

nayib-jose-gloria: nvm--I see your note in the PR description to do this for easy mocking

nayib-jose-gloria (commented):
I got a branch working with a test that uses an anndata fixture rather than the util function mocks; I created a PR for that against this PR. It also takes care of the small nits/checks I asked for in the review (i.e. self.adata.raw exists check, util function refactor). Review and merge if you like the changes 

https://github.com/chanzuckerberg/single-cell-curation/pull/621

cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -801,6 +801,15 @@ def _validate_seurat_convertibility(self):
 
                 self.is_seurat_convertible = False
 
+        # Check to make sure raw.var has the same dimensionality as raw.X
+        if get_num_vars_in_raw_matrix(self.adata) != get_num_vars_in_raw_var(self.adata):
```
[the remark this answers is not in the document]
danieljhegeman: Good to know, thanks!

nayib-jose-gloria (approved):
Address the remaining lint issues, and then it looks ready to merge and flag for QA!

Issue #517, opened 2023-06-16
cellxgene-schema must validate cell_type_ontology_term_id


brianraymor:
# Design

See [cell_type_ontology_term_id](https://github.com/chanzuckerberg/single-cell-curation/blob/main/schema/4.0.0/schema.md#cell_type_ontology_term_id).

### cell_type_ontology_term_id

<table><tbody>
    <tr>
      <th>Key</th>
      <td>cell_type_ontology_term_id</td>
    </tr>
    <tr>
      <th>Annotator</th>
      <td>Curator</td>
    </tr>
    <tr>
      <th>Value</th>
        <td>categorical with <code>str</code> categories. This MUST be a CL term.<br><br>If no appropriate high-level term can be found or the cell type is unknown, then it is STRONGLY RECOMMENDED to use <a href="https://www.ebi.ac.uk/ols4/ontologies/cl/terms?obo_id=CL:0000003"><code>"CL:0000003"</code></a> for <i>native cell</i>. The following terms MUST NOT be used:
        <ul><li>
          <a href="https://www.ebi.ac.uk/ols4/ontologies/cl/terms?obo_id=CL:0000255"><code>"CL:0000255"</code></a> for <i>eukaryotic cell</i>
        </li>
        <li>
          <a href="https://www.ebi.ac.uk/ols4/ontologies/cl/terms?obo_id=CL:0000257"><code>"CL:0000257"</code></a> for <i>Eumycetozoan cell</i>
        </li>
        <li>
            <a href="https://www.ebi.ac.uk/ols4/ontologies/cl/terms?obo_id=CL:0000548"><code>"CL:0000548"</code></a> for <i>animal cell</i>
         </li></ul>
    </tr>
</tbody></table>
<br>

nayib-jose-gloria:
@jahilton / @corismall ready for testing!

jahilton:
Passes QA. Good-to-go
[QA notebook](https://github.com/Lattice-Data/lattice-tools/blob/CXG290-schema4validator-testing/cellxgene_resources/cxg_4.0.0_testing/516_annotate_feature_length.ipynb)

Issue #514, opened 2023-06-16
cellxgene-schema must validate tissue_type


brianraymor:
# Design

This also changes the validation and label application for `tissue_ontology_term_id` and `tissue`.

See [tissue_type](https://github.com/chanzuckerberg/single-cell-curation/blob/main/schema/4.0.0/schema.md#tissue_type).

### tissue_type

<table><tbody>
    <tr>
      <th>Key</th>
      <td>tissue_type</td>
    </tr>
    <tr>
      <th>Annotator</th>
      <td>Curator</td>
    </tr>
    <tr>
      <th>Value</th>
        <td>categorical with <code>str</code> categories. This MUST be <code>"tissue"</code>, <code>"organoid"</code>, or <code>"cell culture"</code>.
    </tr>
</tbody></table>
<br>

### tissue_ontology_term_id

<table><tbody>
    <tr>
      <th>Key</th>
      <td>tissue_ontology_term_id</td>
    </tr>
    <tr>
      <th>Annotator</th>
      <td>Curator</td>
    </tr>
   <tr>
      <th>Value</th>
        <td>categorical with <code>str</code> categories. If <code>tissue_type</code> is <code>"tissue"</code> or <code>"organoid"</code>, this MUST be the most accurate child of <a href="https://www.ebi.ac.uk/ols4/ontologies/uberon/classes?obo_id=UBERON%3A0001062"><code>UBERON:0001062</code></a> for <i>anatomical entity</i>.<br><br> If <code>tissue_type</code> is <code>"cell culture"</code> this MUST follow the requirements for <code>cell_type_ontology_term_id<code>.</td>
     </tr>
</tbody></table>
<br>

### tissue

<table><tbody>
    <tr>
      <th>Key</th>
      <td>tissue</td>
    </tr>
    <tr>
      <th>Annotator</th>
      <td>CELLxGENE Discover</td>
    </tr>
    <tr>
      <th>Value</th>
        <td>categorical with <code>str</code> categories. This MUST be the human-readable name assigned to the value of <code>tissue_ontology_term_id</code>.
        </td>
    </tr>
</tbody></table>

nayib-jose-gloria:
@jahilton / @corismall ready for testing!

[...]

PR #623, opened 2023-09-15
feat: add tissue_type and new validation rules for tissue_ontology_term_id and cell_type_ontology_term_id

nayib-jose-gloria:
#514 
#517 

Changes: 
- added categorical field 'tissue_type' as required curator-annotated field, enforcing value as one of either 'organoid', 'cell culture', or 'tissue'
- enforced new cell_type_ontology_term_id rules (i.e. MUST be CL term but must NOT be one of the listed forbidden terms)
- enforced new tissue_type_ontology_term_id rules
               - do NOT accept suffixes in term_ids and do not append them to the resulting label
               - enforce cell_type_ontology_term_id rules on this term_id IF tissue_type is 'cell culture', 
               - enforce term_id is child term of [UBERON:0001062](https://www.ebi.ac.uk/ols4/ontologies/uberon/classes?obo_id=UBERON%3A0001062) if tissue_type is 'organoid' or 'tissue'
- simplified parts of code that enforced term_id suffix logic (in write-labels and validate modules), as they no longer apply to any terms and we have no plans to support them again
- removed validation of schema definition in 'validate' command (this is already handled by unit tests, we shouldn't expose code validation logic into a user / deployment facing function)
- fixed tests enforcing schema definition (were not being run previously due to a bad if check) and updated them to fit current schema definition structure
- added tests + update h5ad fixtures to contain tissue_type column w/ appropriate values


cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -800,6 +761,13 @@ def _validate_seurat_convertibility(self):
                     )
 
                 self.is_seurat_convertible = False
+        if self.adata.raw and self.adata.raw.X.shape[1] != self.adata.raw.var.shape[0]:
```
nayib-jose-gloria, line 764: this is in main already, not sure why its showing up as a diff...

cellxgene_schema_cli/tests/test_schema_compliance.py
```diff
@@ -281,16 +281,31 @@ def test_assay_ontology_term_id(self):
 
     def test_cell_type_ontology_term_id(self):
         """
-        cell_type_ontology_term_id categorical with str categories. This MUST be a CL term.
+        cell_type_ontology_term_id categorical with str categories. This MUST be a CL term, and must NOT match forbidden
```
Bento007, line 284: Do we have a locally glossary somewhere for these acronyms(CL)?

cellxgene_schema_cli/tests/test_schema_compliance.py
```diff
@@ -281,16 +281,31 @@ def test_assay_ontology_term_id(self):
...
         # Not a valid term
-        self.validator.adata.obs.loc[self.validator.adata.obs.index[0], "cell_type_ontology_term_id"] = "EFO:0000001"
-        self.validator.validate_adata()
-        self.assertEqual(
-            self.validator.errors,
-            ["ERROR: 'EFO:0000001' in 'cell_type_ontology_term_id' is not a valid " "ontology term id of 'CL'."],
-        )
+        with self.subTest(forbidden_term="EFO:0000001"):
+            self.validator.adata.obs.loc[
+                self.validator.adata.obs.index[0], "cell_type_ontology_term_id"
```
Bento007, line 291: What is the benefit of using the numpy method of slicing `foo[ x, y]`, as opposed to pythons way using `foo[x][y]`?

Bento007 (approved):

cellxgene_schema_cli/tests/test_schema_compliance.py
nayib-jose-gloria: I don't think its an acronym, its a prefix used for cell ontology terms. There's a glossary in the user-facing [schema](https://github.com/chanzuckerberg/single-cell-curation/blob/main/schema/4.0.0/schema.md#required-ontologies) documentation

cellxgene_schema_cli/tests/test_schema_compliance.py
nayib-jose-gloria: this is the syntax for the [loc](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.loc.html) accessor function for pandas dataframes

Issue #516, opened 2023-06-16
cellxgene-schema must annotate feature_length


brianraymor:
# Design

@pablo-gar writes:

This information is already calculated from the pinned Gene References and is readily available here `single-cell-curation/cellxgene_schema_cli/cellxgene_schema/ontology_files/` on the files `genes_[organism].csv.gz`

See [feature_length](https://github.com/chanzuckerberg/single-cell-curation/blob/main/schema/4.0.0/schema.md#feature_length). 

### feature_length

<table><tbody>
    <tr>
      <th>Key</th>
      <td>feature_length</td>
    </tr>
    <tr>
      <th>Annotator</th>
      <td>CELLxGENE Discover</td>
    </tr>
    <tr>
      <th>Value</th>
        <td>
        <code>uint</code> number of base-pairs (bps). If the <code>feature_biotype</code> is <code>"gene"</code>, then the value is calculated by creating non-overlapping concatenated exons across all isoforms of the gene, and then adding up their length in base-pairs. This approach is modeled on the "length of merged exons of isoforms of a gene" from <a href="https://doi.org/10.1093/bioinformatics/btac561">GTFtools: a software package for analyzing various features of gene models.</a><br><br> If <code>feature_biotype</code> is NOT <code>"gene"</code> then the value MUST be set to 0.
      </td>
    </tr>
</tbody></table>

nayib-jose-gloria:
@jahilton / @corismall ready for testing!

jahilton:
good to go
[QA Notebook](https://github.com/Lattice-Data/lattice-tools/blob/CXG290-schema4validator-testing/cellxgene_resources/cxg_4.0.0_testing/516_annotate_feature_length.ipynb)

PR #624, opened 2023-09-15
feat: add feature length field, annotated by add-labels function

nayib-jose-gloria:
#516 

Changes:
- annotate feature_length field in add-labels function, based on feature_id and pre-computed gene info CSV files (derived + committed to repo during ontology bumps). Set to 0 for non-gene features (even though a length is calculated + available for spike-in controls) 
- refactor gene ontology tool that fetches info from gene CSV files to accommodate fetching calculated feature_length 
- update tests

### DIFF
--- a/cellxgene_schema_cli/cellxgene_schema/ontology.py
+++ b/cellxgene_schema_cli/cellxgene_schema/ontology.py
@@ -93,7 +93,7 @@ class GeneChecker:
 
         return gene_id in self.gene_dict
 
-    def get_symbol(self, gene_id) -> str:
+    def get_symbol(self, gene_id: str) -> str:
         """
         Gets symbol associated to the ENSEBML id
 
@@ -103,12 +103,12 @@ class GeneChecker:
         :return A gene symbol
         """
 
-        if not self.is_valid_id(gene_id):
+        if self.is_valid_id(gene_id):
+            return self.gene_dict[gene_id][0]
+        else:
             raise ValueError(f"The id '{gene_id}' is not a valid ENSEMBL id for '{self.species}'")
 
-        return self.gene_dict[gene_id][0]
-
-    def get_length(self, gene_id) -> int:
+    def get_length(self, gene_id: str) -> int:
         """
         Gets feature length associated to the ENSEBML id
 
@@ -118,10 +118,10 @@ class GeneChecker:
         :return A gene length
         """
 
-        if not self.is_valid_id(gene_id):
+        if self.is_valid_id(gene_id):
+            return self.gene_dict[gene_id][1]
+        else:
             raise ValueError(f"The id '{gene_id}' is not a valid ENSEMBL id for '{self.species}'")
-
-        return self.gene_dict[gene_id][1]
 
 
 class OntologyChecker:
--- a/cellxgene_schema_cli/cellxgene_schema/ontology.py
+++ b/cellxgene_schema_cli/cellxgene_schema/ontology.py
@@ -66,8 +66,9 @@ class GeneChecker:
                 gene = gene.rstrip().split(",")
                 gene_id = gene[0]
                 gene_label = gene[1]
+                gene_length = int(gene[3])
 
-                self.gene_dict[gene_id] = gene_label
+                self.gene_dict[gene_id] = (gene_label, gene_length)
 
                 # Keeps track of duplicated gene labels
                 if gene_label in gene_labels:
@@ -76,9 +77,9 @@ class GeneChecker:
                     gene_labels.add(gene_label)
 
             # Makes gene labels unique
-            for gene_id, gene_label in self.gene_dict.items():
+            for gene_id, (gene_label, gene_length) in self.gene_dict.items():
                 if gene_label in duplicated_gene_labels:
-                    self.gene_dict[gene_id] = gene_label + "_" + gene_id
+                    self.gene_dict[gene_id] = (gene_label + "_" + gene_id, gene_length)
 
     def is_valid_id(self, gene_id: str) -> bool:
         """
@@ -105,7 +106,22 @@ class GeneChecker:
         if not self.is_valid_id(gene_id):
             raise ValueError(f"The id '{gene_id}' is not a valid ENSEMBL id for '{self.species}'")
 
-        return self.gene_dict[gene_id]
+        return self.gene_dict[gene_id][0]
+
+    def get_length(self, gene_id) -> int:
+        """
+        Gets feature length associated to the ENSEBML id
+
+        :param str gene_id: ENSEMBL gene id
+
+        :rtype int
+        :return A gene length
+        """
+
+        if not self.is_valid_id(gene_id):
+            raise ValueError(f"The id '{gene_id}' is not a valid ENSEMBL id for '{self.species}'")
+
+        return self.gene_dict[gene_id][1]
 
 
 class OntologyChecker:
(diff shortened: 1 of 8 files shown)


danieljhegeman:
>for spike-in proteins

I can't believe I'm asking this but what is a "spike-in _protein_"? From Google:

```Spike-in controls are synthetic nucleic-acid sequences that are added to a user's sample and constitute internal standards for subsequent steps in the next generation sequencing workflow```

File: cellxgene_schema_cli/cellxgene_schema/ontology.py [hunk 1]
```diff
@@ -105,7 +106,22 @@ def get_symbol(self, gene_id) -> str:
         if not self.is_valid_id(gene_id):
             raise ValueError(f"The id '{gene_id}' is not a valid ENSEMBL id for '{self.species}'")
 
-        return self.gene_dict[gene_id]
+        return self.gene_dict[gene_id][0]
```
danieljhegeman, line 109: can you elaborate on this change? I'm not seeing compensatory changes where it is called, so I'm assuming this is an unrelated bug fix? Does it need a test?

nayib-jose-gloria: Its related, and tested in `test_ontology` as well as `test_validate`! We previously only mapped gene_id to a single value in the gene_dict (gene_label). Now that we also want to be able to fetch the gene length by ID, I updated the gene_dict to map to a tuple of (gene_label, gene_length), which we derive from the csv fixtures.

accompanying change to populating the gene_dict is here: https://github.com/chanzuckerberg/single-cell-curation/pull/624/files/fb60f0baba71b9d22da47dd6d8708ba16d75322f#diff-e03634d528c86621165785d45334e8b7255794f658cb01e118edd1f6ce817df9R71

nayib-jose-gloria:
> > for spike-in proteins
> 
> I can't believe I'm asking this but what is a "spike-in _protein_"? From Google:
> 
> `Spike-in controls are synthetic nucleic-acid sequences that are added to a user's sample and constitute internal standards for subsequent steps in the next generation sequencing workflow`

oops--I did mean spike-in control, I have no idea why I wrote protein 😅 sorry!

cellxgene_schema_cli/cellxgene_schema/ontology.py
danieljhegeman: ah makes sense. Thanks

danieljhegeman:
@nayib-jose-gloria tangentially-related: would you mind cleaning up this method comment, i.e. turn it into parseable English 🙏 

https://github.com/chanzuckerberg/single-cell-curation/blob/f6bb2324a42332fa1437b3d94b7f283d61837f59/cellxgene_schema_cli/cellxgene_schema/write_labels.py#L324-L325

cellxgene_schema_cli/cellxgene_schema/ontology.py
```diff
@@ -105,7 +106,22 @@ def get_symbol(self, gene_id) -> str:
...
+
+    def get_length(self, gene_id) -> int:
```
Bento007, line 111: type gene_id

cellxgene_schema_cli/cellxgene_schema/write_labels.py
```diff
@@ -209,6 +209,27 @@ def _get_mapping_dict_feature_biotype(self, ids: List[str]) -> Dict[str, str]:
...
+
+        :param list[str] ids: feature IDs use for mapping
+
+        :return a mapping dictionary: {id: <int>, id: 0, ...}
+        :rtype dict
+        """
+        mapping_dict = {}
+
+        for i in ids:
+            if i.startswith("ENS"):
```
Bento007, line 225: Should we start putting hardcoded string into a constant.py file? Not sure we have enough cases for it to make sense.

cellxgene_schema_cli/tests/test_schema_compliance.py
```diff
@@ -1010,6 +1010,29 @@ def test_should_warn_for_low_gene_count(self):
...
         )
 
+    def test_add_label_fields_are_reserved(self):
+        """
+        Raise an error if column names flagged as 'add_label' -> 'to_column' in the schema definition are not available.
+        """
+        for df in ["var", "raw.var"]:
+            for i in self.validator.schema_def["components"][df]["index"]["add_labels"]:
+                column = i["to_column"]
+                with self.subTest(column=column, df=df):
```
Bento007, line 1020: use https://docs.pytest.org/en/7.3.x/how-to/parametrize.html

cellxgene_schema_cli/tests/test_schema_compliance.py
```diff
@@ -1010,6 +1010,29 @@ def test_should_warn_for_low_gene_count(self):
...
+                    # Resetting validator
+                    self.validator.adata = examples.adata.copy()
```
Bento007, line 1022: I believe these 2 lines can be removed if you switch to a pytest approach.

Bento007 (changes requested):

cellxgene_schema_cli/cellxgene_schema/write_labels.py
danieljhegeman: Might be a good thing to do at the end of the epic? Or we can start now

cellxgene_schema_cli/cellxgene_schema/ontology.py
```diff
@@ -105,7 +106,22 @@ def get_symbol(self, gene_id) -> str:
...
+        """
+        Gets feature length associated to the ENSEBML id
+
+        :param str gene_id: ENSEMBL gene id
+
+        :rtype int
+        :return A gene length
+        """
+
+        if not self.is_valid_id(gene_id):
```
danieljhegeman, line 121: @Bento007 has encouraged me to avoid negatives. 

Could turn this to a one-liner:
```return self.gene_dict[gene_id][1] if self.is_valid_id(gene_id) else raise ValueError(f"The id '{gene_id}' is not a valid ENSEMBL id for '{self.species}'")
```

🙂

cellxgene_schema_cli/tests/test_validate.py
```diff
@@ -136,7 +135,29 @@ def test_get_dictionary_mapping_feature_reference(self):
...
+            0,  # non-gene feature, so set to 0 regardless of csv value
+            42738,
+            4045,
+            3822,
+        ]
+        expected_dict = dict(zip(ids, gene_lengths))
+        self.assertEqual(self.writer._get_mapping_dict_feature_length(ids), expected_dict)
+
+        # Bad
+        ids = ["NO_GENE"]
```
danieljhegeman, line 160: For all of the `test_get_dictionary_*` tests in `TestAddLabelFunctions`, you wish to consider factoring out common dict comparison logic and `ids` initial variables. Could create a helper function or could simply group all under a single test. Not required though.

danieljhegeman (approved):

cellxgene_schema_cli/cellxgene_schema/write_labels.py
nayib-jose-gloria: I don't think it's necessary at this point, we can revisit if we find that it's coming up frequently and causing confusion.

cellxgene_schema_cli/tests/test_schema_compliance.py
nayib-jose-gloria: https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/single-cell-curation/633 tracking a separate issue for this

Bento007 (approved):

Issue #518, opened 2023-06-16
cellxgene-schema must annotate schema_reference


brianraymor:
# Design

See [schema_reference](https://github.com/chanzuckerberg/single-cell-curation/blob/main/schema/4.0.0/schema.md#schema_reference). 

When a dataset is uploaded, CELLxGENE Discover MUST automatically add the `schema_reference` key and set its value to the permanent URL of this document. 

### schema_reference

<table><tbody>
    <tr>
      <th>Key</th>
      <td>schema_reference</td>
    </tr>
    <tr>
      <th>Annotator</th>
      <td>CELLxGENE Discover</td>
    </tr>
    <tr>
      <th>Value</th>
        <td>
          This MUST be <code>"https://github.com/chanzuckerberg/single-cell-curation/blob/main/schema/4.0.0/schema.md"</code>.
        </td>
    </tr>
</tbody></table>

nayib-jose-gloria:
@jahilton @corismall ready for testing! 

`pip install git+https://github.com/chanzuckerberg/single-cell-curation/@main#subdirectory=cellxgene_schema_cli`

jahilton:
--add-labels did not add uns.schema_reference in [our test](https://github.com/Lattice-Data/lattice-tools/blob/CXG290-schema4validator-testing/cellxgene_resources/cxg_4.0.0_testing/518_annotate_schema_ref.ipynb)

PR #626, opened 2023-09-15
feat: add schema_reference field to uns during add-labels

nayib-jose-gloria:
#518 

Changes:
- add schema_reference as reserved column in 'uns' for validation
- populate schema_reference field during add-labels step


danieljhegeman (approved):
LGTM

Issue #523, opened 2023-06-21
cellxgene-schema must annotate observation_joinid


brianraymor:
# Design

See [schema 4](https://github.com/chanzuckerberg/single-cell-curation/blob/main/schema/4.0.0/schema.md#observation_joinid).

The current Census implementation is a base85 (rfc 1924) hash digest of the original AnnData index, implemented in chanzuckerberg/cellxgene-census#666.

# Prototype

Updated prototype based on chanzuckerberg/cellxgene-census#666.

@bkmartinjr writes:

Prototype which adds `observation_joinid` to an H5AD and writes out a new H5AD containing the modification. 

```
import os.path
import sys

import anndata
import numpy as np
from xxhash import xxh3_64_intdigest
from base64 import b85encode

"""
Annotate an H5AD with `observation_joinid`

Uses xxhash (https://xxhash.com/), via https://pypi.python.org/pypi/xxhash/.

NOTE: xxhash is licensed BSD-2
"""


def main():
    h5ad_filename = sys.argv[1]
    ad = anndata.read_h5ad(h5ad_filename)

     ad.obs["observation_joinid"] = (
            ad.obs.index.to_series()
            .map(xxh3_64_intdigest)
            .astype(np.uint64)
            .apply(lambda v: b85encode(v.to_bytes(8, "big")).decode("ascii"))
        )

    fpath, fext = os.path.splitext(h5ad_filename)
    save_to_filename = f"{fpath}_with_joinid.h5ad"
    ad.write(save_to_filename)


if __name__ == "__main__":
    sys.exit(main())
```

bkmartinjr:
One caveat: there are open questions regarding the use of an int64 digest as applies to R users.  There is some chance we may need to move to either an int32 digest, or a string encoded 64 bit digest.  CC: @mlin @pablo-gar

nayib-jose-gloria:
@pablo-gar this ticket is ready for QA. If you run the following command:
```
pip install git+https://github.com/chanzuckerberg/single-cell-curation/@main#subdirectory=cellxgene_schema_cli --upgrade
```

You will install a version of the CLI that outputs a labeled h5ad with an added observation_joinid column in 'obs' when you run 
`cellxgene-schema validate <input_h5ad_filepath> -a <output_labeled_h5ad_filepath>`

Let me know if you need any more info to validate that this is working as expected for the census use-case. Thanks!

nayib-jose-gloria:
[...]

PR #629, opened 2023-09-18
feat: add observation_joinid to obs during add-labels

nayib-jose-gloria:
## Reason for Change

- #523

## Changes

- validate that observation_joinid is NOT already in-use in obs by submitted datasets
- set observation_joinid in obs during add-labels function, using bruce's algorithm (defined in issue)
- introduces new hash library dependency

## Testing

- Added Unit Tests for validating 'observation_joinid' as reserved column name + added observation_joinid to expected obs df fixture, used for comparison in an existing test to assert an unlabeled anndata fixture is annotated as expected by the add-labels module
- tested CLI locally

## Notes for Reviewer


cellxgene_schema_cli/cellxgene_schema/write_labels.py
```diff
@@ -327,6 +330,17 @@ def _remove_categories_with_zero_values(self):
     def _build_schema_reference_url(self, schema_version: str):
         return f"{SCHEMA_REFERENCE_BASE_URL}/{schema_version}/{SCHEMA_REFERENCE_FILE_NAME}"
 
+    def _get_observation_joinid_column(self):
+        """
+        Set column with unique join ID for each row in obs dataframe.
+        """
+        return (
+            self.adata.obs.index.to_series()
```
Bento007, line 338: This command is mouthful and it also being use in the test fixtures. Is there a way to reuse this to avoid the test fixture implementation becoming out of sync with this version.

cellxgene_schema_cli/tests/test_schema_compliance.py
```diff
@@ -224,6 +224,27 @@ def test_column_presence_assay(self):
             ],
         )
 
+    def test_obs_reserved_columns_presence(self):
+        """
+        Reserved columns must NOT be used in obs
+        """
+
+        for reserved_column in self.validator.schema_def["components"]["obs"]["reserved_columns"]:
+            with self.subTest(column=reserved_column):
```
Bento007, line 233: we are using pytest for running the tests and it does not play well with subtests. I suggest using https://docs.pytest.org/en/7.3.x/how-to/parametrize.html to achieve the same behavior.

Bento007 (changes requested):

cellxgene_schema_cli/tests/test_schema_compliance.py
nayib-jose-gloria: Where does it not playing well manifest? When I run `make unit-test` everything seems fine, and we use subtests all throughout the test suite for the repo. I'm down to look into it, just wondering if we have worse test coverage than we think if subtests aren't working properly

nayib-jose-gloria: tracking separate issue for this https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/single-cell-curation/633

cellxgene_schema_cli/tests/test_utils.py
```diff
@@ -107,3 +109,12 @@ def test_adata_with_canonical_raw_X(self, adata_with_raw):
...
         assert adata_without_raw.X.has_canonical_format is True
+
+
+class TestGetHashDigestColumn:
+    def test_get_hash_digest_column(self, adata_with_raw):
+        hash_digest_column = get_hash_digest_column(adata_with_raw.obs)
+        assert isinstance(hash_digest_column, Series)
+        for val in hash_digest_column:
+            assert isinstance(val, str)
+            assert len(val) == 10  # 5 ASCII chars / 4 bytes, each ID is 8 bytes
```
danieljhegeman, line 120: Do you want to instantiate a test df, and test against the expected product?

danieljhegeman (changes requested):
just one request for the test

Bento007 (approved):

cellxgene_schema_cli/tests/test_utils.py
nayib-jose-gloria: updated!

danieljhegeman (approved):

cellxgene_schema_cli/cellxgene_schema/utils.py
```diff
@@ -146,3 +148,16 @@ def _enforce_canonical_format(df):
...
         logger.info("enforce canonical format in raw.X")
         _enforce_canonical_format(adata.raw)
+
+
+def get_hash_digest_column(dataframe):
+    """
+    Get column with hash digest for each row in dataframe.
+    """
+
+    return (
```
bkmartinjr, line 158: this code - as currently used - assumes the index is unique (which is not something AnnData requires).  For sanity's sake, you might want to assert `datafame.index.is_unique` here.

bkmartinjr (commented):
one small (optional) suggestion. Otherwise LGTM

cellxgene_schema_cli/cellxgene_schema/utils.py
nayib-jose-gloria: we validate [uniqueness](https://github.com/chanzuckerberg/single-cell-curation/blob/main/cellxgene_schema_cli/cellxgene_schema/validate.py#L356) of [indices](https://github.com/chanzuckerberg/single-cell-curation/blob/main/cellxgene_schema_cli/cellxgene_schema/validate.py#L628) as part of the validate step, which runs before the add-labels function where this is called! 

Though, probably would still be useful to make it explicit here since it's tucked in a util function rather than in AnnDataLabelAppender

bkmartinjr: Yep, I saw that, but there is nothing that prevents a misconfiguration for `obs.index`, which would be unfortunate and lead to non-unique IDs.

Not sure how you want to handle this, but given that AnnData allows duplicate indices, and we do _not_ allow them, a bit more belt-and-suspenders might be safe...

danieljhegeman: I like Bruce's suggestion of including an `assert`

danieljhegeman: I agree, add the `assert`

Issue #418, opened 2023-04-05
cellxgene-schema must validate that raw matrix does not have more genes than raw var.


Bento007:
## Motivation

See [**#single-cell-data-wrangling**](https://czi-sci.slack.com/archives/C024HCSH9PT/p1680714440373239).

Catching datasets that will not succeed in the Discover submission pipeline before submission. This saves compute resources and speeds the job of curators to validate and submit new data.

It's possible to submit a dataset that will fail seurat conversion because the raw matrix has more genes than raw var. Here is [an example datasets in s3](https://s3.console.aws.amazon.com/s3/object/kuni-tmp?region=us-west-2&prefix=fail_bc_raw_matrix_has_more+genes_than_raw_var.h5ad).

Here is a screenshot of what is in var. 
![screenshot_2023-04-05_at_11.00.49_am.png](https://images.zenhubusercontent.com/5a4d3418a016c5b034bbeac9/38e46943-1cbd-4733-8bb6-aedc938a1dff)
The adata.raw.var is not matching up with adata.raw.X

## Definition of Done

The example datasets should fail validation.

## Tasks

- [ ] Add a check to validation, that `adata.raw.var` matches the same dimension as `adata.raw.X`
- [ ] Add a test for this error condition.

nayib-jose-gloria:
@jahilton / @corismall ready for QA! Same instructions to pull HEAD of main as other tickets

jahilton:
This is still passing validation
See the bottom setup & test in [this notebook](https://github.com/Lattice-Data/lattice-tools/blob/CXG290-schema4validator-testing/cellxgene_resources/cxg_4.0.0_testing/418_raw_matrix_vs_var.ipynb) for the example which has 22356 features in raw.X and only 22000 features in raw.var

danieljhegeman:
@jahilton ok thanks for flagging. I'll investigate.

danieljhegeman:
@jahilton can you share (a link to) the specific datafile you're using?

danieljhegeman:
@jahilton ah, to be clear, the datafile _should_ still pass validation (right @nayib-jose-gloria ?). It simply should give the _warning_ that it will fail Seurat conversion if it is submitted.

danieljhegeman:
Ok. So in those cases we want to fail validation altogether, but in other cases such as where the file is simply too large for Seurat conversion, we want to continue only warn the user.

Thanks for pointing me to the detail @brianraymor

[...]

PR #631, opened 2023-09-19
feat: fail when raw matrix var count > raw var count

danieljhegeman:
This commit escalates the response from the validator when the raw matrix variable count does not exactly equal the count of key-indexed variables in raw.var: instead of producing a warning, it produces an error and fails validation.

## Reason for Change

- #418 

## Changes

- Change warning to failure

## Testing

- Unit test updated
version at HEAD of main with 
```
```

## Notes for Reviewer

### DIFF
--- a/cellxgene_schema_cli/cellxgene_schema/validate.py
+++ b/cellxgene_schema_cli/cellxgene_schema/validate.py
@@ -802,10 +802,9 @@ class Validator:
                 self.is_seurat_convertible = False
 
         if self.adata.raw and self.adata.raw.X.shape[1] != self.adata.raw.var.shape[0]:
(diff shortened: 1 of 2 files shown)


nayib-jose-gloria (approved):
this looks good, but I noticed that Jason's [test notebook](https://github.com/Lattice-Data/lattice-tools/blob/CXG290-schema4validator-testing/cellxgene_resources/cxg_4.0.0_testing/418_raw_matrix_vs_var.ipynb) didn't print out the warning either, when it should have. 

Mind checking (either via unit test or manipulating one of the fixture h5ads in a python notebook) whether you can recreate his test case and make sure its fails validation as expected?

Issue #590, opened 2023-04-04
obsm embeddings must be validated 


brianraymor:
# Context

Placeholder until schema 4 is published. See [Update requirements for obsm embeddings](https://github.com/chanzuckerberg/single-cell-curation/issues/544). 

...

Subsequent code review also revealed that both the `cellxgene-schema CLI `and the CXG conversion process validate embeddings with some duplication, but their validation requirements are different. 

1. [Embedding validation in CXG conversion](https://github.com/chanzuckerberg/single-cell-data-portal/blame/15da6106f6742704d3233d2ab6e128b231ce7ae8/backend/layers/processing/h5ad_data_file.py#L107)
2. [Embedding validation in `cellxgene-schema CLI `](https://github.com/chanzuckerberg/single-cell-curation/blob/79a78b41414441d6f79ecf57baa57e3cea0c5eb3/cellxgene_schema_cli/cellxgene_schema/validate.py#L997)

To avoid curator dissatisfiers (part of how DP objectives are assessed), we want to ensure that validation _fast fails_ prior to the submission and upload of the dataset.

metakuni:
Moving back to `In Progress` to reflect @joyceyan 's change on Friday.

jahilton:
Believe the `unsigned integer` note is unnecessary as (our understanding of this is that) python isn't aware of signed vs unsigned, they'll just be read as int. If this is in correct, then help us understand so we know what to test for.

QA notes...
1. obsm key `X_` passes validation but _{suffix} MUST be at least one character long_
To recreate, take an existing .h5ad with embeddings ([this is a small valid example](https://github.com/Lattice-Data/lattice-tools/blob/CXG290-schema4validator-testing/cellxgene_resources/valid.h5ad)), read inn, then `adata.obsm['X_'] = adata.obsm['X_umap']`

2. obsm key `X_ ` passes validation but _{suffix} MUST be at least one character long_
I wouldn't think whitespace counted as a character, but maybe that's debatable.
`adata.obsm['X_ '] = adata.obsm['X_umap']`

3. obsm with _any_ np.nan fails validation but it should pass so long as _all_ are not np.nan
`coord1 = [4] + [np.nan] * (adata.obsm['X_umap'].shape[0] - 1)`
`coord2 = [-54] + [np.nan] * (adata.obsm['X_umap'].shape[0] - 1)`
`adata.obsm['X_umap'] = np.column_stack((coord1, coord2))`

[...]

PR #635, opened 2023-09-19
feat: fix X_{suffix} validation and add test coverage

joyceyan:
## Reason for Change

https://github.com/chanzuckerberg/single-cell-curation/issues/590

## Changes

- fixes some of the issues caught by @jahilton:
  - only fail if all values are NaN, not if any values are NaN
  - having NaN should not result in the infinity error message getting picked up
  - validate that suffixes are at least one character
  - validate that keys don't have whitespace
  - gracefully fail if values have strings, aka don't fail on the `isinf` call
- adds test coverage for all of the new validation things added

## Testing

version at HEAD of main with 
```
```

## Notes for Reviewer

### DIFF
--- a/cellxgene_schema_cli/cellxgene_schema/validate.py
+++ b/cellxgene_schema_cli/cellxgene_schema/validate.py
@@ -837,7 +837,7 @@ class Validator:
 
                 if len(key) <= 3:
                     self.errors.append(
-                        f"Embedding key in 'adata.obsm' {key} must have a suffix more than one character long."
+                        f"Embedding key in 'adata.obsm' {key} must have a suffix at least one character long."
                     )
                 if len(value.shape) < 2 or value.shape[0] != self.adata.n_obs or value.shape[1] < 2:
                     self.errors.append(
--- a/cellxgene_schema_cli/cellxgene_schema/validate.py
+++ b/cellxgene_schema_cli/cellxgene_schema/validate.py
@@ -822,6 +822,11 @@ class Validator:
 
         obsm_with_x_prefix = 0
         for key, value in self.adata.obsm.items():
+            if " " in key:
+                self.errors.append(
+                    f"Embedding key {key} has whitespace in it, please remove it."
+                )
+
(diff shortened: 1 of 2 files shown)


cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -822,6 +822,9 @@ def _validate_embedding_dict(self):
 
         obsm_with_x_prefix = 0
         for key, value in self.adata.obsm.items():
+            if " " in key:
```
joyceyan, line 825: i think any keys that have whitespace in them should be invalid, regardless of whether it's an `X_{suffix}` key

cellxgene_schema_cli/tests/test_schema_compliance.py
```diff
@@ -1272,6 +1317,28 @@ def test_obsm_values_at_least_one_X(self):
...
+        Embedding keys with whitespace are not valid
+        """
+        self.validator.adata.obsm["X_ umap"] = self.validator.adata.obsm["X_umap"]
+        self.validator.validate_adata()
+        self.assertEqual(
+            self.validator.errors,
+            ["ERROR: Embedding key X_ umap has whitespace in it, please remove it."],
+        )
+
     def test_obsm_shape(self):
```
nayib-jose-gloria, line 1342: can you add a test for the other condition on this if conditional? i.e. must have same number of rows as X

nayib-jose-gloria (commented):
looks good, thank you! flagged 1 test scenario I think we're missing (from before these additions, but would be good to get coverage on while we're changing this block)

cellxgene_schema_cli/tests/test_schema_compliance.py
joyceyan: i updated for the case where the number of rows is not the same as the number of cells. it actually seems to fail when setting the key, not when running the validation. but the test should reflect that

nayib-jose-gloria (approved):

Issue #405, opened 2023-03-15
cellxgene-schema must validate that -add-labels will not result in a write failure during ingestion


brianraymor:
There have been multiple instances where a validated dataset has been uploaded to CELLxGENE Discover and then fails during -`add-labels` with an exception like:

`
ERROR: Writing h5ad was unsuccessful, got exception 'Can't implicitly convert non-string objects to string'. Above error raised while writing key 'categories' of <class 'h5py._hl.group.Group'> to /'."
`

See [**#single-cell-data-wrangling**](https://czi-sci.slack.com/archives/C024HCSH9PT/p1675282747137719)
See [**#single-cell-data-wrangling**](https://czi-sci.slack.com/archives/C024HCSH9PT/p1677766575087969). 
See [**single-cell-data-representation**](https://cziscience.slack.com/archives/C01L65J7PLP/p1678805519018689). 

@jychien has shared a mitigation with other curators:

> I used cellxgene-schema validate --add-labels output_file input_file to trouble shoot the dataset.

**UPDATE:**
Note to assignee from Nayib:
Our own testing/QA would catch if these errors were being introduced by the add-labels step, as those changes are entirely deterministic. Most likely, these scenarios occur when the dataset is already malformed when initially submitted. The errors are triggered by the add-label step because it's the only time we attempt to write to a file, exposing some existing defect on the dataset that our validation does not check for. Consider that when trying to create a test file to validate this during QA, Jenny (see comment below) was not able to because the call to 'anndata.write' throws the aforementioned error and didn't allow her to create the test file.

I think something similar to introducing a dry-run write to a 'dummy' temp file could catch and return these errors during the validate step; however, if we go this route, we should alert users about this write+delete to a temp file while validating locally. And consider memory/performance implications.

jychien:
I am not able to create and write to file another AnnData with the same error for obs having a categorical column containing boolean objects. The same error is given when I write the AnnData to file. I have the [original h5ad](https://drive.google.com/file/d/1ixLnCUAHeV89Bp-gpu7o10-uiPwfcJkt/view?usp=sharing) file that the original curator (Batu) had come across and can test with this file. However, this h5ad file is formatted to schema 3.0, and therefore we will need to stage the testing for this ticket differently. Thanks!

Bento007:
The error mentioned above might get fix in annadata 0.10.0 https://github.com/scverse/anndata/issues/726#issuecomment-1711388396

[...]

PR #643, opened 2023-09-22
feat(4.0): Add check for conditions that would cause anndata.write to fail.

Bento007:
## Reason for Change

- #405

## Changes

- Add a check for categorical columns with none string values
- Add a check for column with mixed types.

## Testing

- Added unit test.


nayib-jose-gloria (approved):
left a suggestion on cleaning up the flag help message, otherwise this looks good! Might be good to get a thumbs up from Jason on this approach + the help text + flag name even before merging

cellxgene_schema_cli/cellxgene_schema/cli.py
```diff
@@ -32,7 +32,16 @@ def schema_cli():
 )
 @click.option("-i", "--ignore-labels", help="Ignore ontology labels when validating", is_flag=True)
 @click.option("-v", "--verbose", help="When present will set logging level to debug", is_flag=True)
-def schema_validate(h5ad_file, add_labels_file, ignore_labels, verbose):
+@click.option(
+    "-w",
+    "--write-check",
+    help="When present will run write_check on of the data. This check if the dataset "
+    "can be written back out with added labels. This is a slower, more memory and "
+    "disk intensive operation. It is not recommended to run this on dataset that will "
```
danieljhegeman, line 40:
```suggestion
    "can be written back out with added labels. This is a slower operation which requires "
    "more memory and disk space. It is not recommended to run this on dataset that will "
```

danieljhegeman (approved):

cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -644,18 +644,30 @@ def _validate_dataframe(self, df_name: str):
...
-                    continue
-                for category in col.dtype.categories:
-                    if category not in col.values:
-                        self.warnings.append(
-                            f"Column '{column}' in dataframe '{df_name}' contains a category '{category}' with "
-                            f"zero observations. These categories will be removed when `--add-labels` flag is present."
-                        )
+        for column_name in df.columns:
+            column = df[column_name]
+            if column.dtype != "category":
```
nayib-jose-gloria, line 649: curious, why check for != category? is it because that dtype would already validate that they do not have mixed values?

cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -644,18 +644,30 @@ def _validate_dataframe(self, df_name: str):
...
+                # Check for columns with mixed values.
+                value_types = {type(x) for x in column.values}
+                if len(value_types) != 1:
+                    self.errors.append(
+                        f"Column '{column_name}' in dataframe '{df_name}' cannot contained mixed types. Found {value_types}."
```
nayib-jose-gloria, line 654: nit: cannot contained --> cannot contain

cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -644,18 +644,30 @@ def _validate_dataframe(self, df_name: str):
...
+                        if category not in column.values:
+                            self.warnings.append(
+                                f"Column '{column_name}' in dataframe '{df_name}' contains a category '{category}' with "
+                                f"zero observations. These categories will be removed when `--add-labels` flag is present."
+                            )
+                # Check for columns that have none string catagories
+                catagory_types = {type(x) for x in column.dtype.categories.values}
+                if len(catagory_types) > 1 or str not in catagory_types:
+                    self.errors.append(
+                        f"Column '{column_name}' in dataframe '{df_name}' must only contain string catagories. Found {catagory_types}."
```
nayib-jose-gloria, line 669: nit catagories -> categories

nayib-jose-gloria: same w/ catagory_types var name

cellxgene_schema_cli/cellxgene_schema/validate.py
Bento007:
```suggestion
                        f"Column '{column_name}' in dataframe '{df_name}' cannot contain mixed types. Found {value_types}."
```

cellxgene_schema_cli/cellxgene_schema/validate.py
nayib-jose-gloria: nvm--I see now that it has its own type-check

cellxgene_schema_cli/cellxgene_schema/validate.py
nayib-jose-gloria: also, I think it is true that we only support str categories in our current schema definition. But are we 100% certain that will always be true? If this is an anndata constraint, mind linking me to a doc or comment specifying this?

cellxgene_schema_cli/cellxgene_schema/validate.py
Bento007: A regular ndarray doesn't require a specifing a dtype so it could have mixed values. I'm assuming that Catagorical restricts to the specified catagories which are typed, but I haven't confirmed that.

cellxgene_schema_cli/cellxgene_schema/validate.py
Bento007: This is not documented in anndata because it is a [bug](https://github.com/chanzuckerberg/single-cell-curation/issues/405#issuecomment-1732144678). Anndata should support not string values as categories and this is fixed in anndata 0.10.0.

nayib-jose-gloria: Gotcha. Mind adding a comment to this effect--maybe a TODO so we remember to remove this check for anndata 0.10.0 and/or if we get a non-string categorical field in schema 5 we remember to highlight this anndata 0.8.0 bug to product?

cellxgene_schema_cli/cellxgene_schema/cli.py
```diff
@@ -32,7 +32,15 @@ def schema_cli():
 )
 @click.option("-i", "--ignore-labels", help="Ignore ontology labels when validating", is_flag=True)
 @click.option("-v", "--verbose", help="When present will set logging level to debug", is_flag=True)
-def schema_validate(h5ad_file, add_labels_file, ignore_labels, verbose):
+@click.option(
+    "-w",
+    "--write-check",
```
Bento007, line 37: @jahilton I'd like to get your thoughts on this new flag before I merge.

Issue #515, opened 2023-06-16
cellxgene-schema must validate self_reported_ethnicity


brianraymor:
**Notes**: 

I recommend that this issue be addressed last when updating `cellxgene-schema` CLI. Its design may be simplified prior to Schema 4 delivery **IF** HANCESTRO is updated to address pending request:

1. The ontology is modeled to make it possible to identify HANCESTRO terms that are appropriate for the self reported use case. The ontologist is planning to ensure that _each term is tagged accordingly so we could infer 2 new parent classes to automatically give us a full list of geographic descriptors and of ethnicity descriptors._ Currently, CELLxGENE must assess the entire ontology and _cherry pick_ terms to block which increases the complexity of curation and validation. 

This design is based on HANCESTRO 2.8, but this issue was addressed in HANCESTRO 3.0:

1. [Preferred term label for HANCESTRO:0015](https://github.com/EBISPOT/ancestro/issues/36)

Also see [**#single-cell-data-wrangling**](https://czi-sci.slack.com/archives/C024HCSH9PT/p1689288801992369?thread_ts=1686097986.283929&cid=C024HCSH9PT).

# Design

See [schema4](https://github.com/chanzuckerberg/single-cell-curation/blob/main/schema/4.0.0/schema.md#self_reported_ethnicity_ontology_term_id)

### self_reported_ethnicity_ontology_term_id

<table>
  <tbody>
    <tr>
      <th>Key</th>
      <td>self_reported_ethnicity_ontology_term_id</td>
    </tr>
    <tr>
      <th>Annotator</th>
      <td>Curator</td>
    </tr>
    <tr>
      <th>Value</th>
      <td>
        categorical with <code>str</code> categories. If
        <code>organism_ontolology_term_id</code> is
        <code>"NCBITaxon:9606"</code> for <i>Homo sapiens</i>,
        the value MUST be formatted as one or more comma-separated (with no leading or trailing spaces) HANCESTRO
        terms in ascending lexical order or <code>"unknown"</code> if unavailable.<br><br>For example, if the terms are <code>"HANCESTRO:0014</code> and <code>HANCESTRO:0005"</code> then the value of <code>self_reported_ethnicity_ontology_term_id</code> MUST be <code>"HANCESTRO:0005,HANCESTRO:0014"</code>.<br><br>The following terms MUST NOT be used:<br /><br />
        <ul>
          <li>
            <a
[...]

PR #646, opened 2023-09-26
feat: support multiple ontology terms per cell in self-reported ethnicity

nayib-jose-gloria:
## Reason for Change

- #515

## Changes

- Validate against a list of 'forbidden' HANCESTRO terms for self-reported ethnicity
- Update error message indicating proper valid format wherever an error is detected
- introduced fields 'multi_term' and 'term_delimiter' to curie_constraints block in schema-definition yaml to distinguish between fields that should be validated only as single ontology terms vs. single ontology terms AND comma-delimited strings containing multiple ontology terms to validate 
- small refactor of validation functions to handle both single-term curie strings and multi-term curie strings while leveraging common functions for shared validation requirements (i.e. multi-term curie strings need each individual term validated as single-term curie strings)

## Testing

- unit tests

## Notes for Reviewer


cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -182,6 +182,39 @@ def _validate_curie_ontology(self, term_id: str, column_name: str, allowed_ontol
...
+
+        :param str term_str: the curie term str to validate
+        :param str column_name: Name of the column in the dataframe
+        :param dict curie_constraints: constraints for the curie term to be validated,
+        this part of the schema definition
+
+        :rtype None
+        """
+        # If there are exceptions and this is one then skip to end
+        if "exceptions" in curie_constraints and term_str in curie_constraints["exceptions"]:
```
nayib-jose-gloria, line 197: moved from '_validate_curie' because exceptions are only valid if they are the only term in the string, not if they are part of a comma-delimited str

cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -182,6 +182,39 @@ def _validate_curie_ontology(self, term_id: str, column_name: str, allowed_ontol
...
+            return
+
+        # If NA is found in allowed ontologies, it means only exceptions should be found. If no exceptions were found
+        # then return error
+        if curie_constraints["ontologies"] == ["NA"]:
```
nayib-jose-gloria, line 202: moved from '_validate_curie' to keep exception-logic together in top-level function

cellxgene_schema_cli/cellxgene_schema/write_labels.py
```diff
@@ -108,6 +108,20 @@ def _flatten_column_def_with_dependencies(self, column_def: dict) -> dict:
 
         return flatten
 
+    def _get_ontology_term_label(self, term_id: str, allowed_ontologies: List[str]) -> str:
+        """
+        TODO:
```
nayib-jose-gloria, line 113: forgot about this TODO, in process of writing docstring

cellxgene_schema_cli/cellxgene_schema/schema_definitions/schema_definition.yaml
```diff
@@ -215,15 +215,45 @@ components:
...
+                                - HANCESTRO:0557
+                                - HANCESTRO:0558
+                                - HANCESTRO:0559
+                                - HANCESTRO:0560
+                                - HANCESTRO:0561
+                                - HANCESTRO:0564
+                                - HANCESTRO:0565
+                                - HANCESTRO:0566
+                            multi_term:
+                                term_delimiter: ","
```
Bento007, line 255: suggestion: `term_delimiter` -> `delimiter`. multi_term already implies multiples

cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -182,6 +182,39 @@ def _validate_curie_ontology(self, term_id: str, column_name: str, allowed_ontol
...
+        Validate a curie str based on some constraints. If there are any errors, it adds them to self.errors
+
+        :param str term_str: the curie term str to validate
+        :param str column_name: Name of the column in the dataframe
+        :param dict curie_constraints: constraints for the curie term to be validated,
+        this part of the schema definition
+
+        :rtype None
+        """
+        # If there are exceptions and this is one then skip to end
```
Bento007, line 196: nit: alternative comment `# Account for exceptions in the constraints.`

cellxgene_schema_cli/cellxgene_schema/write_labels.py
```diff
@@ -108,6 +108,23 @@ def _flatten_column_def_with_dependencies(self, column_def: dict) -> dict:
...
+         error if term_id is not valid).
+
+        :param term_id: str single ontology term ID
+        :param allowed_ontologies: List[str] list of onotlogies to check for term_id label in
+        :return: str term label
+        """
+        for ontology_name in allowed_ontologies:
+            if ontology_name == "NA":
+                continue
+            if ONTOLOGY_CHECKER.is_valid_term_id(ontology_name, term_id):
```
Bento007, line 124:
```suggestion
            elif ONTOLOGY_CHECKER.is_valid_term_id(ontology_name, term_id):
```

cellxgene_schema_cli/tests/test_schema_compliance.py
```diff
@@ -449,16 +450,22 @@ def test_disease_ontology_term_id(self):
...
+        self.validator.adata.obs.loc[
+            self.validator.adata.obs.index[0],
+            "self_reported_ethnicity_ontology_term_id",
+        ] = "unknown"
+        self.assertTrue(self.validator.validate_adata())
+        self.assertEqual(self.validator.errors, [])
 
-        # If organism_ontolology_term_id is "NCBITaxon:9606" for Homo sapiens,
-        # this MUST be either a HANCESTRO term, "multiethnic", or "unknown" if unavailable.
+        # Test condition if organism_ontolology_term_id is "NCBITaxon:9606" for Homo sapiens
```
Bento007, line 468: Split this into a separate tests

cellxgene_schema_cli/tests/test_schema_compliance.py
```diff
@@ -469,9 +476,30 @@ def test_self_reported_ethnicity_ontology_term_id(self):
...
-                "'multiethnic' if more than one ethnicity is reported, or 'unknown' if unavailable."
+                "not a valid ontology term id of 'HANCESTRO'. When 'organism_ontology_term_id' "
+                "is 'NCBITaxon:9606' (Homo sapiens), self_reported_ethnicity_ontology_term_id MUST "
+                "be formatted as one or more comma-separated (with no leading or trailing spaces) "
+                "HANCESTRO terms in ascending lexical order, or 'unknown' if unavailable. Cannot "
+                "match any forbidden HANCESTRO terms listed in schema definition."
+            ],
+        )
+
+        # Test Forbidden Term is not allowed
```
Bento007, line 487: Split into separate test

cellxgene_schema_cli/tests/test_schema_compliance.py
```diff
@@ -497,6 +525,112 @@ def test_self_reported_ethnicity_ontology_term_id(self):
             ],
         )
 
+    def test_self_reported_ethnicity_ontology_term_id__multi_term(self):
+        """
+        self_reported_ethnicity_ontology_term_id categorical with str categories. Tests invalid scenarios involving
+        multiple delimited terms in a single str
+        """
+        # Test 'unknown' term in list of otherwise valid HANCESTRO terms
```
Bento007, line 533: Split into separate tests

Bento007 (changes requested):
Great work. Please split the tests up and this will be good to go.

cellxgene_schema_cli/cellxgene_schema/validate.py
danieljhegeman: IMO this is a line of code that does not need a comment; comments shouldn't describe what the code does literally, only unseen product requirements, exceptional behaviors, etc.

cellxgene_schema_cli/cellxgene_schema/validate.py
danieljhegeman: I'm a little confused wrt `"NA"` here. I do not see reference to `"NA"` anywhere in the schema, but I do see reference to `"na"`. I just want to verify that you are intentionally using `"NA"` in some cases and `"na"` in others, esp since `"na"` is specified in the reqs for this issue, but it is less common to see it as lowercase.

danieljhegeman: Ah. I see this has been specified differentially in the schema. Nvm.

cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -182,6 +182,39 @@ def _validate_curie_ontology(self, term_id: str, column_name: str, allowed_ontol
...
+            return
+
+        # If NA is found in allowed ontologies, it means only exceptions should be found. If no exceptions were found
+        # then return error
+        if curie_constraints["ontologies"] == ["NA"]:
+            self.errors.append(f"'{term_str}' in '{column_name}' is not a valid value of '{column_name}'.")
+            return
+
+        # if multi_term is defined, split str into individual ontology terms and validate each
+        if curie_constraints.get("multi_term", None):
```
danieljhegeman, line 207:
```suggestion
        if "multi_term" in curie_constraints:
```

cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -182,6 +182,39 @@ def _validate_curie_ontology(self, term_id: str, column_name: str, allowed_ontol
...
+            delimiter = curie_constraints["multi_term"]["term_delimiter"]
+            term_ids = term_str.split(delimiter)
+            for term_id in term_ids:
+                self._validate_curie(term_id, column_name, curie_constraints)
+            if curie_constraints["multi_term"].get("sorted", False) and len(term_ids) > 1:
```
danieljhegeman, line 212:
```suggestion
            if "sorted" in curie_constraints["multi_term"] and len(term_ids) > 1:
```

danieljhegeman (commented):
recommend incorporating [this small change](https://github.com/chanzuckerberg/single-cell-curation/pull/650/files) into the testing code

cellxgene_schema_cli/cellxgene_schema/validate.py
nayib-jose-gloria: it is intentional, but not introduced in this PR. This block is being moved from a different function to factor out some shared checks. We can make it consistent, though offhand I'm not sure what the original reason for making them different was.

cellxgene_schema_cli/cellxgene_schema/validate.py
nayib-jose-gloria: I think if we're going to introduce a boolean field, we should consider that someone might try to assign `sorted: False` in the future and account for that in validation

Bento007 (approved):

cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -230,20 +268,28 @@ def _validate_curie(self, term_id: str, column_name: str, curie_constraints: dic
...
-            self.errors.append(f"'{term_id}' in '{column_name}' is not allowed.")
-            return
-
         # Check that term id belongs to allowed ontologies
-        self._validate_curie_ontology(term_id, column_name, curie_constraints["ontologies"])
+        is_valid_term = self._validate_curie_ontology(term_id, column_name, curie_constraints["ontologies"])
 
         # If the term id does not belong to an allowed ontology, the subsequent checks are redundant
-        if self.errors:
+        if not is_valid_term:
```
Bento007, line 275: nit: `is_valid_term` is not used anywhere else.
```suggestion
        if not self._validate_curie_ontology(term_id, column_name, curie_constraints["ontologies"]):
```

Bento007 (changes requested):

cellxgene_schema_cli/cellxgene_schema/schema_definitions/schema_definition.yaml
```diff
@@ -226,31 +228,34 @@ components:
...
+                                    - HANCESTRO:0555
+                                    - HANCESTRO:0557
+                                    - HANCESTRO:0558
+                                    - HANCESTRO:0559
+                                    - HANCESTRO:0560
+                                    - HANCESTRO:0561
+                                    - HANCESTRO:0564
+                                    - HANCESTRO:0565
+                                    - HANCESTRO:0566
+                                ancestors_inclusive:    # these terms AND their children are forbidden
```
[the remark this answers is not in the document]
nayib-jose-gloria: we use 'ancestor' / 'ancestor_inclusive' elsewhere in the schema definition to refer to listing out parent terms, so I'd like to stay consistent. I don't mind switching over everywhere, though I read it as we are listing out the forbidden ancestors (and the '_inclusive' means, in this case, we are counting the term as an ancestor of itself, but we may not always want to do that). Maybe we can re-word the comment instead, but let me know if you think its more clear using 'descendants' instead.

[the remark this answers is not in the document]
Bento007: - it would be easier to to understand if it changed to parent/child(which is more commonly understood programmatically) vs ancestors/descendants. 
- `ancestors_inclusive` sounds like this term and it's ancestors are forbidden.

Is it invalid for a term to be in both `ancestors_inclusive` and `term`? We may need a tests case to know the behavior and possible a schema check to avoid accidentally producing an invalid schema.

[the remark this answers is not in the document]
Bento007: could we drop the `ancestor` part?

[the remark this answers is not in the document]
nayib-jose-gloria: We control the yaml format, so we could re-work it so that terms can be in both term and ancestor. But as written, yes, it would be invalid for a term to be in both. 

We use ancestor/descendant because its in reference to ontologies, which use that terminology. 

We definitely need to keep either ancestor/descendant or parent/child, because we need a way to differentiate between 

1) terms that are forbidden but its children are allowed
2) terms that are forbidden AND its children are forbidden
3) terms that are allowed BUT its children are forbidden 

While not a requirement, it would be nice for that naming convention to be consistent with the opposite use-case which is already using ancestor/ancestor_inclusive, which is whitelists of:

1) terms that are allowed but its children are not
2) terms that are allowed AND its children are allowed
3) terms that are not allowed BUT its children are allowed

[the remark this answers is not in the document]
nayib-jose-gloria: What do you think of simplifying it to forbidden terms and forbidden ancestors? and if its both, the term should just be listed in both.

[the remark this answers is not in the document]
nayib-jose-gloria: it would require a refactor of the whitelist version, which already uses ancestors_inclusive, but that's fine if its clearer overall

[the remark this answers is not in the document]
Bento007: That would make more sense.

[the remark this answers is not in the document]
nayib-jose-gloria: the refactor to make that happen + make the allowlist consistent with that pattern as well was more involved than I anticipated. Pushed it, let me know what you think. If the review around this becomes any more complex I may suggest reverting the change and taking care of it in a separate PR instead

Bento007 (approved):
Looks great. Thank for refactoring.

Issue #610, opened 2023-08-24
cellxgene-schema CLI must validate ndarray size in obsm, obsp, varm, and varp


brianraymor:
## `obsm` (Embeddings)

The size of the ndarray stored for a key in `obsm` MUST NOT be zero.

## `obsp`

The size of the ndarray stored for a key in `obsp` MUST NOT be zero.
<br>

## `varm`

The size of the ndarray stored for a key in `varm` MUST NOT be zero.
<br>

## `varp`
The size of the ndarray stored for a key in `varp` MUST NOT be zero.
<br>

Bento007:
I assume "a key" means any key?

brianraymor:
> I assume "a key" means any key?

Yes since this causes RDS conversion failures - https://github.com/chanzuckerberg/single-cell-curation/issues/597#issue-1842192464

Bento007:
Do we have an existing test case dataset for any of these?

brianraymor:
Not to my knowledge. When I was playing around with it, I realized that it also needed to meet other anndata requirements like obsm embeddings must be the same length as obs like:

```
adata.obsm['test'] = numpy.zeros((adata.n_obs, 0))
adata.obsm['test'].size 
```

Bento007:
@jahilton thank you for these [notebooks](https://github.com/Lattice-Data/lattice-tools/blob/CXG290-schema4validator-testing/cellxgene_resources/cxg_4.0.0_testing/610_array_size.ipynb). They have been invaluable in reproducing the errors.

Bento007:
@jahilton @corismall read for QA

jahilton:
LGTM - [QA notebook](https://github.com/Lattice-Data/lattice-tools/blob/CXG290-schema4validator-testing/cellxgene_resources/cxg_4.0.0_testing/610_array_size.ipynb)

PR #649, opened 2023-09-26
feat(4.0): Check for empty ndarrays in varm, varp, obsm, and obsp

Bento007:
## Reason for Change

- #610
- catch cases where an empty ndarray is found in varm, varp, obsm, and obsp of the h5ad.

## Changes

- replace embedding_dict with annotation_mapping type schema_definition.yaml. It is better description of what this type based on anndatas description of varm, obsm, varp, and obsp.
- move obsm specific validation steps to `Validator._validate_obsm`
- add `Validator._validate_annotation_mapping` to check for ndarray with size 0.
- Change `component` to `component_name` in `Validator._deep_check`. This is a better description of the variable because it is the name of the component.
- Fix the error message for "Unexpected component type". This would have failed if it were ever hit because `component_name` formerly `component` is a string and does not have a ['type']

## Testing

- run unit tests

## Notes for Reviewer


cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -1274,25 +1282,29 @@ def _deep_check(self):
...
+                self._validate_dict(component, component_name, component_def)
+                if component_name == "uns":
+                    self._validate_colors_in_uns_dict(component)
+            elif component_def["type"] == "annotation_mapping":
+                self._validate_annotation_mapping(component_name, component)
+                if component_name == "obsm":
+                    self._validate_obsm()
             else:
-                raise ValueError(f"Unexpected component type '{component['type']}'")
+                raise ValueError(f"Unexpected component type '{component_def['type']}'")
```
nayib-jose-gloria, line 1306: good catch

cellxgene_schema_cli/tests/test_schema_compliance.py
```diff
@@ -1519,6 +1531,64 @@ def test_obsm_shape_same_rows_and_columns(self):
             self.validator.adata.obsm["X_umap"] = three_row_array
             self.validator.validate_adata()
 
+    def test_obsm_size_zero(self):
+        """
+        size of obsm an key cannot be zero.
+        """
+        adata = self.validator.adata
+        adata.obsm["badsize"] = numpy.empty((2, 0))
+        self.validator.adata = self.save_and_read_adata(adata)
```
nayib-jose-gloria, line 1540: why do we need to do this? Does doing the update in-memory ala `self.validator.adata.obsm["badsize"] = numpy.empty((2, 0))` not work for some reason? It works for other tests

nayib-jose-gloria (commented):
Couple of questions, otherwise looks good! 

I also made a PR to fix the unclear 'required: null' == 'is required' attribute in our schema def. I pointed the PR to this commit because I think it would otherwise cause merge conflicts. LMK if you're cool with merging it in to this PR if you feel good about the changes https://github.com/chanzuckerberg/single-cell-curation/pull/651

cellxgene_schema_cli/tests/test_schema_compliance.py
Bento007: The problem is whatever anndata is doing when it saves. Otherwise there is no way to create numpy array with size 0.

cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -1274,25 +1282,29 @@ def _deep_check(self):
...
+            logger.debug(f"Validating component: {component_name}")
+            component = getattr_anndata(self.adata, component_name)
+
             # Skip if component does not exist: only useful for adata.raw.var
-            if getattr_anndata(self.adata, component) is None:
+            if component is None:
+                # Check for required components
                 if "required" in component_def:
-                    self.errors.append(f"'{component}' is missing from adata and it's required.")
+                    pass
```
Bento007, line 1293: I didn't even realize I skipped this. It goes to show that we have no test for this condition.
```suggestion
                    self.errors.append(f"'{component}' is missing from adata and it's required.")
```

nayib-jose-gloria (approved):

PR #653, opened 2023-09-28
fix: rename all instances of 'format' variable to 'matrix_format'

danieljhegeman:
- 'format' is a built-in Python function; avoid shadowing this
- corrects one bug where an assert was testing against this built-in `format` because the previous appearances of the variable were already named `matrix_format`

### DIFF
--- a/cellxgene_schema_cli/cellxgene_schema/validate.py
+++ b/cellxgene_schema_cli/cellxgene_schema/validate.py
@@ -291,9 +291,9 @@ class Validator:
         logger.debug(f"Counting non-zero values in {matrix_name}")
 
         nnz = 0
-        format = get_matrix_format(self.adata, matrix)
+        matrix_format = get_matrix_format(self.adata, matrix)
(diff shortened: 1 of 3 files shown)


cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -898,7 +898,7 @@ def _validate_sparsity(self):
             matrix_format = get_matrix_format(self.adata, x)
             if matrix_format == "csr":
                 continue
-            assert format != "unknown"
+            assert matrix_format != "unknown"
```
danieljhegeman, line 901: @bkmartinjr tagging you as the (un)lucky introducer of this line only because you might have context for ramifications that are not immediately apparent to me, i.e., in case we were ever actually relying on that assert to fail. 🙏

bkmartinjr (approved):
Seems reasonable to me.

Unrelated to the PR approval:  suggest adopting mypy in the CI, to help detect these types of error.

PR #654, opened 2023-09-28
feat: verify all obsm keys start with X_ suffix

joyceyan:
## Reason for Change

After lots of discussion on this ticket: https://github.com/chanzuckerberg/single-cell-curation/issues/590 we've decided that we want to validate that all obsm keys start with the `X_` suffix, but only raise a warning. If an embedding key doesn't start with `X_` and also has other validation failures (ex: not an array), we would want to surface all of these but as warnings.

## Changes

- We check to see if an obsm key starts with `X_`. If not, we raise a warning that the key doesn't start with `X_`.
- Run all subsequent validations against the key, but raise a warning if it doesn't start with `X_` and an error if it does start with `X_`
- I'm keeping the requirement that at least one obsm key must start with `X_` to raise a validation error if there are no valid obsm keys

## Testing

added a new test `test_obsm_values_must_start_with_X`

## Notes for Reviewer

### DIFF
--- a/cellxgene_schema_cli/cellxgene_schema/validate.py
+++ b/cellxgene_schema_cli/cellxgene_schema/validate.py
@@ -1025,41 +1025,44 @@ class Validator:
 
         obsm_with_x_prefix = 0
         for key, value in self.adata.obsm.items():
-            # Checks for invalid keys
             if " " in key:
                 self.errors.append(f"Embedding key {key} has whitespace in it, please remove it.")
-            if len(key) <= 3:
-                self.errors.append(
-                    f"Embedding key in 'adata.obsm' {key} must start with X_ and have a suffix at least one character long."
-                )
-            if not key.startswith("X_"):
-                self.errors.append(f"Embedding key in 'adata.obsm' {key} does not start with X_")
-            else:
-                obsm_with_x_prefix += 1
 
+            issue_list = self.errors
+            if key.startswith("X_"):
+                obsm_with_x_prefix += 1
+                if len(key) <= 3:
+                    self.errors.append(
+                        f"Embedding key in 'adata.obsm' {key} must start with X_ and have a suffix at least one character long."
+                    )
+            else:
+                self.warnings.append(f"Embedding key in 'adata.obsm' {key} does not start with X_")
+                issue_list = self.warnings
+
+            # For all subsequent checks, we want to raise an error if it's an X_ embedding key, and a warning otherwise
             if not isinstance(value, np.ndarray):
-                self.errors.append(
+                issue_list.append(
                     f"All embeddings have to be of 'numpy.ndarray' type, " f"'adata.obsm['{key}']' is {type(value)}')."
                 )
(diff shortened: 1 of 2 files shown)


cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -1027,38 +1029,41 @@ def _validate_obsm(self):
...
-
                 if len(key) <= 3:
                     self.errors.append(
-                        f"Embedding key in 'adata.obsm' {key} must have a suffix at least one character long."
-                    )
-                if len(value.shape) < 2 or value.shape[0] != self.adata.n_obs or value.shape[1] < 2:
-                    self.errors.append(
-                        f"All embeddings must have as many rows as cells, and at least two columns."
-                        f"'adata.obsm['{key}']' has shape of '{value.shape}'."
+                        f"Embedding key in 'adata.obsm' {key} must start with X_ and have a suffix at least one character long."
```
atarashansky, line 1037:
```suggestion
                        f"Embedding key in 'adata.obsm' {key} must have a suffix at least one character long."
```

cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -1027,38 +1029,41 @@ def _validate_obsm(self):
             if " " in key:
                 self.errors.append(f"Embedding key {key} has whitespace in it, please remove it.")
```
atarashansky, line 1030: Is this a requirement for embeddings that do not start with "X_"?

atarashansky (approved):

cellxgene_schema_cli/cellxgene_schema/validate.py
joyceyan: My understanding from the discussion on the ticket was that whitespace should never be allowed in any embedding key, but that definitely isn't clear from the schema definition. I'll make sure to bring it up with Signe/Brian/Jason but I am reasonably confident from an implementation perspective we don't want to allow whitespace characters in any embedding key.

brianraymor: As written, it's only for X_ - cases that **might** be presented in Explorer.  I don't see why we would enforce naming conventions on embeddings intended for use in other scenarios.

brianraymor: CC: @jahilton 

Also - the warnings were an implementation detail rather than a schema requirement per the top-level note:

**Note: @jahilton has requested that obsm embeddings that do not start with X_ also be validated against the CXG requirements in X_{suffix} and generate warnings that include the embedding name. This is not a schema requirement because additional embeddings may be present for [other scenarios](https://github.com/chanzuckerberg/single-cell-curation/issues/590#issuecomment-1735947750), but an implementation requirement to provide information to the curators to assess the state of embeddings in the dataset.  @joyceyan has noted this in #654.**

Issue #614, opened 2023-08-30
cellxgene-schema CLI must validate raw matrices


brianraymor:
With the exception of Accessibility assays, the requirements for raw matrices have been updated to include:

Each cell MUST contain at least one non-zero value. All non-zero values MUST be positive integers converted to `float`.

## `X` (Matrix Layers)

The following table describes the matrix data and layers requirements that are **assay-specific**. If an entry in the table is empty, the schema does not have any other requirements on data in those layers beyond the ones listed above.

| Assay | "raw" required? | "raw" location | "normalized" required? | "normalized" location |
|-|-|-|-|-|
| scRNA-seq (UMI, e.g. 10x v3) | REQUIRED. Values MUST be de-duplicated molecule counts. Each cell MUST contain at least one non-zero value. All non-zero values MUST be positive integers stored as `numpy.float32`.| `AnnData.raw.X` unless no "normalized" is provided, then `AnnData.X` | STRONGLY RECOMMENDED | `AnnData.X` |
| scRNA-seq (non-UMI, e.g. SS2) | REQUIRED. Values MUST be one of read counts (e.g. FeatureCounts) or  estimated fragments (e.g. output of RSEM). Each cell MUST contain at least one non-zero value. All non-zero values MUST be positive integers stored as `numpy.float32`. | `AnnData.raw.X` unless no "normalized" is provided, then `AnnData.X` | STRONGLY RECOMMENDED | `AnnData.X` |
| Accessibility (e.g. ATAC-seq, mC-seq) | NOT REQUIRED | | REQUIRED | `AnnData.X` | STRONGLY RECOMMENDED |
|||||

danieljhegeman:
@brianraymor clarifying some language here

> Each cell MUST contain...

"cell" is a biological cell, i.e., each row of data MUST contain...

danieljhegeman:
Additionally, you have code-blocked the word float as `float` in the description, which might be intended to specify the Python primitive class `float`...? Was that intentional?

Or is the requirement just the the values be some sort of (presumably relatively inter-operable) float, including but not limited to `float`, `numpy.float32`, `numpy.float64`, etc...? I see that `numpy.float32` has been specified for the two assays...

danieljhegeman:
and does `STRONGLY RECOMMENDED` mean it is expected that we produce a _warning_ if this condition is not met?

brianraymor:
[...]

PR #658, opened 2023-10-02
feat: validate raw matrix values for datasets with RNA assay types

nayib-jose-gloria:
## Reason for Change

- #614

## Changes

- leverage existing _is_raw matrix validation method to include additional raw matrix value validations (doing so in this existing method means we avoid an additional costly loop through the matrix)
- rename _is_raw method to _has_valid_raw for clarity 
- fix tests and test fixtures to meet new raw matrix requirements 

## Testing

- unit tests
- TODO: add new unit tests

## Notes for Reviewer
- _has_valid_raw method is already gated by conditions that ensure only raw matrices from datasets with rna-based assays are validated against these rules 
- [Jason expressed a preference](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/single-cell-curation/614) to fail early on the first detected failed value and row for a raw matrix validation rule rather than attempt to find every row/value where a validation error occurs


cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -1132,21 +1132,46 @@ def _is_raw(self, force: bool = False) -> bool:
...
             self._raw_layer_exists = True
+            has_nonzero_in_every_row = True
+            all_nonzero_valid = True
             for matrix_chunk, _, _ in self._chunk_matrix(x):
-                data = matrix_chunk if isinstance(matrix_chunk, np.ndarray) else matrix_chunk.data
-                if (data % 1 > 0).any():
-                    self._raw_layer_exists = False
+
+                if has_nonzero_in_every_row:
+                    if matrix_format in ("csr", "csc", "coo"):  # full set of sparse matrix types
```
Bento007, line 1148: - nit: make reusable `("csr", "csc", "coo")` 
- `matrix_format in ("csr", "csc", "coo")` this can be move outside the loop and result used here. Replace 
  `matrix_format = get_matrix_format(self.adata, x)` 
   with 
   `sparse_matrix = get_matrix_format(self.adata, x) in ("csr", "csc", "coo")`

cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -1132,21 +1132,46 @@ def _is_raw(self, force: bool = False) -> bool:
...
+                            has_nonzero_in_every_row = False
+                    elif matrix_format == "dense" and not all(np.apply_along_axis(np.any, axis=1, arr=matrix_chunk)):
+                        has_nonzero_in_every_row = False
+
+                if all_nonzero_valid:
+                    data = matrix_chunk if isinstance(matrix_chunk, np.ndarray) else matrix_chunk.data
+                    if np.any((data % 1 > 0) | (data < 0)):
+                        all_nonzero_valid = False
+
+                if not has_nonzero_in_every_row and not all_nonzero_valid:
```
Bento007, line 1160: This is a very logic heavy statement with double negative (not nonzero). It would be easier to understand with positive logic. If has a value in every row and all values are valid. Value is kind of vague and doesn't imply not zero. We could also try reversing the logic in the `has_nonzero_in_every_row = True` to `has_row_of_zeros = False` same goes for `all_nonzero_valid`->`invalid_nonzero_value`

```suggestion
                if has_row_of_zeros or invalid_nonzero_value):
```

cellxgene_schema_cli/tests/test_schema_compliance.py
```diff
@@ -780,30 +787,6 @@ def test_self_reported_ethnicity_ontology_term_id__duplicates_in_multi_term(self
             ],
         )
 
-    def test_self_reported_ethnicity_ontology_term_id__multi_term_list(self):
```
Bento007, line 783: was this removed intentionally?

Bento007 (changes requested):

cellxgene_schema_cli/tests/test_schema_compliance.py
nayib-jose-gloria: didn't mean to push, I will re-add when I merge in the pytest PR

cellxgene_schema_cli/cellxgene_schema/validate.py
nayib-jose-gloria: re-worked, should be clearer now

cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -1112,41 +1119,67 @@ def _get_raw_x_loc(self) -> str:
...
+                    # else, must be dense matrix, confirm that all rows have at least 1 nonzero value
+                    elif not all(np.apply_along_axis(np.any, axis=1, arr=matrix_chunk)):
+                        has_row_of_zeros = True
+
+                if not has_invalid_nonzero_value:
+                    data = matrix_chunk if isinstance(matrix_chunk, np.ndarray) else matrix_chunk.data
+                    if np.any((data % 1 > 0) | (data < 0)):
+                        has_invalid_nonzero_value = True
+
+                if has_row_of_zeros and has_invalid_nonzero_value:
```
Bento007, line 1167:
```suggestion
                if has_row_of_zeros or has_invalid_nonzero_value:
```

Bento007 (changes requested):

cellxgene_schema_cli/cellxgene_schema/validate.py
nayib-jose-gloria: I decided to ask Jason about this in the [issue](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/single-cell-curation/614), since I'm not sure if we want to keep validating for one rule if the other is already found to cause an error. If he says to fail immediately on violation of either, we can even get rid of the flags altogether and just `break` once we encounter either error

nayib-jose-gloria: https://github.com/chanzuckerberg/single-cell-curation/issues/614#issuecomment-1743863137

Looks like we want to keep the `and`, so we validate the other rule even if one has already failed, and we break if both rules have errored

Bento007 (approved):
The code is good. It looks like there are some issue with your branch and main.

cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -1119,41 +1119,67 @@ def _get_raw_x_loc(self) -> str:
...
-        :return False if at least one value is not an integer, True otherwise
         """
         if force:
             self._raw_layer_exists = None
 
         if self._raw_layer_exists is None:
             # Get potential raw_X
-            raw_loc = self._get_raw_x_loc()
-            x = self.adata.raw.X if raw_loc == "raw.X" else self.adata.X
+            x = self.adata.raw.X if self._get_raw_x_loc() == "raw.X" else self.adata.X
```
danieljhegeman, line 1140: Should use `.get_raw_x()` instead of running logic here with `_get_raw_x_loc()`. It already does this for you.

cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -1119,41 +1119,67 @@ def _get_raw_x_loc(self) -> str:
...
+                self.errors.append("Raw matrix values must have type numpy.float32.")
+                return self._raw_layer_exists
 
             matrix_format = get_matrix_format(self.adata, x)
             assert matrix_format != "unknown"
             self._raw_layer_exists = True
+            has_row_of_zeros = False
+            has_invalid_nonzero_value = False
+            is_sparse_matrix = matrix_format in SPARSE_MATRIX_TYPES
             for matrix_chunk, _, _ in self._chunk_matrix(x):
```
danieljhegeman, line 1152: @nayib-jose-gloria I think we have a problem here cc @Bento007 

This is returning a _chunk_ of the matrix, therefore a possible subset of all values. But the logic in the loop appears to be treating it as if it comprises _all_ values. We're comparing the n_obs of the entire dataset to the number of rows represented in the chunk, and if they aren't equal we're asserting that a row of zeros must exist. This seems like a mistake to me.

nayib-jose-gloria: great catch--I believe this wouldn't have been caught in QA because the chunk size is 10,000 rows and the test datasets are all smaller than that. 

I can make the fix

danieljhegeman: >this wouldn't have been caught in QA because the chunk size is 10,000 rows and the test datasets are all smaller than that

Yeah I had suspected that would be the case.

Btw do you know much about the `_chunk_matrix` function? Does using the `[]` indexing notation to read off of a matrix not actually load the data into mem until you make that call? @bkmartinjr 's comments in the docstring about "read fast-path" are something I'm not familiar with.

danieljhegeman: @nayib-jose-gloria you might consider pulling in the logic loop changes from my original feature branch. Or did you read over it and decide it needed to be overhauled..? let me know

danieljhegeman: I did this somewhere else, I think for seurat convertability, but you should make the default chunk size in `_chunk_matrix` a variable that you can patch in the unit tests. That way you can test compatibility with datasets whose size is greater than the chunk size while still using the same test dataset(s).

PR #659, opened 2023-10-02
fix: attempt to write bad dataset as part of the write tests

Bento007:
## Reason for Change

- Eventually this test and check will no longer be needed after upgrading to anndata 0.10.0+. This add a check if the bug is fixed and if the related check can be removed. 

## Changes

- add write_h5ad to tests if the anndata write error still exists.

## Testing

- updated unit tests

## Notes for Reviewer


nayib-jose-gloria (approved):
good idea

PR #660, opened 2023-10-03
fix(raw matrix validation): compare number of nonzero rows in matrix chunk against matrix chunk size, not full matrix size

nayib-jose-gloria:
## Reason for Change

- #614 
- Fix bug introduced in previous PR for this issue

## Changes

- check that number of nonzero rows in a matrix chunk equals matrix chunk size rather than full matrix size
- small fix to DRY up code when fetching raw matrix in has_valid_raw
- added test

## Testing

- unit tests

## Notes for Reviewer

### DIFF
--- a/cellxgene_schema_cli/cellxgene_schema/validate.py
+++ b/cellxgene_schema_cli/cellxgene_schema/validate.py
@@ -1137,7 +1137,7 @@ class Validator:
 
         if self._raw_layer_exists is None:
             # Get potential raw_X
-            x = self.adata.raw.X if self._get_raw_x_loc() == "raw.X" else self.adata.X
+            x = self._get_raw_x()
             if x.dtype != np.float32:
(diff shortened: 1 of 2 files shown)


Bento007 (commented):
Is there also a tests where it fails for one condition an not the other?

nayib-jose-gloria:
> Is there also a tests where it fails for one condition an not the other?

Sorry, can you clarify? Which condition?

Bento007:
For these conditions:
`
[
            "ERROR: All non-zero values in raw matrix must be positive integers of type numpy.float32.",
            "ERROR: Raw data is missing: there is only a normalized matrix in X and no raw.X"
]
`
I realize this is outside the scope of this PR, but those tests don't exists we should add them.

Bento007 (approved):

Issue #513, opened 2023-06-16
{column}_colors must be validated


brianraymor:
**Note**: Per Arathi's direction in [#single-cell-planning](https://czi-sci.slack.com/archives/C04PHE5TV9B/p1683819000421289):

 **_Any validation code against the CXG schema will also be owned by the Data Visualization team._**

# Design

### {column}_colors

<table><tbody>
  <tr>
    <th>Key</th>
      <td>
        {column}_colors where {column} MUST be the name of a <code>category</code> data type column in <code>obs</code> that<br> is annotated by the data submitter or curator. The following columns that are annotated by CELLxGENE<br> Discover MUST NOT be specified as {column}:<br><br>
      <ul>
        <li>assay</li>
        <li>cell_type</li>
        <li>development_stage</li>
        <li>disease</li>
        <li>organism</li>
        <li>self_reported_ethnicity</li>
        <li>sex</li>
        <li>tissue</li>       
      </ul><br>
      Instead annotate {column}_ontology_term_id_colors for these columns such as <code>assay_ontology_term_id</code>.<br><br>
    </td>
  </tr>
  <tr>
    <th>Annotator</th>
    <td>Curator</td>
  </tr>
  <tr>
    <th>Value</th>
      <td>
        <code>numpy.ndarray</code>. This MUST be a 1-D array of shape <code>(, c)</code>, where <code>c</code> is greater than or equal to the<br> number of unique categories in the {column} as calculated by:<br><br>
           <samp>len(anndata.obs.{column}.unique())</samp><br><br>
        The color code at the Nth position in the <code>ndarray</code> corresponds to the Nth category of <samp>anndata.obs.{column}.unique()</samp>.<br><br>For example, if <code>cell_type_ontology_term_id</code> includes two unique categories:<br><br>
        <samp>anndata.obs.cell_type_ontology_term_id.unique()</samp><br><br>
        <samp>['CL:0000057', 'CL:0000115']<br>Categories (2, object): ['CL:0000057', 'CL:0000115']</samp><br><br>then <code>cell-type_ontology_term_id_colors</code> MUST contain two or more colors such as:<br><br>
[...]

PR #666, opened 2023-10-06
feat: update how we check for existing obs categorical columns

joyceyan:
## Reason for Change

- based on discussion [here](https://github.com/chanzuckerberg/single-cell-curation/issues/513), we should be considering the obs column's dtype directly, instead of what we currently do, which is checking the schema def

## Changes

- check for categorical obs column by looking at the column's dtype instead of the column definition

## Testing

- wrote a test case where we have a new categorical column added to `obs[test_column]` that's not part of the schema def, and has valid colors in `uns[test_column_colors]`. this should pass validation, despite the column not existing in the schema def

## Notes for Reviewer

### DIFF
--- a/cellxgene_schema_cli/cellxgene_schema/validate.py
+++ b/cellxgene_schema_cli/cellxgene_schema/validate.py
@@ -689,7 +689,7 @@ class Validator:
 
         for column_name in df.columns:
             column = df[column_name]
-            if column.dtype != "category":
+            if column.dtype.name != "category":
                 # Check for columns with mixed values, which is not supported by anndata 0.8.0
                 # TODO: check if this can be removed after upgading to anndata 0.10.0
                 value_types = {type(x) for x in column.values}
@@ -744,7 +744,7 @@ class Validator:
         # Check for categorical dtypes in the dataframe directly
         for column_name in df.columns:
             column = df[column_name]
-            if column.dtype == "category":
+            if column.dtype.name == "category":
                 category_mapping[column_name] = column.nunique()
 
         for key, value in uns_dict.items():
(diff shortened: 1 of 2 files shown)


cellxgene_schema_cli/tests/test_schema_compliance.py
```diff
@@ -1580,15 +1591,15 @@ def test_colors_not_numpy_array(self, validator_with_adata):
 
     def test_colors_bool_obs_counterpart(self, validator_with_adata):
         validator = validator_with_adata
-        validator.adata.uns["is_primary_data_colors"] = ["green", "purple"]
+        validator.adata.uns["is_primary_data_colors"] = numpy.array(["green", "purple"])
```
joyceyan, line 1594: not related to the primary change, but i updated these two test cases to be numpy arrays since while they should fail for having invalid `obs` counterparts, the colors should be valid numpy arrays

cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -737,19 +737,15 @@ def _validate_dataframe(self, df_name: str):
...
-                if column_name not in df.columns:
-                    # Skip this, dataframe validation should already append an error for this
-                    continue
-
-                if column_def.get("type") == "categorical":
-                    category_mapping[column_name] = df[column_name].nunique()
+        # Check for categorical dtypes in the dataframe directly
+        for column_name in df.columns:
+            column = df[column_name]
+            if column.dtype == "category":
```
nayib-jose-gloria, line 747: nit: use `column.dtype.name` for comparison, as this comparison has weird edge cases that throw errors instead of failing gracefully in older versions of numpy (source: https://stackoverflow.com/questions/26924904/check-if-dataframe-column-is-categorical#comment123655621_26925340)

nayib-jose-gloria:
```suggestion
            if column.dtype.name == "category":
```

joyceyan: makes sense - i also updated the piece of code i copied and pasted this snippet from 😅

nayib-jose-gloria (approved):
ty!

Issue #572, opened 2023-07-25
Determine GENCODE migration mappings


jahilton:
We need to investigate what Ensembl IDs should be updated from the current schema version (GENCODE v38) to a new Ensembl ID reflected in the next pinned annotation version in the schema (most recent is [currently v44](https://www.gencodegenes.org/human/)). And we need to establish a reproducible way to identify these version-to-version updates to 1) use again during future migrations and 2) establish some mappings in curation (as contributors will bring us data from older versions).

Plan: use v38 & v44 as our test case. Take a close look at what v38 gene IDs are not in v44, but we believe are represented by updated gene IDs. Design logic around determining those updates. Share the v38/v44 mapping for migration. Apply that logic to previous versions-to-v44 and include those mappings to our curation notebook so we are saving as much as possible during curation.

brianraymor:
@jahilton notes _ENSG migration - hoping to have a proposal ready for review **Sept 8th**_

jahilton:
The curation team is proposing a migration solution that does not require manual decisions to be made so that we ensure an unbiased solution that is simple enough to be followed & reproduced by users. We feel this simplicity is vital to ensuring provenance, and prevents the curation team from the burden of a tedious evaluation that will be difficult to estimate time needed.
For example, there are cases where 2 gene IDs have been collapsed to 1, thus 1 of the 2 gene IDs is deprecated while the other remains, and the curation team may "wish" the deprecated gene ID would have been the one to retain, but we will defer to the annotation resources.

We recommend 2-3 weeks allowance for curation team and comp bios to review the proposed migration implications before the migration plan is finalized. This will allow us to survey CELLxGENE data for the deprecated genes and identify any 'surprises' in the updated annotation version.

**Current draft proposal**
Definitions
_current_ = the pinned annotation that the schema currently references
_future_ = the pinned annotation that the new schema version will reference

- Any gene ID that is present in the _current_ version and the _future_ version will be retained
- Any gene ID that is present in the _current_ version but not in the _future_ version will be evaluated and retained with a revised gene ID if:
  - the transcript IDs (ENST without version number appended) attributed to that gene match exactly to a _future_ gene ID
AND/OR
  - the seqname (chromosome), start, stop, and strand all match exactly to a _future_ gene ID
  - AND the matching _future_ gene ID must not be present in the _current_ version

[...]

PR #667, opened 2023-10-06
feat: branching schema bump workflows based on major vs minor schema version bump

nayib-jose-gloria:
## Reason for Change

- #572

## Changes

- schema bump update workflow takes an input, a flag for whether the bump is a major schema bump. Defaults to false
- If major schema bump, bump the RC version only (assumes major bump occurs during dev work + qa) 
- If minor schema bump, run minor version bump + RC bump
- If minor schema bump, release to PyPI (major bump PyPI release is done AFTER migrate script is merged, whereas minor schema bump flow is set-up to release the ontology bump validation package and then release again after the migrate script is done for migration)

## Testing

- will try triggering test runs in a branch

## Notes for Reviewer

### DIFF
--- a/.github/workflows/schema-version-bump.yml
+++ b/.github/workflows/schema-version-bump.yml
@@ -1,7 +1,12 @@
-name: Minor Schema Version Bump Workflow
+name: Schema Version Bump Workflow
 
 on:
   workflow_dispatch
+    inputs:
+      is-major-schema-bump:
+        description: "Set to 'true' to kickoff major schema update; otherwise trigger a minor schema update"
+        required: true
+        default: 'false'
 
 permissions:
   id-token: write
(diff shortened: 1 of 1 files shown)


.github/workflows/schema-version-bump.yml
```diff
@@ -239,13 +245,21 @@ jobs:
       - name: remove diff files
         run: |
           git rm --ignore-unmatch ./cellxgene_schema_cli/cellxgene_schema/ontology_files/*_diff.txt
-      - name: Bump Version
+      - name: Bump Patch Version
+        if: ${{ github.event.inputs.is-major-schema-bump }} == 'false'
```
danieljhegeman, line 249: isn't this condition already gated by the addition on line 117? 👆

danieljhegeman (approved):
LGTM I think! I assume this is configured correctly, but where does the user give the `is-major-schema-bump` set? Also it's now dawning on me that hyphens `-` actually _are_ allowed in GHA input and output var names...

nayib-jose-gloria:
> LGTM I think! I assume this is configured correctly, but where does the user give the `is-major-schema-bump` input? Also it's now dawning on me that hyphens `-` actually _are_ allowed in GHA input and output var names...

it'll appear, alongside the description, in the github UI!

nayib-jose-gloria:
Screenshot 2023-10-17 at 10 08 35 AM
@danieljhegeman for reference

PR #668, opened 2023-10-11
fix: change how we detect strings in uns[{column}_colors]

joyceyan:
## Reason for Change

- discovered through QA in [this thread](https://github.com/chanzuckerberg/single-cell-curation/issues/513) that apparently if you create a numpy array of strings and use it to generate the anndata, anndata types the numpy dtype as `object`. this isn't a problem if you create the anndata object first, and then update `adata.uns[{column}_colors]` after
- we need to update how we detect if a numpy array contains strings or not

## Changes

- instead of checking the dtype of the numpy array, we'll instead just check every element within the array and see if they are instances of python `str`
- this PR also incorporates one of the suggested changes to the english in one of the error messages
> When a corresponding obs field is present but it is not categorical, current is: ERROR: Colors field uns[{column}_colors] does not have a corresponding categorical field in obs. Would help the curator diagnose if it was appended with: {column} is present but is dtype {}

## Testing

- i created a new test fixture `adata_with_colors` that creates `uns` with colors for the categorical columns
- this test failed before i changed the logic in `validate.py`

## Notes for Reviewer


cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -750,11 +750,13 @@ def _validate_colors_in_uns_dict(self, uns_dict: dict) -> None:
...
             if key.endswith("_colors"):
                 # 1. Verify that the corresponding categorical field exists in obs
-                obs_unique_values = category_mapping.get(key.replace("_colors", ""))
+                column_name = key.replace("_colors", "")
+                obs_unique_values = category_mapping.get(column_name)
                 if not obs_unique_values:
-                    self.errors.append(
-                        f"Colors field uns[{key}] does not have a corresponding categorical field in obs"
-                    )
+                    error_message = f"Colors field uns[{key}] does not have a corresponding categorical field in obs"
```
nayib-jose-gloria, line 756: nit: add a period, for clarity in cases where the second error sentence is appended

joyceyan: updated!

nayib-jose-gloria (commented):
would it be straightforward to add a unit test that mimics Jason's use case that led to this change? i.e. confirm numpy "objects" that are strings are correctly evaluated as python str instances and therefore pass validation

nayib-jose-gloria:
> would it be straightforward to add a unit test that mimics Jason's use case that led to this change? i.e. confirm numpy "objects" that are strings are correctly evaluated as python str instances and therefore pass validation

or is that already taken care of by using the adata_with_colors fixture?

joyceyan:
yep, that's already handled with the `adata_with_colors` fixture. i wrote the test first, and it failed against the original code i had written for string validation since the colors were detected as `object` instead of `string`. i probably could have made that a bit more clear in the PR description 😅

nayib-jose-gloria (approved):

PR #671, opened 2023-10-19
feat: add mypy to the scripts/ module

joyceyan:
## Reason for Change

- https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/single-cell-curation/662

## Changes

- adds mypy config to `pyproject.toml`
- adds a make command, `make mypy`, which runs `mypy --config-file pyproject.toml`
- adds the `run_mypy_and_annotate.py` script to the root directory of the project, which runs the following command: `mypy --config-file pyproject.toml` and annotates offending lines with `# type: ignore`
- adds a pre-commit hook to run mypy

## Testing

- ran the script, and verified `make mypy` can run with no issues

## Notes for Reviewer


.github/workflows/push_tests.yml
```diff
@@ -16,8 +16,14 @@ jobs:
         with:
           ref: ${{ github.event.pull_request.head.ref }}
       - uses: actions/setup-python@v3
-      - name: check backend
+      - name: Pre-commit checks
         uses: pre-commit/action@v3.0.0
+      - name: Install dependencies
+        run: |
+          pip install -r ./requirements-dev.txt
+      - name: Run mypy
```
joyceyan, line 24: this can't go in the pre-commit check because for some reason when you invoke `mypy --config-file pyproject.toml` as part of `per-commit run --all-files` it will run mypy across all the files instead of just the package specified in `pyproject.toml`

pyproject.toml
```diff
@@ -29,3 +29,11 @@ select = ["E", "F", "B", "SIM", "C4", "I"]
 ignore = ["E501", "E402", "C408", "SIM112"]
 line-length = 120
 target-version = "py38"
+
+[tool.mypy]
+packages = ["scripts"]
```
joyceyan, line 34: here is where we tell mypy to only check the `scripts` package. note that we're using `packages = ["scripts"]` instead of `modules = ["scripts"]` because `packages` will recursive check subdirectories, while `modules` does not.

https://mypy.readthedocs.io/en/stable/config_file.html#confval-packages

requirements-dev.txt
```diff
@@ -1,2 +1,7 @@
 coverage
 pytest
+mypy
+types-requests
```
joyceyan, line 4: these typing dependencies were copied from [census](https://github.com/chanzuckerberg/cellxgene-census/blob/main/.pre-commit-config.yaml), and were all needed to get various types of errors to go away

run_mypy_and_annotate.py
```diff
@@ -0,0 +1,37 @@
+import re
+import subprocess
+
+
+def run_mypy():
```
joyceyan, line 5: here is the script where we annotate offending lines with `# type: ignore`! i think it is worth having someone review the script, though maybe we don't want to actually commit it to the repo

cellxgene_schema_cli/cellxgene_schema/ontology.py
```diff
@@ -63,7 +63,7 @@ def __init__(self, species: SupportedOrganisms):
             duplicated_gene_labels = set()
 
             for gene in genes:
-                gene = gene.rstrip().split(",")
```
joyceyan, line 66: i'm not entirely sure why this file was picked up from running mypy against the `scripts/` directory. my best guess is that there's something within the `scripts/` directory that's importing this in, though i couldn't find a direct import

.github/workflows/push_tests.yml
atolopko-czi: So once we expand the mypy coverage to `cellxgene_schema_cli` directory, do you think we'll still run into this problem? It seems important to resolve, since running `pre-commit` is how we'll all expect to run mypy. If nothing else, move your comment into the yml source. If we can't figure out, make an issue for the future

Makefile
```diff
@@ -11,6 +11,9 @@ install:
 clean:
 	rm -rf cellxgene_schema_cli/build cellxgene_schema_cli/dist cellxgene_schema_cli/cellxgene_schema.egg-info
 
+mypy:
+	mypy --config-file pyproject.toml
+
```
atolopko-czi, line 16: seems like fmt/lint/mypy could all just be folded together into something like `check` or even `pre-commit`. Out-of-scope fort the PR, perhaps, but can't flake8 also be run by pre-commit?

cellxgene_schema_cli/cellxgene_schema/ontology.py
atolopko-czi: Your theory re:import dependency sound right. Not a big deal either way. If you feed explicit file names to mypy instead of packages or modules, it may not do the transitive checking (https://mypy.readthedocs.io/en/stable/running_mypy.html#reading-a-list-of-files-from-a-file). But not sure if mypy will be happy if this file does _not_ have correct or ignored typing hints.

run_mypy_and_annotate.py
```diff
@@ -0,0 +1,37 @@
...
+                        if "# type: ignore" not in line:
+                            comment_index = line.find("#")
+                            if comment_index >= 0:
+                                line_parts = [line[:comment_index], line[comment_index:]]
+                            else:
+                                line_parts = [line, ""]
+                            updated_line = f"{line_parts[0]}  # type: ignore{line_parts[1]}\n"
+                            lines[line_number - 1] = updated_line
+                            file.seek(0)
+                            file.writelines(lines)
```
atolopko-czi, line 32: nit: might want to break this down into a few functions, for readability. it's a one-time'ish script, so not a big deal!

atolopko-czi (changes requested):
Some comments for your review before merging, but overall looks reasonable! 

Also, @Bento007 might be another worthwhile reviewer on this.

.github/workflows/push_tests.yml
joyceyan: Yes, I believe we'll still run into the problem unless we update this to run across all files in the repo. My guess is that the `--all-files` argument in `pre-commit` overrides the more narrow scanning we specify in `pyproject.toml`, though I can't find any documentation in the pre-commit project actually supporting that. I'll add a comment to both this yaml file and the pre-commit yaml file though, that's a good idea!

Makefile
joyceyan: yeah that seems reasonable, i can consolidate

run_mypy_and_annotate.py
joyceyan: oh cool, i didn't know that repo existed. definitely seems like this belongs there!

.github/workflows/push_tests.yml
```diff
@@ -26,7 +26,7 @@ jobs:
       # the more narrow specifying of files in pyproject.toml
```
atolopko-czi, line 26:
```suggestion
      # the more narrow specifying of files in pyproject.toml
      # TODO: Remove this if/when all files are mypy-compliant
```

atolopko-czi (approved):
LGTM! Just one comment suggestion.

.github/workflows/push_tests.yml
joyceyan: good suggestion - added :)

PR #688, opened 2023-11-20
chore: remove mypy

danieljhegeman:
## Reason for Change

- We should keep the mypy recipe in the makefile to help people make use of mypy where they wish to. It can be a useful guide. It should not be a check. Or, if it is a check, it should be its own check so its failure does not mask the success/failure of any other checks.

## Changes

- Do not run mypy as a GHA job step.

## Testing

version at HEAD of main with 
```
```

## Notes for Reviewer

### DIFF
--- a/.github/workflows/push_tests.yml
+++ b/.github/workflows/push_tests.yml
@@ -18,16 +18,6 @@ jobs:
       - uses: actions/setup-python@v3
       - name: Pre-commit checks
         uses: pre-commit/action@v3.0.0
-      - name: Install dependencies
-        run: |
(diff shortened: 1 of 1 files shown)


atolopko-czi (commented):
Unenforced mypy would be useless, no? Any attempt to use it "on demand" would produce a list of all the violations that accumulated. I agree it should be it’s own GHA check, so it’s obvious why the workflow failed.

joyceyan (approved):

joyceyan (changes requested):
unapproving this while we discuss how we want to approach this problem

PR #692, opened 2023-11-21
fix: add necessary corpus transformations uncovered by dev migration

nayib-jose-gloria:
## Changes

- added ontology term replacements corpus-wide for 3 terms we are now denying in schema 4 but are currently found in the corpus, per discussion in single-cell-four with Jason
- added check for datasets with RNA assays to meet new schema requirements--if raw matrix dtype is not np.float32 and the matrix values can be coerced safely into np.float32, do so and report the change as an output of the migration script (the migration SFM uses the output to report to slack in the curator report, mapping to the dataset_id and collection_id)
      - Because the validator already enforces that raw matrix values for RNA assays are positive integers, we don't have to worry about floating point precision being lost. The only concern for safe casting is whether any values are too large to be represented as np.float32; so we check the max value of the matrix against the np.float32 max.

## Testing

-  Local testing against prod datasets that failed in dev migration--validation succeeds after running with this updated script

## Note for Reviewer

- Writing to the raw matrix requires reading the dataset into memory, which we were not doing before. Our machines for the migration should be large enough to handle our worst case, but it is an added risk for this run. Another Dev migration run would catch such issues

### DIFF
--- a/cellxgene_schema_cli/cellxgene_schema/migrate.py
+++ b/cellxgene_schema_cli/cellxgene_schema/migrate.py
@@ -6,6 +6,7 @@ import numpy as np
 import pandas as pd
 
 from . import utils
+from . import ontology
 
 # fmt: off
 # ONTOLOGY TERMS TO UPDATE ACROSS ALL DATASETS IN CORPUS
@@ -20,6 +21,8 @@ ONTOLOGY_TERM_MAPS = {
     "assay": {
     },
     "cell_type": {
+        "CL:0000255": "CL:0000003",
+        "CL:0000548": "CL:0000003",
     },
     "development_stage": {
     },
@@ -28,6 +31,7 @@ ONTOLOGY_TERM_MAPS = {
     "organism": {
     },
     "self_reported_ethnicity": {
+        "HANCESTRO:0306": "unknown"
     },
     "sex": {
     },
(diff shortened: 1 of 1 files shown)


cellxgene_schema_cli/cellxgene_schema/migrate.py
```diff
@@ -192,6 +194,43 @@ def color_key_is_valid(key: str, value: str) -> bool:
...
+        if bool(accessibility_assays & term_ancestors) is False:
+            has_rna_assay = True
+            break
+
+    if has_rna_assay:
+        dataset = dataset.to_memory()
+        if dataset.raw:
+            raw = dataset.raw.X
+        else:
+            raw = dataset.X
```
Bento007, line 223: does this do a deep copy?

Bento007 (commented):
Why isn't this check already done as part of validation and write_labels?

nayib-jose-gloria:
> Why isn't this check already done as part of validation and write_labels?

it is validated against, hence the failures we're seeing in prod

It is a new check added for schema 4 (product requirement), so the prod corpus has several datasets that previously passed (correctly, since it wasn't a hard requirement in schema 3 and below) that must now be migrated to pass

cellxgene_schema_cli/cellxgene_schema/migrate.py
nayib-jose-gloria: Nope, its a view

nayib-jose-gloria: I verified locally that by updating this reference var, it updates the dataset

Bento007 (approved):

PR #709, opened 2023-12-13
feat: validate obsm embeddings start with letter

joyceyan:
## Reason for Change

https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/single-cell-data-portal/6340

## Changes

- raises an error if either the obsm embedding key starts with a non-letter, or if it begins with X_ and the suffix starts with a non-letter

## Testing

- added some unit tests to verify that this works as expected

## Notes for Reviewer

### DIFF
--- a/cellxgene_schema_cli/cellxgene_schema/validate.py
+++ b/cellxgene_schema_cli/cellxgene_schema/validate.py
@@ -921,30 +921,23 @@ class Validator:
         for key, value in self.adata.obsm.items():
             issue_list = self.errors
 
-            if not key[0].isalpha():
-                self.errors.append(f"Embedding key in 'adata.obsm' {key} must start with a letter.")
+            regex_pattern = r'^[a-zA-Z][a-zA-Z0-9]*$'
 
             if key.startswith("X_"):
                 obsm_with_x_prefix += 1
-                if len(key) <= 3:
+                if not re.match(regex_pattern, key[2:]):
                     self.errors.append(
-                        f"Embedding key in 'adata.obsm' {key} must start with X_ and have a suffix at least one character long."
+                        f"Suffix for embedding key in 'adata.obsm' {key} does not match the regex pattern {regex_pattern}."
                     )
-                else:
-                    suffix_first_char = key[2]
-                    if not suffix_first_char.isalpha():
-                        self.errors.append(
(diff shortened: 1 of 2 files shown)


cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -920,12 +920,22 @@ def _validate_obsm(self):
         obsm_with_x_prefix = 0
         for key, value in self.adata.obsm.items():
             issue_list = self.errors
+
+            if not key[0].isalpha():
```
nayib-jose-gloria, line 924: The actual Seurat v5 regex is matching against `^[a-zA-Z][a-zA-Z0-9]*$` for keys (or suffixes of keys that begin with `X_`). Which also encompasses making sure the key is alphanumeric overall. If that isn't already captured elsewhere, we might want to just check against this regex.

brianraymor: That was not captured in the original Data Viz requirements which only prohibited white space characters. 

I was writing an update for Nik this morning:

- {suffix} MUST be at least one character in length.
- The first character of {suffix} MUST be a letter of the alphabet.
- {suffix} MUST NOT contain whitespace characters.

I've been trying hard to avoid ABNF in the schema, but perhaps it's time to consider using RegEx expressions for clarity; otherwise:

- {suffix} MUST be at least one character in length.
- The first character of {suffix} MUST be a letter of the alphabet.
- The remaining characters of {suffix} MUST be alphanumeric characters.

brianraymor: It does seem simpler to just write:

{suffix} MUST match the regular expression pattern "^[a-zA-Z][a-zA-Z0-9]*$".

joyceyan: @nayib-jose-gloria Good point - I'll just have the code use the regex pattern.

@brianraymor For clarity including the regular expression pattern is definitely useful. But I think it would be better to include the regex in addition to the English, rather than replacing it, since most people can't read regex and immediately understand what it means.

nayib-jose-gloria (approved):

PR #748, opened 2024-01-31
feat: Validate that uns array values are all > 0 in length

joyceyan:
## Reason for Change

https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/single-cell-curation/728

## Changes

- Validates that uns values cannot be empty arrays, dictionaries, or strings

## Testing

- Wrote unit tests to make sure error is raised for various types of zero-length values

## Notes for Reviewer

### DIFF
--- a/cellxgene_schema_cli/cellxgene_schema/validate.py
+++ b/cellxgene_schema_cli/cellxgene_schema/validate.py
@@ -746,7 +746,7 @@ class Validator:
                         self.warnings.append(column_def["warning_message"])
                     self._validate_column(column, column_name, df_name, column_def)
 
-    def _validate_colors_in_uns_dict(self, uns_dict: dict) -> None:
+    def _validate_uns_dict(self, uns_dict: dict) -> None:
         df = getattr_anndata(self.adata, "obs")
 
         # Mapping from obs column name to number of unique categorical values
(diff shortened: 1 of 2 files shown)


cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -759,6 +759,8 @@ def _validate_colors_in_uns_dict(self, uns_dict: dict) -> None:
                 category_mapping[column_name] = column.nunique()
 
         for key, value in uns_dict.items():
+            if len(value) == 0:
+                self.errors.append(f"uns['{key}'] cannot be an empty value.")
```
danieljhegeman, line 763: I think this should be moved down to be integrated with the `if len(value) < obs_unique_values` statement on ln 801, because this is a higher-order error, and so it should strictly supercede the error condition on 801. Integrate as an `if len(value) == 0... elif len(value) < obs_unique_values`

danieljhegeman: Although admittedly that is not a colors-related check... so it would be strange to have it as part of the colors logic loop.

Maybe instead change ln 801 to be `if len(value) > 0 and len(value) < obs_unique_values`

Add a comment to indicate the zero condition is checked independently earlier in the function.

danieljhegeman: Yeah we have to do it that way because we need to check all values for zero length, not just the ones that have a `_colors` key, so disregard my first comment in this thread

danieljhegeman (changes requested):
One request

danieljhegeman (approved):

Issue #735, opened 2024-01-24
cellxgene-schema CLI must update validation for cell_type_ontology_term_id and annotation for cell_type


brianraymor:
# Context

See [Update now-obsolete 'native cell' in schema](https://github.com/chanzuckerberg/single-cell-curation/issues/622).

---

# Changelog

  * Updated the requirements for `cell_type` to annotate`"unknown"` as the label when the `cell_type_ontology_term_id`value  is  `"unknown"`. 
  * Updated the requirements for `cell_type_ontology_term_id` to replace `"CL:0000003"` for *native cell* with `"unknown"` to indicate that the cell type is unknown. 

---

# Schema

See [cell_type_ontology_term_id](https://github.com/chanzuckerberg/single-cell-curation/blob/main/schema/4.1.0/schema.md#cell_type_ontology_term_id)
See [cell_type](https://github.com/chanzuckerberg/single-cell-curation/blob/main/schema/4.1.0/schema.md#cell_type)

danieljhegeman:
This validation change should also be applied to `tissue_ontology_term_id` per conversation with @brianraymor 

Brian, do you plan on making analogous updates to the schema text for `tissue_ontology_term_id`?

brianraymor:
Per my response on the call this morning, I felt this was already addressed by the requirement in `tissue_ontology_term_id`:

> If tissue_type is "cell culture" this MUST follow the requirements for cell_type_ontology_term_id.

danieljhegeman:
Acknowledged. I had missed the connecting `tissue_type` designator.

PR #750, opened 2024-02-01
feat: deprecate CL:0000003 'native cell' and replace with "unknown"

danieljhegeman:
- add exception for "unknown" to schema def yml
- update test_convert_template.py
- migrate script should get auto-updated during migration workflow

## Reason for Change

- #735 

## Changes

- Permit `"unknown"` for `cell_type_ontology_term_id`
  - Annotate `cell_type` as `"unknown"`
- Permit `"unknown"` for `tissue_ontology_term_id` when `tissue_type` is `"cell culture"`
  - Annotate `tissue` as `"unknown"`
- Various unrelated typo fixes
- Fix test `unmigrated_X` in `examples_validate.py` to reference correct obs df shape as source of row count when instantiating

## Testing

- unit tests added to test against scenarios where the ontology term for cell type or tissue is `"unknown"`
- test label writes
```
```

## Notes for Reviewer

### DIFF
--- a/cellxgene_schema_cli/cellxgene_schema/validate.py
+++ b/cellxgene_schema_cli/cellxgene_schema/validate.py
@@ -464,7 +464,7 @@ class Validator:
         self, df: pd.DataFrame, df_name: str, column_name: str, dependencies: List[dict]
     ) -> pd.Series:
         """
-        Validates subset of columns based on dependecies, for instance development_stage_ontology_term_id has
+        Validates subset of columns based on dependencies, for instance development_stage_ontology_term_id has
         dependencies with organism_ontology_term_id -- the allowed values depend on whether organism is human, mouse
         or something else.
 
@@ -474,7 +474,7 @@ class Validator:
         :param pd.DataFrame df: pandas dataframe containing the column to be validated
         :param str df_name: the name of dataframe in the adata object, e.g. "obs"
         :param str column_name: the name of the column to be validated
-        :param list dependencies: a list of dependecy definitions, which is a list of column definitions with a "rule"
+        :param list dependencies: a list of dependency definitions, which is a list of column definitions with a "rule"
         """
 
         all_rules = []
--- a/cellxgene_schema_cli/cellxgene_schema/validate.py
+++ b/cellxgene_schema_cli/cellxgene_schema/validate.py
@@ -1255,9 +1255,9 @@ class Validator:
 
             # Do it for columns that map to other columns, for post-upload annotation
             if "columns" in component_def:
-                for _column, columns_def in component_def["columns"].items():
-                    if "add_labels" in columns_def:
-                        self._check_single_column_availability(component, columns_def["add_labels"])
+                for column_def in component_def["columns"].values():
+                    if "add_labels" in column_def:
+                        self._check_single_column_availability(component, column_def["add_labels"])
 
             # Do it for index that map to columns
             if "index" in component_def:
--- a/cellxgene_schema_cli/tests/fixtures/examples_validate.py
+++ b/cellxgene_schema_cli/tests/fixtures/examples_validate.py
@@ -250,7 +250,7 @@ adata_with_colors = anndata.AnnData(
 )
 
 # anndata for testing migration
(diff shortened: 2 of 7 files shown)


cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -1255,9 +1255,9 @@ def _check_column_availability(self):
 
             # Do it for columns that map to other columns, for post-upload annotation
             if "columns" in component_def:
-                for _column, columns_def in component_def["columns"].items():
-                    if "add_labels" in columns_def:
-                        self._check_single_column_availability(component, columns_def["add_labels"])
+                for column_def in component_def["columns"].values():
+                    if "add_labels" in column_def:
+                        self._check_single_column_availability(component, column_def["add_labels"])
```
danieljhegeman, line 1260: unrelated code refactor

cellxgene_schema_cli/tests/fixtures/examples_validate.py
```diff
@@ -297,12 +297,12 @@
     ],
 )
 
-unmigrated_X = numpy.zeros([good_obs.shape[0], var_unmigrated.shape[0]], dtype=numpy.float32)
-adata_with_lables_unmigrated = anndata.AnnData(
+unmigrated_X = numpy.zeros([unmigrated_obs.shape[0], var_unmigrated.shape[0]], dtype=numpy.float32)
```
danieljhegeman, line 300: I believe this was what you intended @Bento007 (not the typo fix from `umigrated_obs` => `unmigrated_obs`, but the reference to the unmigrated obs df rather than the "good" obs df). But everything was working by chance since each df had 2 rows.

Confirm?

cellxgene_schema_cli/tests/test_schema_compliance.py
```diff
@@ -576,6 +576,48 @@ def test_self_reported_ethnicity_ontology_term_id__unknown(self, validator_with_
...
 
+    def test_cell_type_ontology_term_id__unknown(self, validator_with_adata):
+        """
+        Test 'unknown' cell_type_ontology_term_id is valid
+        """
+        validator = validator_with_adata
+        obs = validator.adata.obs
+        obs.at["Y", "cell_type_ontology_term_id"] = "unknown"
+        assert validator.validate_adata()
+        assert validator.errors == []
```
joyceyan, line 587: i think this line is redundant since `validator.validate_adata()` should only return True if there's no errors

cellxgene_schema_cli/tests/test_schema_compliance.py
```diff
@@ -576,6 +576,48 @@ def test_self_reported_ethnicity_ontology_term_id__unknown(self, validator_with_
...
+        # Arrange -- relies on "tissue_type" value for index "Y" being "cell culture", set explicitly
+        obs.at["Y", "tissue_type"] = "cell culture"
+        obs.at["Y", "tissue_ontology_term_id"] = "unknown"
+
+        assert validator.validate_adata()
+        assert validator.errors == []
+
+    def test_tissue_ontology_term_id__unknown_invalid(self, validator_with_adata):
+        """
+        Test 'unknown' tissue_ontology_term_id is valid if tissue_type is 'cell culture'
```
joyceyan, line 605: i'm guessing this comment was copied and pasted, but can you update this to reflect what the test actually does? it would also be fine to remove this entirely, since the comment on line 610 explains the test sufficiently imo

joyceyan (changes requested):

cellxgene_schema_cli/tests/test_schema_compliance.py
danieljhegeman: thank you 🙏 😅

cellxgene_schema_cli/tests/test_schema_compliance.py
danieljhegeman: Fair, but prefer to leave it explicit esp since this is the pattern used by other tests

joyceyan (approved):

PR #753, opened 2024-02-02
feat: allow none / true / false values

joyceyan:
## Reason for Change

follow up to https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/single-cell-curation/728

## Changes

allows none, true, and false values for `uns` fields

## Testing

added unit tests

## Notes for Reviewer


danieljhegeman (approved):
One suggestion but totally up to you

PR #803, opened 2024-03-08
fix: bool check for uns empty keys

joyceyan:
## Reason for Change

we have datasets with `adata.uns["log1p"] = numpy.bool_(True)`, so our check for `type(value) is not bool` doesn't work. for some reason, i can't directly connect this PR to this issue: https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/single-cell-curation/728 but it does patch our previous work on that ticket, in addition to this one: https://github.com/chanzuckerberg/single-cell/issues/684

## Changes

updates to also verify that `type(value)` is not `np.bool_` in both the migrate script and the schema validation

## Testing

added unit test coverage

## Notes for Reviewer

### DIFF
--- a/cellxgene_schema_cli/cellxgene_schema/migrate.py
+++ b/cellxgene_schema_cli/cellxgene_schema/migrate.py
@@ -1,4 +1,5 @@
 import anndata as ad
+import numpy as np
 
 from . import utils
 
@@ -366,7 +367,7 @@ def migrate(input_file, output_file, collection_id, dataset_id):
     # Delete any uns keys with an empty value, logic taken from:
     # https://github.com/chanzuckerberg/single-cell-curation/blob/43f891005fb9439dbbb747fa0df8f0435ebf3f7c/cellxgene_schema_cli/cellxgene_schema/validate.py#L761-L762
     for key, value in list(dataset.uns.items()):
-        if value is not None and type(value) is not bool and len(value) == 0:
+        if value is not None and type(value) is not bool and type(value) is not np.bool_ and len(value) == 0:
             del dataset.uns[key]
 
     if "X_.umap_MinDist_0.2_N_Neighbors_15" in dataset.obsm:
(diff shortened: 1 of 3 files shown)


cellxgene_schema_cli/cellxgene_schema/migrate.py
```diff
@@ -366,7 +367,7 @@ def migrate(input_file, output_file, collection_id, dataset_id):
     # Delete any uns keys with an empty value, logic taken from:
     # https://github.com/chanzuckerberg/single-cell-curation/blob/43f891005fb9439dbbb747fa0df8f0435ebf3f7c/cellxgene_schema_cli/cellxgene_schema/validate.py#L761-L762
     for key, value in list(dataset.uns.items()):
-        if value is not None and type(value) is not bool and len(value) == 0:
+        if value is not None and type(value) is not bool and type(value) is not np.bool_ and len(value) == 0:
```
danieljhegeman, line 370:
```suggestion
        if value is not None and not (value == False or value == True) and len(value) == 0:
```

danieljhegeman: This needs to be generalized to catch all boolean-like values, because there are actually other flavors of boolean in the numpy suite (and who knows what else is out there that could show up later).

cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -759,7 +759,7 @@ def _validate_uns_dict(self, uns_dict: dict) -> None:
                 category_mapping[column_name] = column.nunique()
 
         for key, value in uns_dict.items():
-            if value is not None and type(value) is not bool and len(value) == 0:
+            if value is not None and type(value) is not bool and type(value) is not np.bool_ and len(value) == 0:
```
danieljhegeman, line 762:
```suggestion
            if value is not None and not (value == False or value == True) and len(value) == 0:
```

danieljhegeman: same as above

danieljhegeman (changes requested):

cellxgene_schema_cli/cellxgene_schema/migrate.py
danieljhegeman, line 370:
```suggestion
        if value is not None and type(value) is not bool and not (isinstance(value, np.bool_) or isinstance(value, np.bool)) and len(value) == 0:
```

danieljhegeman (approved):

Issue #674, opened 2023-10-20
Add requirements for Visium Spatial Gene Expression assay


brianraymor:
# Context

This addresses the [census requirements](https://docs.google.com/document/d/1x5jqTvI7xb-r2cqSS91FrriE0oVZvGf5FzCFDGx7Fc4/edit#heading=h.sb44irnz9vkx) authored by @pablo-gar. 

See [Improve CELLxGENE’s value proposition for data submitters and consumers by supporting visium and slide-seq experiments](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/single-cell/473) and [Data Platform changes required to support visium experiments](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/single-cell/512)

# Design

## Pending

[[NTR] Version specific Visium assays](https://github.com/EBISPOT/efo/issues/2178)

For easier review, these requirements are **additive** to the corresponding sections in the [schema 4 draft](https://github.com/chanzuckerberg/single-cell-curation/blob/main/schema/4.0.0/schema.md). 

## General Requirements

...

<b>Visium Spatial Gene Expression.</b> It is STRONGLY RECOMMENDED that <i>Visium Spatial Gene Expression</i> datasets represent one Space Ranger output for a single tissue section. This representation is referenced throughout the schema as <b>Visium Single</b>.

Visium datasets that represent multiple Space Ranger outputs MAY be submitted. This representation is referenced throughout the schema as <b>Visium Multiple</b> which will have limited support in CELLxGENE experiences:
* There are no image underlays in CELLxGENE Explorer.
* Such datasets are not included in CELLxGENE Discover Census.
* Such datasets are not converted to Seurat for CELLxGENE Discover downloads.

---

## `X` (Matrix Layers)

...

The following table describes the matrix data and layers requirements that are **assay-specific**. If an entry in the table is empty, the schema does not have any other requirements on data in those layers beyond the ones listed above.

| Assay | "raw" required? | "raw" location | "normalized" required? | "normalized" location |
|-|-|-|-|-|
[...]

PR #823, opened 2024-03-22
updated X requirements for visium

brianraymor:


schema/5.1.0/schema.md
```diff
@@ -183,7 +183,8 @@ The following table describes the matrix data and layers requirements that are *
...
+|||||
```
jahilton, line 187: Why does this have additional specifications about sparse.csr_matrix & implicit zeros
(as opposed to other data which are just `In any layer, if a matrix has 50% or more values that are zeros, it is STRONGLY RECOMMENDED that the matrix be encoded as a scipy.sparse.csr_matrix`)?

brianraymor: I was responding to Pablo's recommendation (which he [previously approved](https://github.com/chanzuckerberg/single-cell-curation/issues/674#issuecomment-2000514827) in its earlier iteration. 

Pablo Garcia-Nieto
(https://czi-sci.slack.com/archives/C02TSE92DBM/p1707268059182289?thread_ts=1707240003.577709&cid=C02TSE92DBM)
Ah great to hear! Census is not going to ingest normalized data, so no strong opinions on that. For all other uses I believe it’s ok to add these dummy 0s, but then I would strongly recommend to store these matrices in scipy sparse format and to encode the 0's implicitly. 

 **this should be enforced by the validator.**

pablo-gar: I think the original text pointed out by Jason with slight modification captures my recommendation:

> In any layer, if a matrix has 50% or more values that are zeros, it is STRONGLY RECOMMENDED that the matrix be encoded as a scipy.sparse.csr_matrix. **If the matrix is encoded as a scipy.sparse.csr_matrix, then zero values MUST be encoded as <a href="https://docs.scipy.org/doc/scipy/tutorial/sparse.html#sparse-arrays-implicit-zeros-and-duplicates">implicit zeros</a>.**

And then you can remove the following

> <br>This MUST be a <a href="https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.csr_matrix.html"><code>scipy.sparse.csr_matrix</code></a>. If the <code>obs['in_tissue']</code> is <code>0</code> for a cell, then values MUST be encoded as <a href="https://docs.scipy.org/doc/scipy/tutorial/sparse.html#sparse-arrays-implicit-zeros-and-duplicates">implicit zero</a>.

Issue #827, opened 2024-03-27
cellxgene-schema CLI must add validation for uns['spatial']


brianraymor:
# Design

See [`uns['spatial']](https://github.com/chanzuckerberg/single-cell-curation/blob/main/schema/5.1.0/schema.md#spatial-1).

### spatial

<table><tbody>
    <tr>
      <th>Key</th>
      <td>spatial</td>
    </tr>
    <tr>
      <th>Annotator</th>
      <td>Curator MUST annotate if <code>assay_ontology_term_id</code> is <a href="https://www.ebi.ac.uk/ols4/ontologies/efo/classes?obo_id=EFO%3A0010961"><code>"EFO:0010961"</code></a> for <i>Visium Spatial Gene Expression</i> or <a href="https://www.ebi.ac.uk/ols4/ontologies/efo/classes?obo_id=EFO%3A0030062"><code>"EFO:0030062"</code></a> for <i>Slide-seqV2</i>; otherwise, this key MUST NOT be present.</td>
    </tr>
    <tr>
      <th>Value</th>
        <td><code>dict</code>. The requirements for the key-value pairs are documented in the following sections:
          <ul>
          <li>spatial['is_single']</li>         
          <li>spatial[<i>library_id</i>]</li>
          <li>spatial[<i>library_id</i>]['images']</li>
          <li>spatial[<i>library_id</i>]['images']['fullres']</li>
          <li>spatial[<i>library_id</i>]['images']['hires']</li>
          <li>spatial[<i>library_id</i>]['scalefactors']</li>
          <li>spatial[<i>library_id</i>]['scalefactors']['spot_diameter_fullres']</li>
          <li>spatial[<i>library_id</i>]['scalefactors']['tissue_hires_scalef']</li>
         </ul><br>Additional key-value pairs MUST NOT be present.
        </td>
    </tr>
</tbody></table>
<br>

#### is_single

<table><tbody>
    <tr>
      <th>Key</th>
      <td>is_single</td>
    </tr>
    <tr>
      <th>Annotator</th>
[...]

PR #858, opened 2024-04-24
feat(schema 5.1.0): validate uns[spatial]

MillenniumFalconMechanic:
## Reason for Change

- #827 

## Changes

- Added `_check_spatial()` to validate `uns["spatial"]` values.
- Added Seurat conversion warning for Visium datasets.

## Testing

- Added unit tests.
- Tested with 5.1.0 datasets.

### DIFF
--- a/cellxgene_schema_cli/cellxgene_schema/validate.py
+++ b/cellxgene_schema_cli/cellxgene_schema/validate.py
@@ -1336,7 +1336,7 @@ class Validator:
         is_supported_spatial_assay = self._is_supported_spatial_assay()
         if uns_spatial_specified and not is_supported_spatial_assay:
             self.errors.append(
-                "uns['spatial'] is only allowed for adata.obs['assay_ontology_term_id'] values "
+                "uns['spatial'] is only allowed for obs['assay_ontology_term_id'] values "
                 "'EFO:0010961' (Visium Spatial Gene Expression) and 'EFO:0030062' (Slide-seqV2)."
             )
             return
@@ -1348,7 +1348,7 @@ class Validator:
         # spatial is required for supported spatial assays.
         if not uns_spatial_specified:
             self.errors.append(
-                "uns['spatial'] is required for adata.obs['assay_ontology_term_id'] values "
+                "uns['spatial'] is required for obs['assay_ontology_term_id'] values "
                 "'EFO:0010961' (Visium Spatial Gene Expression) and 'EFO:0030062' (Slide-seqV2)."
             )
             return
@@ -1380,7 +1380,7 @@ class Validator:
         is_visium_and_uns_is_single = is_visium and uns_is_single
         if len(library_ids) > 0 and not is_visium_and_uns_is_single:
             self.errors.append(
-                "uns['spatial'][library_id] is only allowed for adata.obs['assay_ontology_term_id'] "
+                "uns['spatial'][library_id] is only allowed for obs['assay_ontology_term_id'] "
                 "'EFO:0010961' (Visium Spatial Gene Expression) and uns['spatial']['is_single'] is True."
             )
             # Exit as library_id is not allowed.
@@ -1393,7 +1393,7 @@ class Validator:
         # library_id is required if assay is Visium and is_single is True.
         if len(library_ids) == 0:
             self.errors.append(
-                "uns['spatial'] must contain the key 'library_id' for adata.obs['assay_ontology_term_id'] "
+                "uns['spatial'] must contain the key 'library_id' for obs['assay_ontology_term_id'] "
                 "'EFO:0010961' (Visium Spatial Gene Expression) and uns['spatial']['is_single'] is True."
             )
             # Exit as library_id is missing.
@@ -1507,9 +1507,7 @@ class Validator:
         :rtype bool
         """
         assay_ontology_term_id = self.adata.obs.get("assay_ontology_term_id")
-        return (
-            assay_ontology_term_id is not None and (assay_ontology_term_id == ASSAY_VISIUM).any()
-        )
+        return assay_ontology_term_id is not None and (assay_ontology_term_id == ASSAY_VISIUM).any()
 
     def _validate_float(self, name: str, value: float):
         """
--- a/cellxgene_schema_cli/cellxgene_schema/validate.py
+++ b/cellxgene_schema_cli/cellxgene_schema/validate.py
@@ -1362,7 +1362,7 @@ class Validator:
 
         # is_single must be a boolean.
         uns_is_single = uns_spatial["is_single"]
-        if not isinstance(uns_is_single, (np.bool_, np.bool)):
+        if not isinstance(uns_is_single, (np.bool_, np.bool, bool)):
             self.errors.append(f"uns['spatial']['is_single'] must be of boolean type, it is {type(uns_is_single)}.")
             # Exit if is_single is not valid as all further checks are dependent on its value.
             return
@@ -1433,7 +1433,13 @@ class Validator:
 
             # fullres is optional.
             uns_fullres = uns_images.get("fullres")
-            if uns_fullres is not None:
+            if uns_fullres is None:
+                # Warn if no fullres is specified as it is strongly recommended.
+                self.warnings.append(
+                    "No uns['spatial'][library_id]['images']['fullres'] was found. "
+                    "It is STRONGLY RECOMMENDED that uns['spatial'][library_id]['images']['fullres'] is provided."
+                )
+            else:
                 self._validate_spatial_image_shape("fullres", uns_fullres)
 
         # scalefactors is required.
--- a/cellxgene_schema_cli/cellxgene_schema/validate.py
+++ b/cellxgene_schema_cli/cellxgene_schema/validate.py
@@ -1341,29 +1341,30 @@ class Validator:
             )
             return
 
+        # Exit if we aren't dealing with a supported spatial assay as no further checks are necessary.
+        if not is_supported_spatial_assay:
+            return
+
         # spatial is required for supported spatial assays.
-        if not uns_spatial_specified and is_supported_spatial_assay:
+        if not uns_spatial_specified:
             self.errors.append(
                 "uns['spatial'] is required for adata.obs['assay_ontology_term_id'] values "
                 "'EFO:0010961' (Visium Spatial Gene Expression) and 'EFO:0030062' (Slide-seqV2)."
             )
             return
 
-        # Exit if we aren't dealing with a supported spatial assay as no further checks are necessary.
-        if not is_supported_spatial_assay:
-            return
-
         # is_single is required.
         uns_spatial = self.adata.uns["spatial"]
         if "is_single" not in uns_spatial:
             self.errors.append("uns['spatial'] must contain the key 'is_single'.")
-            # Exit if is_single is not specified as all further checks are dependent on its value.
+            # Exit if is_single is missing as all further checks are dependent on its value.
             return
 
         # is_single must be a boolean.
         uns_is_single = uns_spatial["is_single"]
         if not isinstance(uns_is_single, np.bool_):
             self.errors.append(f"uns['spatial']['is_single'] must be of boolean type, it is {type(uns_is_single)}.")
+            # Exit if is_single is not valid as all further checks are dependent on its value.
             return
 
         # Check there is at most one library_id.
@@ -1371,6 +1372,7 @@ class Validator:
         library_ids = list(filter(lambda x: x != "is_single", uns_spatial_keys))
         if len(library_ids) > 1:
             self.errors.append("uns['spatial'] must contain only one library_id.")
+            # Exit if there is more than one library_id as we don't know which library_id to validate.
             return
 
         # library_id is forbidden if assay is not Visium or is_single is false.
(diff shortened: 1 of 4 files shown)


cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -923,6 +926,13 @@ def _validate_seurat_convertibility(self):
             )
             self.is_seurat_convertible = False
 
+        # Seurat conversion is not supported for Visium datasets.
```
nayib-jose-gloria, line 929: move this block as the first check in this function, and return if `self._is_visium`. That will allow us to skip checking the X matrix, which is a more expensive operation, if we don't need to because seurat convertability is already False due to the visium check.

cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -1282,6 +1292,251 @@ def _check_column_availability(self):
...
+            return
+
+        # uns spatial validation is dependent on obs.assay_ontology_term_id; exit if not specified. Error is
+        # reported in core validate functionality.
+        obs_component = getattr_anndata(self.adata, "obs")
+        if obs_component is None or "assay_ontology_term_id" not in obs_component:
+            return
+
+        # spatial is forbidden if assay it not a supported spatial assay.
+        uns_spatial_specified = "spatial" in self.adata.uns
```
nayib-jose-gloria, line 1315: nit: if you can replace uns_spatial_specified with `uns_spatial = self.adata.uns.get("spatial")` since it returns None if uns_spatial is not specified. one less var to maintain

cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -1282,6 +1292,251 @@ def _check_column_availability(self):
...
+            return
+
+        # Exit if we're not dealing with Visium and _is_single True as no further checks are necessary.
+        if not is_visium_and_uns_is_single:
+            return
+
+        # library_id is required if assay is Visium and is_single is True.
+        if len(library_ids) == 0:
+            self.errors.append(
+                "uns['spatial'] must contain the key 'library_id' for obs['assay_ontology_term_id'] "
```
nayib-jose-gloria, line 1376: nit: since the key can be anything, we should reword this so as not to imply it must contain the literal string "library_id" as the key. 

Maybe something like: `...must contain at least one key representing the library_id`

cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -1282,6 +1292,251 @@ def _check_column_availability(self):
...
+                "'EFO:0010961' (Visium Spatial Gene Expression) and uns['spatial']['is_single'] is True."
+            )
+            # Exit as library_id is missing.
+            return
+
+        # Confirm shape of library_id is valid: allowed keys are images and scalefactors.
+        library_id_key = library_ids[0]
+        uns_library_id = uns_spatial[library_id_key]
+        if not self._has_no_extra_keys(uns_library_id, ["images", "scalefactors"]):
+            self.errors.append("uns['spatial'][library_id] can only contain the keys 'images' and 'scalefactors'.")
```
nayib-jose-gloria, line 1386: nit: now that we have validated there is only one library_id, we can substitute in the key we're referencing (here and in the subsequent error messages below)

i.e. `f"uns['spatial']['{uns_library_id}'] can only contain the keys 'images' and 'scalefactors'.""`

this more of a nice-to-have, but pointing it out since there are a couple of other requested changes

cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -1282,6 +1292,251 @@ def _check_column_availability(self):
...
+
+        # Confirm shape of image is valid: allowed shape is (,,3).
+        if not self._is_valid_visium_image_shape(image):
+            self.errors.append(
+                f"uns['spatial'][library_id]['images']['{image_name}'] must have shape (,,3), "
+                f"it has shape {image.shape}."
+            )
+
+        # Confirm max dimension of image, if specified, is valid.
+        if max_dimension is not None and max(image.shape) > 2000:
```
nayib-jose-gloria, line 1534: accidental hard-code

```suggestion
        if max_dimension is not None and max(image.shape) > max_dimension:
```

cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -1282,6 +1292,251 @@ def _check_column_availability(self):
...
+            self.errors.append(
+                f"uns['spatial'][library_id]['images']['{image_name}'] has a max dimension of 2000 pixels, "
```
nayib-jose-gloria, line 1536:
```suggestion
                f"uns['spatial'][library_id]['images']['{image_name}'] has a max dimension of {max_dimension} pixels, "
```

cellxgene_schema_cli/cellxgene_schema/validate.py
nayib-jose-gloria, line 1534: also, I would assume this is the right interpretation of the rule, but there's some vagueness with how the rule is written:

`Its largest dimension MUST be 2000 pixels.`

I will double check with Brian whether the intended validation is the largest dimension is <= 2000 or actually == 2000

cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -1282,6 +1292,251 @@ def _check_column_availability(self):
...
+                )
+
+            # spot_diameter_fullres is required.
+            if "spot_diameter_fullres" not in uns_scalefactors:
+                self.errors.append(
+                    "uns['spatial'][library_id]['scalefactors'] must contain the key 'spot_diameter_fullres'."
+                )
+            # spot_diameter_fullres is specified: proceed with validation.
+            else:
+                self._validate_float(
```
nayib-jose-gloria, line 1432: nit: I think it'd be useful to include more details in the error message (i.e. "This must be the value of the spot_diameter_fullres field from scalefactors_json.json"), in which case we probably want to pull this out of a generalized `_validate_float` method

cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -1282,6 +1292,251 @@ def _check_column_availability(self):
...
+        """
+        Determine if the assay_ontology_term_id is Visium (EFO:0010961).
+
+        :return True if assay_ontology_term_id is Visium, False otherwise.
+        :rtype bool
+        """
+        assay_ontology_term_id = self.adata.obs.get("assay_ontology_term_id")
+        return assay_ontology_term_id is not None and (assay_ontology_term_id == ASSAY_VISIUM).any()
+
+    def _validate_float(self, name: str, value: float):
```
nayib-jose-gloria, line 1492: see comments above; I think we can get rid of this.

If you disagree and would like to keep it, I think it'd make more sense to generalize into a `_validate_type` method instead and add `type` as a parameter. And generalize the docstring, which doesn't need to be specific to scalefactors

cellxgene_schema_cli/tests/fixtures/examples_validate.py
```diff
@@ -190,6 +296,36 @@
...
+        },
+    },
+}
+
+good_uns_with_slide_seqV2_spatial = {
+    "title": "A title",
+    "default_embedding": "X_umap",
+    "X_approximate_distribution": "normal",
+    "batch_condition": ["is_primary_data"],
+    "spatial": {"is_single": numpy.bool_(True)},
```
nayib-jose-gloria, line 326: let's make "is_single" a non-numpy boolean True here, since both should be valid. Keep the numpy bool in the visium uns so we can cover both cases

nayib-jose-gloria (changes requested):
looks good! a few suggested nits (optional but I think would be helpful), a couple requests, and 1 rule I'm going to clarify with Brian just in case

nayib-jose-gloria (commented):
LGTM, thank you! Just waiting on clarification from Brian re: max dimension requirement. Shouldn't block testing though, since presumably it'll come up there as well

Bento007 (changes requested):
Reviewed the validaiton changes. Looking for tests now.

cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -1282,6 +1293,248 @@ def _check_column_availability(self):
...
+            return
+
+        # is_single is required.
+        if "is_single" not in uns_spatial:
+            self.errors.append("uns['spatial'] must contain the key 'is_single'.")
+            # Exit if is_single is missing as all further checks are dependent on its value.
+            return
+
+        # is_single must be a boolean.
+        uns_is_single = uns_spatial["is_single"]
```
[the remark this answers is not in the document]
MillenniumFalconMechanic: I think the regular bool case is covered already (see [np.bool_](https://github.com/chanzuckerberg/single-cell-curation/pull/858/files#diff-bc0069b341eb34932270df020d516a80b1356a4e4f46fe97939c24631d50aa32R307) and [regular bool](https://github.com/chanzuckerberg/single-cell-curation/pull/858/files#diff-bc0069b341eb34932270df020d516a80b1356a4e4f46fe97939c24631d50aa32R326) test cases) but let me know if I am missing something here.

Bento007 (approved):

brianraymor:
> I will double check with Brian whether the intended validation is the largest dimension is <= 2000 or actually == 2000

@nayib-jose-gloria - apologies if I was @mentioned somewhere and missed it.

The answer is "== 2000" although I agree that the space ranger documentation is a bit ambiguous. I did validate earlier against the highres visium images in the corpus. There was always a 2000 present in one dimension. See this [report](https://docs.google.com/spreadsheets/d/12FbdQ2QF9pJQ8rsEXN2cEroHeeyKYNKJ/edit#gid=1186820972&range=H:H). The schema was also reviewed by 10X PM and CB.

cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -1282,6 +1293,254 @@ def _check_column_availability(self):
...
+
+        # Confirm shape of image is valid: allowed shape is (,,3).
+        if not self._is_valid_visium_image_shape(image):
+            self.errors.append(
+                f"uns['spatial'][library_id]['images']['{image_name}'] must have shape (,,3), "
+                f"it has shape {image.shape}."
+            )
+
+        # Confirm max dimension of image, if specified, is valid.
+        if max_dimension is not None and max(image.shape) > max_dimension:
```
nayib-jose-gloria, line 1538: https://github.com/chanzuckerberg/single-cell-curation/pull/858#issuecomment-2078156945

looks like we need to update this check + the accompanying comments to reflect that we want the max dimension to == max_dimension exactly

nayib-jose-gloria (changes requested):
https://github.com/chanzuckerberg/single-cell-curation/pull/858#issuecomment-2078156945

we'll need to update the max_dimension check accordingly, ready to go after that

Bento007 (approved):

MillenniumFalconMechanic:
Hi @brian-mott, this is ready for review!

cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -1305,6 +1313,245 @@ def _check_column_availability(self):
...
+        # spatial is required for supported spatial assays.
+        if uns_spatial is None:
+            self.errors.append(
+                "uns['spatial'] is required for obs['assay_ontology_term_id'] values "
+                "'EFO:0010961' (Visium Spatial Gene Expression) and 'EFO:0030062' (Slide-seqV2)."
+            )
+            return
+
+        # is_single is required.
+        if "is_single" not in uns_spatial:
```
brian-mott, line 1358: If I set `uns['spatial']` to an int, bool, or float, that raises a TypeError and stops validation: "TypeError: argument of type {'int' | 'bool' | 'float'} is not iterable." 

Other collections like lists, numpy arrays, and pandas dataframes are handled, validation completes, and the errors with validation are displayed correctly.

I've covered most of the testing but I'm still working on some items. I wanted to comment as I've uncovered issues. Let me know if you prefer the more formal review or different way to do this.

cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -1305,6 +1313,245 @@ def _check_column_availability(self):
...
+                self._validate_spatial_image_shape("fullres", uns_fullres)
+
+        # scalefactors is required.
+        if "scalefactors" not in uns_library_id:
+            self.errors.append("uns['spatial'][library_id] must contain the key 'scalefactors'.")
+        # scalefactors is specified: proceed with validation of scalefactors.
+        else:
+            # Confirm shape of scalefactors is valid: allowed keys are spot_diameter_fullres and tissue_hires_scalef.
+            uns_scalefactors = uns_library_id["scalefactors"]
+            if not self._has_no_extra_keys(uns_scalefactors, ["spot_diameter_fullres", "tissue_hires_scalef"]):
```
brian-mott, line 1452: If I set `uns['spatial'][library_id][scalefactors]` = anything other than a `dict` or `pandas.Dataframe`, I get an AttributeError: "AttributeError: '{type}' object has no attribute 'keys'" This behaves similarly to my above comment where the raised exception prematurely exits validation.

`uns['spatial'][library_id][scalefactors]` is an [output](https://scanpy.readthedocs.io/en/latest/generated/scanpy.read_visium.html) from `scanpy.read_visium()` which would suggest typing should be pretty consistent and not the wide range I've tested against. 

But we are finding through the curation towards 5.1.0 that images that are labeled as a given resolution aren't necessarily the correct resolution or set at the proper scale factor. So we do have to manually adjust these keys and values and we could easily wrangle to something other than a dict with correct keys and values.

Issue #853, opened 2024-04-20
cellxgene-schema CLI must update validation for obsm embeddings


Bento007:
# Context

See [#cell-science-data-wrangling](https://czi-sci.slack.com/archives/C024HCSH9PT/p1713559401168969?thread_ts=1713557432.295309&cid=C024HCSH9PT)

# Requirements

See #855

brianraymor:
Per my comments in slack, I was planning to open the schema/CLI issues today. I believe that this CLI issue should be moved to single-cell-curation?

PR #859, opened 2024-04-25
fix(schema 5.1.0): throw error if obsm is not type np.ndarray

Bento007:
## Reason for Change

- #853
- https://github.com/chanzuckerberg/single-cell-curation/pull/860
## Changes

- change warning to an error when obsm field is not type numpy.ndarray.
- validating that obsm ndarray that don't match spatial or X_ must have a column length of at least 1.
- split out the checks in `_validate_obsm` to make it clear why an error occured.

## Testing

- Update the unit tests to match expect behavior
- Added a test for unknown obsm keys

### DIFF
--- a/cellxgene_schema_cli/cellxgene_schema/validate.py
+++ b/cellxgene_schema_cli/cellxgene_schema/validate.py
@@ -1043,7 +1043,7 @@ class Validator:
 
                 if unknown_key and value.shape[1] < 1:
                     self.errors.append(
-                        f"All other embeddings must have at least one column. 'adata.obsm['{key}']' has columns='{value.shape[1]}'."
+                        f"All unspecified embeddings must have at least one column. 'adata.obsm['{key}']' has columns='{value.shape[1]}'."
                     )
 
                 if not unknown_key and value.shape[1] < 2:
--- a/cellxgene_schema_cli/cellxgene_schema/validate.py
+++ b/cellxgene_schema_cli/cellxgene_schema/validate.py
@@ -999,7 +999,7 @@ class Validator:
                 continue
 
             if len(value.shape) < 2 or value.shape[0] != self.adata.n_obs or value.shape[1] < 2:
-                issue_list.append(
+                self.errors.append(
                     f"All embeddings must have as many rows as cells, and at least two columns."
                     f" 'adata.obsm['{key}']' has shape of '{value.shape}'."
                 )
--- a/cellxgene_schema_cli/cellxgene_schema/validate.py
+++ b/cellxgene_schema_cli/cellxgene_schema/validate.py
@@ -983,7 +983,7 @@ class Validator:
                 issue_list = self.warnings
 
             if not isinstance(value, np.ndarray):
-                issue_list.append(
+                self.errors.append(
                     f"All embeddings have to be of 'numpy.ndarray' type, " f"'adata.obsm['{key}']' is {type(value)}')."
                 )
                 # Skip over the subsequent checks that require the value to be an array
--- a/cellxgene_schema_cli/cellxgene_schema/validate.py
+++ b/cellxgene_schema_cli/cellxgene_schema/validate.py
@@ -1301,8 +1301,7 @@ class Validator:
             for column in adata_component.columns:
                 if column in component_columns:
                     raise ValueError(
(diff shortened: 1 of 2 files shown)


cellxgene_schema_cli/tests/test_schema_compliance.py
```diff
@@ -1996,7 +1996,9 @@ def test_obsm_values_warn_start_with_X(self, validator_with_adata):
         assert validator.warnings == [
             "WARNING: Dataframe 'var' only has 4 rows. Features SHOULD NOT be filtered from expression matrix.",
             "WARNING: Embedding key in 'adata.obsm' harmony does not start with X_ and thus will not be available in Explorer",
-            "WARNING: All embeddings have to be of 'numpy.ndarray' type, 'adata.obsm['harmony']' is <class 'pandas.core.frame.DataFrame'>').",
+        ]
+        assert validator.errors == [
+            "WARNING: All embeddings have to be of 'numpy.ndarray' type, 'adata.obsm['harmony']' is <class 'pandas.core.frame.DataFrame'>')."
```
nayib-jose-gloria, line 2001: nit: we should be consistent and append errors with ERROR (blocks validation) rather than WARNING (doesn't)

MillenniumFalconMechanic (approved):

nayib-jose-gloria (commented):
Change looks good for ndarray check, but I think we're missing validation for the rest of this rule:

> The value for each str key MUST be a numpy.ndarray **of shape (n_obs, m), where n_obs is the number of rows in X and m >= 1.**

We throw a warning if a non-spatial, non-X-prefix embedding value has a row count != n_obs--that should become an error as well. And I don't think we check for column count >= 1 at all; we warn on column count < 2 (and error for spatial / X-prefix embedding keys)

nayib-jose-gloria (changes requested):
Change looks good for ndarray check, but I think we're missing validation for the rest of this rule:

> The value for each str key MUST be a numpy.ndarray **of shape (n_obs, m), where n_obs is the number of rows in X and m >= 1.**

We throw a warning if a non-spatial, non-X-prefix embedding value has a row count != n_obs--that should become an error as well. And I don't think we check for column count >= 1 at all; we warn on column count < 2 (and error for spatial / X-prefix embedding keys)

cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -998,11 +1003,26 @@ def _validate_obsm(self):
                 # Skip over the subsequent checks that require the value to be an array
                 continue
 
-            if len(value.shape) < 2 or value.shape[0] != self.adata.n_obs or value.shape[1] < 2:
-                issue_list.append(
-                    f"All embeddings must have as many rows as cells, and at least two columns."
-                    f" 'adata.obsm['{key}']' has shape of '{value.shape}'."
+            if len(value.shape) < 2:
+                self.errors.append(
+                    f"All embeddings must at least two dimensions. 'adata.obsm['{key}']' has a shape length of '{len(value.shape)}'."
```
nayib-jose-gloria, line 1008: nit: I think it must be exactly two dimensions

cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -998,11 +1003,26 @@ def _validate_obsm(self):
...
                 )
+            else:
+                if value.shape[0] != self.adata.n_obs:
+                    self.errors.append(
+                        f"All embeddings must have as many rows as cells. 'adata.obsm['{key}']' has rows='{value.shape[0]}'."
+                    )
+
+                if unknown_key and value.shape[1] < 1:
+                    self.errors.append(
+                        f"All other embeddings must have at least one column. 'adata.obsm['{key}']' has columns='{value.shape[1]}'."
```
nayib-jose-gloria, line 1018: nit: in this context, I don't think the curator reading would know what "other embeddings" mean. Maybe "any embeddings not specified in the schema reference"?

nayib-jose-gloria (approved):
1 nit about wording on an error statement but otherwise looks good. Ready for product review after that

nayib-jose-gloria:
@brian-mott ready for review!

brian-mott (approved):
Looks good, thanks Trent and Nayib!

Issue #806, opened 2024-03-11
cellxgene-schema CLI must add validation for obsm['spatial']


brianraymor:
# Context

See [**#cell-science-modalities**](https://czi-sci.slack.com/archives/C02TSE92DBM/p1709320744434789?thread_ts=1709319232.712039&cid=C02TSE92DBM) for the rationale for the addition of the  **MAY** requirement:

Curator MAY annotate if <code>uns['spatial']['is_single']</code> is <code>False</code>.

Also note that:

if <code>uns['spatial']['is_single']</code> is <code>True</code>

implicitly depends on `is_single` to enforce:

if <code>assay_ontology_term_id</code> is <a href="https://www.ebi.ac.uk/ols4/ontologies/efo/classes?obo_id=EFO%3A0010961"><code>"EFO:0010961"</code></a> for <i>Visium Spatial Gene Expression</i> or <a href="https://www.ebi.ac.uk/ols4/ontologies/efo/classes?obo_id=EFO%3A0030062"><code>"EFO:0030062"</code></a> for <i>Slide-seqV2</i>; otherwise, this key MUST NOT be present.

See [spatial](https://github.com/chanzuckerberg/single-cell-curation/blob/main/schema/5.1.0/schema.md#spatial) in schema 5.1.0.

### spatial

<table><tbody>
    <tr>
      <th>Key</th>
      <td>spatial</td>
    </tr>
    <tr>
      <th>Annotator</th>
      <td>Curator MUST annotate if <code>uns['spatial']['is_single']</code> is <code>True</code>.<br><br>Curator MAY annotate if <code>uns['spatial']['is_single']</code> is <code>False</code>.
      <br><br>Otherwise, this key MUST NOT be present.</td>
    </tr>
        <tr>
      <th>Value</th>
        <td><code>numpy.ndarray</code> with the following requirements<br><br>
          <ul>
          <li>MUST have the same number of rows as <code>X</code> and MUST include at least two columns</li>
          <li>MUST be a <a href="https://numpy.org/doc/stable/reference/generated/numpy.dtype.kind.html"><code>numpy.dtype.kind</code></a> of <code>"f"</code>, <code>"i"</code>, or "<code>u"</code></li>
[...]

PR #863, opened 2024-04-29
feat: cellxgene-schema CLI must add validation for obsm['spatial']

nayib-jose-gloria:
## Reason for Change

- #806 

## Changes

- Update validate_obsm to account for new obsm['spatial'] embedding rules (see ticket)
- validating value of 'spatial' embedding is already covered by existing code in validate_obsm because they mirror validation for X_{suffix} embeddings; just had to add an elif conditional to make sure the checks are marked as 'error' if they fail, rather than 'warning' (previous behavior for non-X prefixed embeddings). Updated tests for these checks to pass in "spatial" as well as "X_{suffix}" to ensure coverage
- Update warning msg in validate_obsm about obsm embeddings without the "X_" prefix not appearing in Explorer. "spatial" embedding key WILL appear in Explorer despite not having an X_ prefix. So far, this is the only supported case for non X_ prefixed obsm embedding keys appearing in Explorer
- Update test_validate tests for adata_with_visium and adata_with_slide_seq_v2 to test validating the entire adata, not just the spatial rules. test_validate has historically been used for end-to-end adata validation testing, and test_schema_compliance for testing specific subsetted error cases. Does not require adding new tests, since the success case of validating the full adata will also cover the success case of validating spatial rules. 

## Notes for Reviewer
- Confirmation from Brian R in Slack that 'spatial' embedding key WILL be reflected in Explorer, despite not starting with "X_"

### DIFF
--- a/cellxgene_schema_cli/cellxgene_schema/validate.py
+++ b/cellxgene_schema_cli/cellxgene_schema/validate.py
@@ -980,13 +980,14 @@ class Validator:
                     self.errors.append(
                         f"Suffix for embedding key in 'adata.obsm' {key} does not match the regex pattern {regex_pattern}."
                     )
-            else:
+            elif key.lower() != "spatial":
                 if not re.match(regex_pattern, key):
                     self.errors.append(
                         f"Embedding key in 'adata.obsm' {key} does not match the regex pattern {regex_pattern}."
                     )
                 self.warnings.append(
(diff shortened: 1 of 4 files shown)


Bento007 (approved):

MillenniumFalconMechanic (approved):
LGTM, thanks Nayib!

Issue #834, opened 2024-03-27
cellxgene-schema CLI must update validation for obs['is_primary_data']


brianraymor:
# Design

See [is_primary_data](https://github.com/chanzuckerberg/single-cell-curation/blob/main/schema/5.1.0/schema.md#is_primary_data).

The new requirement is:

 This MUST be <code>False</code> if <code>uns['spatial']['is_single']</code> is <code>False</code>. 

### is_primary_data

<table><tbody>
    <tr>
      <th>Key</th>
      <td>is_primary_data</td>
    </tr>
    <tr>
      <th>Annotator</th>
      <td>Curator MUST annotate.</td>
    </tr>
    <tr>
      <th>Value</th>
        <td><code>bool</code>. This MUST be <code>False</code> if <code>uns['spatial']['is_single']</code> is <code>False</code>. This MUST be <code>True</code> if this is the canonical instance of this cellular observation and <code>False</code> if not. This is commonly <code>False</code> for meta-analyses reusing data or for secondary views of data.
        </td>
    </tr>
</tbody></table>
<br>

nayib-jose-gloria:
@brianraymor @brian-mott for validation of obs columns that depend on spatial metadata like this--do we need to account for datasets having both spatial and non-spatial assay rows? Would that scenario ever happen?

If so, should `This MUST be False if uns['spatial']['is_single'] is False.` only apply to rows with Visium Spatial Gene Expression or Slideseqv2 assays? Or all rows as long as uns['spatial']['is_single'] exists and is False?

brianraymor:
> do we need to account for datasets having both spatial and non-spatial assay rows? Would that scenario ever happen?

No. The use of `uns` for spatial implicitly indicates that we're "allowing" only Visium Spatial Gene Expression or Slideseqv2 in the dataset (not a mixture of assays per observations). I think we may have had a hard requirement in `assay_ontology_term_id` in an earlier draft that required all observations to have the same term id when Visium Spatial Gene Expression or Slideseqv2. We could clarify. CC: @jahilton for thoughts

CC: @jahilton

jahilton:
Yes, I can see the gap in the current documentation & think we can close that by enforcement in assay_ontology_term_id. If _any_ observation is assay:Visium then _all_ must be assay:Visium. If _any_ observation is assay:Slide-seqV2 then _all_ must be assay:Slide-seqV2.
[...]

PR #865, opened 2024-04-30
 feat: update validation for obs['is_primary_data'] to enforce False if uns['spatial']['is_single'] is False

nayib-jose-gloria:
## Reason for Change

- #834 

## Changes

- add _validate_spatial_is_primary_data to _check_spatial_obs
- refactored some unit tests to use a new pytest fixture, validator with a dataset that is spatial and is_single is False (since we were reusing that set-up in a few spots)


cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -1000,13 +1000,14 @@ def _validate_obsm(self):
                     self.errors.append(
                         f"Suffix for embedding key in 'adata.obsm' {key} does not match the regex pattern {regex_pattern}."
                     )
-            else:
+            elif key.lower() != "spatial":
```
nayib-jose-gloria, line 1003: diff is from https://github.com/chanzuckerberg/single-cell-curation/issues/827 merged into main

cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -1000,13 +1000,14 @@ def _validate_obsm(self):
...
                 if not re.match(regex_pattern, key):
                     self.errors.append(
                         f"Embedding key in 'adata.obsm' {key} does not match the regex pattern {regex_pattern}."
                     )
                 self.warnings.append(
-                    f"Embedding key in 'adata.obsm' {key} does not start with X_ and thus will not be available in Explorer"
+                    f"Embedding key in 'adata.obsm' {key} is not 'spatial' nor does it start with 'X_'. Thus, it will "
```
nayib-jose-gloria, line 1009: diff is from https://github.com/chanzuckerberg/single-cell-curation/issues/827 merged into main

cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -1037,6 +1038,21 @@ def _validate_obsm(self):
...
+        has_spatial_embedding = "spatial" in self.adata.obsm
+        if is_single and not has_spatial_embedding:
+            self.errors.append(
+                "'spatial' embedding is required in 'adata.obsm' if " "adata.uns['spatial']['is_single'] is True."
+            )
+        elif is_single is None and has_spatial_embedding:
+            self.errors.append(
+                "'spatial' embedding is forbidden in 'adata.obsm' if " "adata.uns['spatial']['is_single'] is not set."
+            )
+
```
nayib-jose-gloria, line 1055: diff is from https://github.com/chanzuckerberg/single-cell-curation/issues/827 merged into main

cellxgene_schema_cli/tests/fixtures/examples_validate.py
```diff
@@ -142,7 +142,7 @@
             "HANCESTRO:0575",
             "HsapDv:0000003",
             "donor_1",
-            "nucleus",
+            "na",
```
nayib-jose-gloria, line 145: diff is from https://github.com/chanzuckerberg/single-cell-curation/issues/827 merged into main

cellxgene_schema_cli/tests/test_schema_compliance.py
```diff
@@ -1993,30 +2010,18 @@ def test_obsm_values_no_X_embedding__non_spatial_dataset(self, validator_with_ad
             "WARNING: Validation of raw layer was not performed due to current errors, try again after fixing current errors.",
         ]
 
-    @pytest.mark.parametrize(
```
nayib-jose-gloria, line 1996: split this test up into two for simplicity in the set-up

cellxgene_schema_cli/tests/test_schema_compliance.py
```diff
@@ -2025,24 +2030,9 @@ def test_obsm_values_spatial_embedding_missing__is_single_true(self, validator_w
...
+        validator = validator_with_spatial_and_is_single_false
         del validator.adata.obsm["spatial"]
-        # Format adata.obs into valid shape for Visium and is_single False.
-        validator.adata.obs.pop("array_col")
-        validator.adata.obs.pop("array_row")
-        validator.adata.obs.pop("in_tissue")
-        validator.validate_adata()
-        assert validator.errors == []
-
-    def test_obsm_values_spatial_embedding_present__is_single_false(self, validator_with_visium_assay):
```
nayib-jose-gloria, line 2039: moved this test into `test_validate.py` as `test__validate_spatial_is_single_false_ok` and renamed now that we have the fixture `adata_spatial_is_single_false` with spatial embedding present

cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -1382,6 +1382,25 @@ def _check_spatial_obs(self):
         self._validate_spatial_tissue_position("array_row", 0, 77)
         self._validate_spatial_tissue_position("in_tissue", 0, 1)
 
+        self._validate_spatial_is_primary_data()
+
+    def _validate_spatial_is_primary_data(self):
+        """
+        Validate is_primary_data for spatial datasets.
+        """
+        is_single = (
```
MillenniumFalconMechanic, line 1391: Could we use `_is_single()` here once this is rebased on `main`?

nayib-jose-gloria: ah yes good call

cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -1382,6 +1382,25 @@ def _check_spatial_obs(self):
...
+            self.adata.uns["spatial"]["is_single"]
+            if "spatial" in self.adata.uns and "is_single" in self.adata.uns["spatial"]
+            else None
+        )
+        obs = getattr_anndata(self.adata, "obs")
+        if obs is None or "is_primary_data" not in obs:
+            return
+        if is_single is False and obs["is_primary_data"].any():
+            self.errors.append(
+                "When uns['spatial']['is_single'] is False, " "obs['is_primary_data'] must be False for all rows."
```
MillenniumFalconMechanic, line 1401: There's a break in the string here (`is False, " "obs['is_primary_data']`).

nayib-jose-gloria (commented):
Note: I decided not to explicitly add a test_schema_compliance test case for allowing is_primary_data to be True or False if uns["spatial"]["is_single"] is None, because we implicitly cover this in `test_validate.py` by running `test__validate_with_h5ad_valid_and_labels`, which uses a fixture that has is_primary_data values True and False and no spatial in uns. I could be convinced to add the test case for clarity, but leaning towards no. LMK what you think

cellxgene_schema_cli/tests/test_schema_compliance.py
```diff
@@ -1993,30 +2010,18 @@ def test_obsm_values_no_X_embedding__non_spatial_dataset(self, validator_with_ad
...
-        validator.adata.obs["assay_ontology_term_id"] = assay_ontology_term_id
-        validator.adata.obs["suspension_type"] = "na"
-        validator.adata.obs.loc[:, ["suspension_type"]] = validator.adata.obs.astype("category")
         validator.validate_adata()
         assert validator.errors == []
         assert validator.is_spatial is True
 
+    def test_obsm_values_no_X_embedding__slide_seq_v2_dataset(self, validator_with_slide_seq_v2_assay):
+        validator = validator_with_slide_seq_v2_assay
+        validator.adata.uns["default_embedding"] = "spatial"
```
MillenniumFalconMechanic, line 2023: Looks like this test is missing `validator.validate_adata()` and the corresponding assertions.

cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -1382,6 +1382,20 @@ def _check_spatial_obs(self):
...
+        self._validate_spatial_is_primary_data()
+
+    def _validate_spatial_is_primary_data(self):
+        """
+        Validate is_primary_data for spatial datasets.
+        """
+        obs = getattr_anndata(self.adata, "obs")
+        if obs is None or "is_primary_data" not in obs:
+            return
+        if self._is_single() is False and obs["is_primary_data"].any():
```
MillenniumFalconMechanic, line 1394: How are "secondary errors" typically handled? For example, should this error be reported for a non-spatial dataset that has (incorrectly) specified `uns.spatial.is_single = False`?

MillenniumFalconMechanic:
Looks good, thanks Nayib! I have added a couple of comments.

cellxgene_schema_cli/tests/test_schema_compliance.py
```diff
@@ -1993,30 +2010,18 @@ def test_obsm_values_no_X_embedding__non_spatial_dataset(self, validator_with_ad
...
-        self, validator_with_adata, assay_ontology_term_id, adata_spatial
-    ):
-        validator = validator_with_adata
-        validator.adata.obsm["harmony"] = validator.adata.obsm["X_umap"]
-        validator.adata.uns = adata_spatial.uns
-        validator.adata.uns["default_embedding"] = "harmony"
-        validator.adata.obsm["spatial"] = validator.adata.obsm["X_umap"]
+    def test_obsm_values_no_X_embedding__visium_dataset(self, validator_with_visium_assay):
+        validator = validator_with_visium_assay
+        validator.adata.uns["default_embedding"] = "spatial"
```
MillenniumFalconMechanic, line 2015: I am not familiar with the complete, required shape for `adata` here but I see there are a couple of differences between the previous version and `validator_with_visium_assay` (e.g. `validator.adata.obsm["harmony"]` is no longer defined, `validator.adata.uns["default_embedding"]` is now `spatial`). Are these updates expected/OK?

cellxgene_schema_cli/tests/test_schema_compliance.py
nayib-jose-gloria: sad, it must've gotten lost along the way during a merge resolution

cellxgene_schema_cli/cellxgene_schema/validate.py
nayib-jose-gloria: since we accumulate all errors by appending to "self.errors", something like that would be handled wherever the uns check is being done. In this case, we should handle that in check_spatial_uns. I'll double check if we already are and add it in if not.

cellxgene_schema_cli/tests/test_schema_compliance.py
nayib-jose-gloria: yes, its expected

for non-spatial datasets, not having an X-prefix embedding like `X_umap` should result in an error (see `test_obsm_values_no_X_embedding__non_spatial_dataset`). 

However, for spatial datasets, we do not require an X-prefix embedding, just 'spatial' embedding. This test case was originally modeled after `test_obsm_values_no_X_embedding__non_spatial_dataset`, which used 'harmony' as its non-X-prefix embedding to trigger the error. But I realized spatial is already a non-X-prefix embedding so we didn't need to add another one to test.

cellxgene_schema_cli/cellxgene_schema/validate.py
nayib-jose-gloria: Oh wait I see what you mean--yes, we do report secondary errors like this. That way, if the problem was that they did mean for it to be spatial but accidentally used the wrong assay, we still tell them in 1 run all the other issues associated with it. Generally, we try to report every possible error in one go even if they'd all be fixed with 1 upstream correction. With some exceptions--we don't validate the X matrix, for instance, if there are issues with dependent columns before then, because its an expensive operation to run if we know there will be failures. But we do print a warning letting them know that this was the case--`WARNING: "Validation of raw layer was not performed due to current errors, try again after fixing current errors."`

nayib-jose-gloria: I do think its fine that we exit early w/o checking spatial values in `check_spatial_uns` if assay is not 'EFO:0010961' (Visium Spatial Gene Expression) or 'EFO:0030062' (Slide-seqV2), since curators have not requested otherwise and the errors would end up being the same for each one (i.e. `this key should not exist if its not visium` X times)

MillenniumFalconMechanic: Got it, thanks Nayib!

MillenniumFalconMechanic (approved):
LGTM, thanks Nayib!

nayib-jose-gloria:
@brian-mott ready for your review!

brian-mott:
Looks good, thanks Nayib!

Issue #828, opened 2024-03-27
cellxgene-schema CLI must update validation in X (Matrix Layers)


brianraymor:
# Design

See [X (Matrix Layers)](https://github.com/chanzuckerberg/single-cell-curation/blob/main/schema/5.1.0/schema.md#x-matrix-layers).

| Assay | "raw" required? | "raw" location | "normalized" required? | "normalized" location |
|-|-|-|-|-|
| Visium Spatial Gene Expression | REQUIRED. Values MUST be de-duplicated molecule counts. All non-zero values MUST be positive integers stored as `numpy.float32`.<br><br>If <code>uns['spatial']['is_single']</code> is <code>False</code> then each cell MUST contain at least one non-zero value.<br><br>If <code>uns['spatial']['is_single']</code> is <code>True</code> then the unfiltered feature-barcode matrix (<code>raw_feature_bc_matrix</code>) MUST be used. See <a href="https://www.10xgenomics.com/support/software/space-ranger/analysis/outputs/space-ranger-feature-barcode-matrices">Space Ranger Feature-Barcode Matrices</a>. This matrix MUST contain 4992 rows. If the <code>obs['in_tissue']</code> value is <code>1</code>, then the cell MUST contain at least one non-zero value. If any <code>obs['in_tissue']</code> values are <code>0</code>, then at least one cell corresponding to a <code>obs['in_tissue']</code> with a value of <code>0</code> MUST contain a non-zero value.| `AnnData.raw.X` unless no "normalized" is provided, then `AnnData.X` | STRONGLY RECOMMENDED | `AnnData.X` |
|||||

Bento007:
This change is blocking lattice from uploading new datasets set portal. After this is merged we need to create a patched version of the CLI with this change that lattice can use to upload datasets to the portal.

PR #876, opened 2024-05-03
feat: update X matrix validation to validate Visium assays

nayib-jose-gloria:
## Reason for Change

- #828

## Changes

- refactor _has_valid_raw to split raw data validation into two branches: validating visium assay datasets with is_single True and in_tissue 0 values (introduces totally new validation rule that requires row-by-row iteration: rows in the matrix corresponding to `obs['in_tissue'] == 0` rows CAN be all zeros--as long as there is at least ONE non-zero value among the group of in_tissue == 0 rows)
- I chose not to introduce matrix chunking to the visium + is_single + in_tissue 0 data validation branch because we previously validate that its length is 4992 exactly, which is less than our default chunking size. Memory shouldn't be a concern in this case 
- moved typical data validation pathway for all other cases to _validate_raw_data method
       - this includes datasets where visium + is_single + in_tissue is 1 for all cases because that pathway has the same value ruleset as the other assays we validate for. The only diff is validating 4992 rows for visium + is_single datasets, so that check is done outside of this helper function.

### DIFF
--- a/cellxgene_schema_cli/cellxgene_schema/validate.py
+++ b/cellxgene_schema_cli/cellxgene_schema/validate.py
@@ -1188,9 +1188,8 @@ class Validator:
                 elif not all(np.apply_along_axis(np.any, axis=1, arr=matrix_chunk)):
                     has_row_of_zeros = True
 
-            if not has_invalid_nonzero_value:
-                if self._matrix_has_invalid_nonzero_values(matrix_chunk):
-                    has_invalid_nonzero_value = True
+            if not has_invalid_nonzero_value and self._matrix_has_invalid_nonzero_values(matrix_chunk):
+                has_invalid_nonzero_value = True
 
             if has_row_of_zeros and has_invalid_nonzero_value:
                 # Fail fast, exit loop and report
--- a/cellxgene_schema_cli/cellxgene_schema/validate.py
+++ b/cellxgene_schema_cli/cellxgene_schema/validate.py
@@ -25,6 +25,9 @@ ONTOLOGY_PARSER = OntologyParser(schema_version=f"v{schema.get_current_schema_ve
 ASSAY_VISIUM = "EFO:0010961"
 ASSAY_SLIDE_SEQV2 = "EFO:0030062"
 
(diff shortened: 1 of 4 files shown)


cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -1142,37 +1146,114 @@ def _has_valid_raw(self, force: bool = False) -> bool:
...
-                    if is_sparse_matrix:
-                        row_indices, _ = matrix_chunk.nonzero()
-                        if len(set(row_indices)) != matrix_chunk.shape[0]:
-                            has_row_of_zeros = True
-                    # else, must be dense matrix, confirm that all rows have at least 1 nonzero value
-                    elif not all(np.apply_along_axis(np.any, axis=1, arr=matrix_chunk)):
-                        has_row_of_zeros = True
-
-                if not has_invalid_nonzero_value:
-                    data = matrix_chunk if isinstance(matrix_chunk, np.ndarray) else matrix_chunk.data
```
nayib-jose-gloria, line 1159: moved to _matrix_has_invalid_nonzero_values for re-use

cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -38,6 +41,7 @@ def __init__(self, ignore_labels=False):
         self.schema_def = dict()
         self.schema_version: str = None
         self.ignore_labels = ignore_labels
+        self.visium_and_is_single_true_matrix_size = VISIUM_AND_IS_SINGLE_TRUE_MATRIX_SIZE
```
Bento007, line 44: nit: in the future you can patch `VISIUM_AND_IS_SINGLE_TRUE_MATRIX_SIZE` in the tests to avoid having this class variable. Just one less duplicated variable.

Bento007 (approved):

MillenniumFalconMechanic (approved):

brian-mott (approved):
Look good everyone, works as defined with spatial and non-spatial datasets and +/- a normalized matrix. Thanks!

Issue #862, opened 2024-04-29
fix: type checking for spatial


Bento007:
- [x] If I set uns['spatial'][library_id][scalefactors] = anything other than a dict or pandas.Dataframe, I get an AttributeError: "AttributeError: '{type}' object has no attribute 'keys'" This behaves similarly to my above comment where the raised exception prematurely exits validation.
  
  uns['spatial'][library_id][scalefactors] is an [output](https://scanpy.readthedocs.io/en/latest/generated/scanpy.read_visium.html) from scanpy.read_visium() which would suggest typing should be pretty consistent and not the wide range I've tested against.
  
  But we are finding through the curation towards 5.1.0 that images that are labeled as a given resolution aren't necessarily the correct resolution or set at the proper scale factor. So we do have to manually adjust these keys and values and we could easily wrangle to something other than a dict with correct keys and values.
- [x] If I set `uns['spatial']` to an int, bool, or float, that raises a TypeError and stops validation: "TypeError: argument of type {'int' | 'bool' | 'float'} is not iterable." 

Originally posted by @brian-mott in:
 - https://github.com/chanzuckerberg/single-cell-curation/pull/858#discussion_r1581581088
 - https://github.com/chanzuckerberg/single-cell-curation/pull/858#discussion_r1581606399

PR #886, opened 2024-05-07
fix: check type of uns["spatial"][library_id] is dict

Bento007:
## Reason for Change

- #862 

## Changes

- add acheck that uns["spatial"][library_id] is dict

## Testing

- add new unit test

## Notes for Reviewer


brian-mott (approved):
Looks good, now only dict is allowed for `uns['spatial'][library_id]` values.
Thanks!

MillenniumFalconMechanic (approved):

PR #891, opened 2024-05-09
chore: update requirements.txt

Bento007:
## Reason for Change

- dependencies are out of data.
- There are features in anndata 0.10 we'd like to use.

## Changes

- remove subtest test from test_determine_seurat_convertibility because it throws `ValueError: Observations annot. `var` must have as many rows as `X` has columns (4), but has 3 rows.` This wasn't a valid datasets to begin with and now anndata blocks us from creating such a dataset.
- test_fail_categorical_bool no longer valid with updated anndata
- updated tiledb version for schema_bump_dry_run_genes
- update anndata to v0.12
- update pandas to v2.x.x
- pin dependencies to latest major version.

## Testing

- update unit test

## Notes for Reviewer
schema_bump_dry_run_genes unit test started failing due to reasons outside of the changes of this PR. The tests are failing on main as well https://github.com/chanzuckerberg/single-cell-curation/actions/runs/9486687113/job/27234784964.


cellxgene_schema_cli/requirements.txt
```diff
@@ -1,12 +1,12 @@
...
+PyYAML<7
 wheel==0.40.0
-scipy<1.13.0    # scipy 1.13.0 is not compatible with anndata <=0.10.6, revisit before next release
-semver==3.0.0
-xxhash==3.3.0
-matplotlib==3.7.3
+scipy==1.13.0    # scipy 1.13.0 is not compatible with anndata <=0.10.6, revisit before next release
+semver<4
+xxhash<4
+matplotlib<4
```
nayib-jose-gloria, line 12: the schema reference provides a definition for supported "colors" to use as values for "colors" keys in the uns. We should double check that having that list is compatible with updating this library. Or maybe include which version of matplotlib we use to validate in the schema reference.

cellxgene_schema_cli/requirements.txt
```diff
@@ -1,12 +1,12 @@
...
+Cython<4
+numpy<2
+pandas<2
+PyYAML<7
 wheel==0.40.0
-scipy<1.13.0    # scipy 1.13.0 is not compatible with anndata <=0.10.6, revisit before next release
-semver==3.0.0
-xxhash==3.3.0
-matplotlib==3.7.3
+scipy==1.13.0    # scipy 1.13.0 is not compatible with anndata <=0.10.6, revisit before next release
```
nayib-jose-gloria, line 9: can bump if we do bump anndata

Bento007: I'll unpin and run again for 5.2

cellxgene_schema_cli/requirements.txt
Bento007: looking at the matplotlib change logs, there have been no changes to colors between 3.7.3 and 3.9.0. The portal has all the [named colors](https://github.com/chanzuckerberg/single-cell-data-portal/blob/6357aa21d5a5da3986310f1dd595769f0e03422e/backend/common/utils/color_conversion_utils.py) hardcoded and those also haven't change. I think we should remove this as a requirement from the CLI and hardcode `matplotlib.colors` somewhere in the code.

cellxgene_schema_cli/tests/test_schema_compliance.py
```diff
@@ -2433,19 +2433,21 @@ def test_obs_added_tissue_type_label__unknown(self, validator_with_adata):
         obs.at["Y", "tissue_type"] = "cell culture"  # Already set in example data, just setting explicitly here
         obs.at["Y", "tissue_ontology_term_id"] = "unknown"  # Testing this term case
         validator_with_adata.validate_adata()  # Validate
-        AnnDataLabelAppender(validator_with_adata)._add_labels()  # Annotate
+        labeler = AnnDataLabelAppender(validator_with_adata)
```
Bento007, line 2436: a new adata container is being created in the `AnnDataLabelAppender`, so we need to check that the `AnnDataLabelAppender.adata` is being updated when we call `_add_labels`

Bento007:
refresh PR

Bento007:
> Mind looking into the failing test before merging?

unit-test-genes-dry-run has started failing on main as well https://github.com/chanzuckerberg/single-cell-curation/actions/runs/9486687113/job/27234784964

cellxgene_schema_cli/requirements.txt
```diff
@@ -1,12 +1,11 @@
...
-scipy<1.13.0    # scipy 1.13.0 is not compatible with anndata <=0.10.6, revisit before next release
-semver==3.0.2
-xxhash==3.4.1
-matplotlib==3.9.0
+anndata<0.12
+cellxgene-ontology-guide==1.0.0 # update before a schema migration
+click<9
+Cython<4
+numpy<2
+pandas<3
```
Bento007, line 6: I verified this works with 1.4.4 and 2.2.2. The portal still uses 1.4.4, so we will let portal decide what version of pandas is used as long as it's <3

Issue #877, opened 2024-05-06
Do not check if X has zero values and release a 5.0 patch.


Bento007:
# Problem

See [Problem slide](https://docs.google.com/presentation/d/1uQ4iCRyR416y7oi3iukaNCZOqDX64sA0x67XzEdWQIY/edit#slide=id.g2c9903b59f1_0_8) and [Ideas slide](https://docs.google.com/presentation/d/1uQ4iCRyR416y7oi3iukaNCZOqDX64sA0x67XzEdWQIY/edit#slide=id.g2c9903b59f1_0_0).

Per 5.1.0 X (Matrix Layers), a Visium dataset may contain zero values:
REQUIRED … If any obs['in_tissue'] values are 0, then at least one cell corresponding to a obs['in_tissue'] with a value of 0 MUST contain a non-zero value.
But the existing 5.0.0 X (Matrix Layers) requirements blocks curators from submitting such a revision of a Visium dataset prior to migration:
REQUIRED. Values MUST be de-duplicated molecule counts. Each cell MUST contain at least one non-zero value. All non-zero values MUST be positive integers stored as numpy.float32.

# Solution

We will create a patch release of the CLI for lattice that does not check for zero values in the X matrix. This will unblock lattice, and allow those datasets to be fully validated during the 5.1 migration.

nayib-jose-gloria:
Note: If Explorer doesn't support just "spatial" embedding yet, need to allow "X_spatial" obsm embeddings in the patch (this will be forbidden in 5.1.0 proper)

PR #895, opened 2024-05-09
Release version 5.0.3

Bento007:
## Reason for Change

- #877 

## Changes

- allow X_spatial

## Notes for Reviewer


.bumpversion.cfg
```diff
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 5.0.2
+current_version = 5.0.3-rc.0
```
nayib-jose-gloria, line 2: should we remove the rc, or keep it and pin that version in data-portal?

nayib-jose-gloria (commented):
looks good! 1 question re: version with rc

Also, mind building this locally + validating against one of the spatial datasets?

nayib-jose-gloria (approved):
actually i had a spatial downloaded already so I built and validated against it. Looks good from an engineering POV, but I still get this error:

`ERROR: uns['spatial'][library_id]['images']['hires'] must be of type numpy.uint8, it is float32.
`
I think the conclusion was that curators were ok to fix this on their ends, right? If we confirm that, we're ready to go.

jahilton:
>I think the conclusion was that curators were ok to fix this on their ends, right? If we confirm that, we're ready to go.

Yes, we are prepared to start curating the images to uint8 instead of float32
and we plan for migrate.py to convert anything already in the portal

PR #927, opened 2024-05-16
chore: dependabot tuning

Bento007:
## Reason for Change

adjusting the dependabot settings to avoid commit to pre-major versions which may include breaking changes.

## Changes

- treat anndata and tiledb special because they don't have a major release and they need special consideration before upgrading.
- remove pyarrow requirement because it is unused in the actual code.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -27,6 +27,8 @@ updates:
         update-types:
           - patch
           - minor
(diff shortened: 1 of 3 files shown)


Bento007:
Merging in to see if the dependabot changes work as expected. Will back out if it doesn't work.

Issue #963, opened 2024-07-08
Process h5ad in constant memory


Bento007:
## Motivation

- Make memory requirements predictable
- Reduce datasets processing failure due to OOM

## Definition of Done
- Validate and write datasets in constant memory
- eliminate the use of `Anndata.to_memory` in the code.

## Tasks
- It's possible this may not be possible for write scenarios with the current anndata library
- accomplish this for the X and raw.X Matrix

## Notes
- if we may need to suggest changes to the anndata library to accomplish this.
- conscider looking at [dask arrays](https://anndata.readthedocs.io/en/latest/tutorials/notebooks/anndata_dask_array.html#Initalizing)
- from the [anndata docs](https://anndata.readthedocs.io/en/latest/tutorials/notebooks/anndata_dask_array.html#Initalizing):
> ### IO operations
> Read/Write operations on h5ad and Zarr are supported. One should note that the lazy objects are materialized when this is called. For now, the anndata loaded from file won’t be loaded with dask arrays in it.
- [h5py](https://docs.h5py.org/en/latest/index.html) is the library anndata uses for h5ad IO.
- [Reading h5py file directly from s3 ](https://docs.h5py.org/en/latest/high/file.html#file-drivers)
- [chunk cache](https://docs.h5py.org/en/latest/high/file.html#chunk-cache)

Bento007:
enforcing canonical format is the biggest consumer of memory because it requires reading in the whole X and raw.X matrix. If we were to instead warn that a matrix is not in canonical format we can avoid reading the memory explosion. This PR does [this](https://github.com/chanzuckerberg/single-cell-curation/pull/964)

Bento007:
Another recommendation is to consolidate the [sparse checkers in portal](https://github.com/chanzuckerberg/single-cell-data-portal/blob/c84e910e3b483b153121bfc73429a57029b0dac0/backend/layers/processing/utils/matrix_utils.py#L6-L6) and [cli ](https://github.com/chanzuckerberg/single-cell-curation/blob/adbeba6be6df1a462c1d00dedc8271829f2a05b7/cellxgene_schema_cli/cellxgene_schema/validate.py#L895-L895) into one. The CLI version is memory efficient and makes use of chunks.

Bento007:
Resulting issues
- https://github.com/chanzuckerberg/single-cell-data-portal/issues/7310
- https://github.com/chanzuckerberg/single-cell-data-portal/issues/7309
[...]

PR #964, opened 2024-07-11
fix: avoid reading matrix into memory during validation

Bento007:
## Reason for Change

- related to https://github.com/chanzuckerberg/single-cell-curation/issues/963
- The new anndata version makes it easier to read a sparse matrix in chunk. We can take advantage of this to reduce memory consumption.
 
## Changes
- read [SparseDataset](https://anndata.readthedocs.io/en/latest/generated/anndata.experimental.sparse_dataset.html) in chunks to avoid reading into memory.
- Cache get_matrix_format. This is a minor performance improvement to reduce the number of disk reads for a backed dataset.
- remove `import scipy` and use `sparse` to reduce python lookups
- add file path to log messages to make debugging easier.
- Add comments relating to future optimizations

## Testing

- verified unit test all pass.

## Notes for Reviewer


cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -947,9 +956,10 @@ def _validate_seurat_convertibility(self):
             to_validate.append((self.adata.raw.X, "raw.X"))
         # Check length of component arrays
         for matrix, matrix_name in to_validate:
-            matrix_format = get_matrix_format(self.adata, matrix)
+            matrix_format = self.get_matrix_format(matrix_name, matrix)
             if matrix_format in SPARSE_MATRIX_TYPES:
                 effective_r_array_size = self._count_matrix_nonzero(matrix_name, matrix)
+                self._check_canonical_format(matrix_name, matrix)
```
Bento007, line 962:
```suggestion
```

Issue #990, opened 2024-08-20
cellxgene-schema CLI must update annotation of var['feature_length']


brianraymor:
## Implementation notes copied from #960

@Bento007 wrote:
> The code changes will be easy. There are no unit tests at the moment which should be written, and this will make this work take longer. I estimate 2-3 days.

@nayib-jose-gloria wrote:

> Note: after re-writing `_get_gene_lengths_from_gtf` to use the new approach, we will also need to re-run the Gene Processing [script](https://github.com/chanzuckerberg/single-cell-curation/blob/main/cellxgene_schema_cli/Makefile#L5) and push the [resulting updated CSVs to */genecode_files/*](https://github.com/chanzuckerberg/single-cell-curation/tree/main/cellxgene_schema_cli/cellxgene_schema/gencode_files)

## Design

There are *two* changes in requirements.

1. The feature length for the "spike-in" `feature_biotype` must now be calculated. Previously, the feature length calculation was limited to the "gene" `feature_biotype`.
2. The calculation is updated from "merged" to "median" using GTFtools implementation.

See [**feature_length**](https://github.com/chanzuckerberg/single-cell-curation/blob/main/schema/5.2.0/schema.md#feature_length).

Implementation details from @pablo-gar :

Current implementation in our code base is here: 

https://github.com/chanzuckerberg/single-cell-curation/blob/0c77179d2e794846861f8109c037b723507959cb/cellxgene_schema_cli/scripts/gene_processing.py#L129

The new implementation should be taken from here:

https://github.com/RacconC/gtftools/blob/140fc21003a565a0f69b5176db734b9a04a004a4/gtftools/gtftools.py#L670-L688

PR #1005, opened 2024-09-04
feat: use isoform lengths median to determine gene feature_length + write feature_length for 'spike-in' features in var

nayib-jose-gloria:
## Reason for Change

- #990 

## Changes

- changed the calculation of gene length from "merged" to "median" using [GTFtools](https://github.com/RacconC/gtftools/blob/140fc21003a565a0f69b5176db734b9a04a004a4/gtftools/gtftools.py#L670-L688) implementation.
- we previously set feature_length to 0 for rows in var with feature_type "spike-in", despite having the feature length pre-calculated. We will now write the feature_length in write-labels for all feature types; this command is run after an H5AD is uploaded + validated in the CxG data portal
- ran gene_processing.py and generated new gene metadata files (genes_*.csv.gz) using new feature-length calculation. these files are the source of truth for the feature-length annotated onto H5AD var df after upload. 

## Notes for Reviewer
- no changes to genes_ercc.csv.gz due to always having a different method for calculating feature length, unrelated to above
- no changes to genes_sars_cov_2.csv.gz due to each gene in that file only having 1 isoform in gtf, so no changes between merged vs. median calculation


Bento007 (approved):

sidneymbell (approved):
This implementation LGTM! Very tidy. Thank you! 

Note that I haven't had time to review the output files in detail, but the code looks good.

PR #1032, opened 2024-10-07
atac-seq schema

jahilton:
Not ready to merge. PR for review purposes only.
As the standards & how those are written have solidified, we can update the PR to merge them into a draft of the full schema. It only includes the fragment file standards, but I will add standards for a genome track data product.


schema/atac_schema.md
```diff
@@ -0,0 +1,473 @@
+## scATAC-seq assay types
+
+<i>paired assay</i>: any descendant of <a href="https://www.ebi.ac.uk/ols4/ontologies/efo/classes?obo_id=EFO%3A0010891"><code>"EFO:0010891"</code></a> for <i>scATAC-seq</i> that is also a descendant of <a href="https://www.ebi.ac.uk/ols4/ontologies/efo/classes?obo_id=EFO%3A0008913"><code>"EFO:0008913"</code></a> for <i>single-cell RNA sequencing</i>
+
```
brianraymor, line 4: paired is effectively only "10x multiome" ? I'm trying to understand the rationale for writing the general definition. For example, are there pending NTR(s) for other assays that would meet the requirement?

schema/atac_schema.md
```diff
@@ -0,0 +1,473 @@
...
+<i>unpaired assay</i>: <a href="https://www.ebi.ac.uk/ols4/ontologies/efo/classes?obo_id=EFO%3A0010891"><code>"EFO:0010891"</code></a> for <i>scATAC-seq</i> or its descendants that is not a descendant of <a href="https://www.ebi.ac.uk/ols4/ontologies/efo/classes?obo_id=EFO%3A0008913"><code>"EFO:0008913"</code></a> for <i>single-cell RNA sequencing</i>
+
```
brianraymor, line 6: Is there an expectation that there would be support for scATAC-seq (cell index)and scATAC-seq (Microfluidics)

schema/atac_schema.md
```diff
@@ -0,0 +1,473 @@
...
+* the <code>var['feature_reference']</code> values MUST include one of <code>"NCBITaxon:9606"</code> for <i>Homo sapiens</i> or <code>"NCBITaxon:10090"</code> for <i>Mus musculus</i>, but not both. The value that is present will determine the appropriate Chromosome Table for standards.
+
+If the <code>obs['assay_ontology_term_id']</code> values are all <i>paired assays</i> then a fragment file MAY be attached to the Dataset.
+
+If the <code>obs['assay_ontology_term_id']</code> values are all <i>unpaired assays</i> then a fragment file MUST be attached to the Dataset.
+
+## Fragment File
+
+This MUST be a gzipped tab-separated values (TSV) file.
+
```
brianraymor, line 21: This section is a minor _variation_ of [Column Definitions](https://www.10xgenomics.com/support/software/cell-ranger-arc/latest/analysis/outputs/fragments-file#column-definitions) for the most part?

brianraymor: Is there a standard name for the fragment file?

schema/atac_schema.md
```diff
@@ -0,0 +1,473 @@
...
+* the <code>obs['assay_ontology_term_id']</code> values MUST all be <i>paired assays</i> or MUST all be <i>unpaired assays</i>
+* the <code>obs['is_primary_data']</code> values MUST be all `True`
+* the <code>var['feature_reference']</code> values MUST include one of <code>"NCBITaxon:9606"</code> for <i>Homo sapiens</i> or <code>"NCBITaxon:10090"</code> for <i>Mus musculus</i>, but not both. The value that is present will determine the appropriate Chromosome Table for standards.
+
+If the <code>obs['assay_ontology_term_id']</code> values are all <i>paired assays</i> then a fragment file MAY be attached to the Dataset.
+
+If the <code>obs['assay_ontology_term_id']</code> values are all <i>unpaired assays</i> then a fragment file MUST be attached to the Dataset.
+
+## Fragment File
+
```
brianraymor, line 19: The tabix documentation states:

> The input data file must be position sorted and compressed by bgzip which has a gzip(1) like interface.

The 10X documentation only mentions:

> The data is block-gzipped to allow indexing and to save disk space.

schema/atac_schema.md
jahilton: No specific NTR issues open, but this is future-proofing. And we would quickly go to EFO with suggestions to group more under "single-cell RNA sequencing" to ensure future NTRs get links appropriately.

schema/atac_schema.md
jahilton: Yes.

schema/atac_schema.md
jahilton: >This section is a minor variation of Column Definitions for the most part?

Correct

>Is there a standard name for the fragment file?

As in _file_ name? Similar to what we see with the `.h5` raw matrices, it's the pipeline output file amended somehow with a library/sample name to distinguish each file in their study.
So no.

schema/atac_schema.md
```diff
@@ -0,0 +1,473 @@
...
+        </td>
+    </tr>
+</tbody></table>
+<br>
+
+## Fragment File index
+
+CELLxGENE Discover MUST generate a <a href="https://www.htslib.org/doc/tabix.html">tabix</a> index of the fragment intervals from the fragment file. The file name MUST be the name of the corresponding fragment file appended with `.tbi`.
+
+## Chromosome Tables
```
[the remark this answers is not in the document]
jahilton: Yes. I will share that.

schema/atac_schema.md
jahilton: So `This MUST be a gzipped tab-separated values (TSV) file.` is not strict enough?
`This MUST be a tab-separated values (TSV) file position-sorted and compressed by bgzip.`

brianraymor: I'd recommend that we test to confirm the requirement.

schema/atac_schema.md
brianraymor: Depending on the final submission design, we MAY want to standardize the naming. For example, `<dataset canonical UUID>-fragments.tsv.gz`.

schema/atac_schema.md
brianraymor: Is there the possibility that an unexpected term might be added causing validation issues - a'la _One of These Things (Is Not Like the Others)_. ? I guess we could simply include it in the review of EFO updates. Just thinking out loud.

jahilton: We wouldn't accept a new assay without reviewing it in EFO or adding it there ourselves. So the edge case would be a term we have reviewed and accepted later gets the ontology links updated to where it suddenly fits the criteria. So possible? yes. Though seems unlikely.

schema/atac_schema.md
```diff
@@ -0,0 +1,473 @@
...
+## Fragment File Dataset Criteria
+
+A Dataset MUST meet each of the following criteria in order to be eligible for an attached Fragment File:
+* the <code>obs['assay_ontology_term_id']</code> values MUST all be <i>paired assays</i> or MUST all be <i>unpaired assays</i>
+* the <code>obs['is_primary_data']</code> values MUST be all `True`
+* the <code>var['feature_reference']</code> values MUST include one of <code>"NCBITaxon:9606"</code> for <i>Homo sapiens</i> or <code>"NCBITaxon:10090"</code> for <i>Mus musculus</i>, but not both. The value that is present will determine the appropriate Chromosome Table for standards.
+
+If the <code>obs['assay_ontology_term_id']</code> values are all <i>paired assays</i> then a fragment file MAY be attached to the Dataset.
+
+If the <code>obs['assay_ontology_term_id']</code> values are all <i>unpaired assays</i> then a fragment file MUST be attached to the Dataset.
```
Bento007, line 16: This requirement will complicate validation in the ingest pipeline. We will have to wait for the fragment file to be present before knowing if the anndata is valid. If more requirements like this are expected in the future then we can pay the engineer cost to simplify change like this for the future.

schema/atac_schema.md
Bento007: Have we looked into rolling our own vs tabix?

jahilton: I don't know what "rolling our own" looks like, so no haven't looked into it.
The common analysis software consumer both the tsv & the index together so it must be interoperable with those.

schema/atac_schema.md
Bento007: > It would seem easier to simply block publication of the collection if a fragment file is not attached to a dataset?

💯 we need to block in this case. I'm thinking about how the backend will process this when multiple files are required before being a valid dataset.

brianraymor: > when multiple files are required before being a valid dataset.

I'm suggesting that the dataset is valid, but the collection is invalid when there's a missing/required fragment file.

* Upload and validate dataset - set collection as invalid due to missing requirement fragment file.
* Upload and validate fragment file - set collection as valid if fragment validation passes. 

Another approach is to always upload+validate the dataset and fragment file together. The two files are tarred or gzipped together.

schema/atac_schema.md
```diff
@@ -0,0 +1,473 @@
...
+
+## Fragment File
+
+This MUST be a gzipped tab-separated values (TSV) file.
```
Bento007, line 20: So the ingest format is going to be this tsv.gz, and we also want to make this same format available to the users to download?

schema/drafts/atac_schema.md
```diff
@@ -0,0 +1,511 @@
...
+
+Asset file specifications TBD based on the visualization solution. Accepting <a href="https://genome.ucsc.edu/goldenpath/help/bigWig.html">.bigWig format</a> is a requirement.
+
+## Chromosome Tables
+
+As determined by the reference assembly used by the gene annotation versions pinned for this version of the schema. Only chromosomes or scaffolds that have at least one gene feature present are included.
+
+### <a href="https://ftp.ebi.ac.uk/pub/databases/gencode/Gencode_human/release_44/GRCh38.primary_assembly.genome.fa.gz">human (GRCh38.p14)</a>
+
+<table>
```
Bento007, line 147: Is this the expected order that chromosome should be sorted by? When doing a alphabetical sort I get something very different.

```
['chrY',
 'chrX',
 'chrM',
 'chr9',
 'chr8',
 'chr7',
 'chr6',
 'chr5',
 'chr4',
 'chr3',
 'chr22',
 'chr21',
 'chr20',
 'chr2',
 'chr19',
 'chr18',
 'chr17',
 'chr16',
 'chr15',
 'chr14',
 'chr13',
 'chr12',
 'chr11',
 'chr10',
 'chr1',
 'KI270750.1',
 'KI270744.1',
 'KI270734.1',
 'KI270733.1',
 'KI270731.1',
 'KI270728.1',
 'KI270727.1',
 'KI270726.1',
 'KI270721.1',
 'KI270713.1',
 'KI270711.1',
 'KI270442.1',
 'GL000225.1',
 'GL000220.1',
 'GL000219.1',
 'GL000218.1',
 'GL000216.2',
 'GL000213.1',
 'GL000205.2',
 'GL000195.1',
 'GL000194.1',
 'GL000009.2']
```

Bento007: NVM, this depend on the ordering that tabix expects.

schema/atac_schema.md
jahilton: I added the naming convention to the updated version

PR #1061, opened 2024-10-23
feat: add fruit fly gencode file

joyceyan:
## Reason for Change

https://app.zenhub.com/workspaces/cs-platform-661fffc5c6339904f8f42cb2/issues/gh/chanzuckerberg/single-cell-curation/1044

## Changes

adds the fruit fly gencode file

## Testing

updated `good_var` with a fruit fly gene, and updated tests accordingly

## Notes for Reviewer


cellxgene_schema_cli/scripts/gene_processing.py
```diff
@@ -84,14 +84,19 @@ def _parse_gtf(self, gtf_path: str, gene_info_description: str):
 
                 # Extract features (column 9 of GTF)
                 current_features = gtf_tools._get_features(line)  # type: ignore
+
+                # Set gene_name to gene_id if not present in GTF line
```
joyceyan, line 88: jason confirmed we can just set the gene name to be the gene id: https://czi-sci.slack.com/archives/C02TSE92DBM/p1729717238997349

cellxgene_schema_cli/scripts/gene_processing.py
```diff
@@ -248,13 +253,29 @@ def process_gene_infos(self, gene_infos: dict) -> None:
                     )
                 self.write_gzip(output_to_print, new_file)
 
-                if self.digest(new_file) == self.digest(previous_ref_filepath):
-                    print("New gene reference is identical to previous gene reference", gene_info_description)
-                    os.remove(new_file)
-                else:
-                    print("generating gene reference diff for", gene_info_description)
-                    self.generate_gene_ref_diff(gene_info_description, new_file, previous_ref_filepath)
+                # Rename new_genes_{filename} to genes_{filename} if this is the first time we're introducing a new gene file type
```
joyceyan, line 256: the way the script is currently written is that it expects you to have a previous version to reference. i've updated it to account for if we're introducing a new gencode file type for the first time by:
- renaming `new_genes_fruit_fly.csv.gz` to `genes_fruit_fly.csv.gz` and skipping over the check against the previous version, since there is no previous version to reference
- generating a `fruit_fly_diff.txt` file which includes all of the new genes we're introducing

cellxgene_schema_cli/scripts/gene_processing.py
```diff
@@ -248,13 +253,29 @@ def process_gene_infos(self, gene_infos: dict) -> None:
...
+                if not os.path.exists(previous_ref_filepath):
                     os.replace(new_file, previous_ref_filepath)
+
+                    new_ref_gene_ids = set()
```
nayib-jose-gloria, line 260: the gencode *_diff files are a list of which genes have been deprecated in between versions (see `generate_gene_ref_diff`), for use by the migrate script. So in the case of first-time gencode files, we should just ignore generating it, since there is nothing to deprecate

nayib-jose-gloria (commented):
let's not generate a diff.txt for new gencode files--the rest of the changes look good!

cellxgene_schema_cli/cellxgene_schema/gencode_files/gene_info.yml
```diff
@@ -13,3 +13,6 @@ mouse:
 sars_cov_2:
   description: sars_cov_2
   url: ftp://ftp.ensemblgenomes.org/pub/viruses/gtf/sars_cov_2/Sars_cov_2.ASM985889v3.101.gtf.gz
+fruit_fly:
+  description: fruit_fly
```
nayib-jose-gloria, line 17: nit: we use the scientific name for the description for everything else--let's use 'drosophila_melanogaster'

nayib-jose-gloria (commented):
there's some corresponding changes required in the validator to use a new gencode file--see [here](https://github.com/chanzuckerberg/single-cell-curation/blob/main/cellxgene_schema_cli/cellxgene_schema/gencode.py). Feel free to suggest refactors as well if adding new organism types is too onerous as-is. Also feel free to tackle in a separate PR if you prefer

nayib-jose-gloria (commented):
re: testing--for this one, I think we can rely on updating unit tests, like this [one](https://github.com/chanzuckerberg/single-cell-curation/blob/cb0c19ca08ccdee11e98769e775a7f5113c82300/cellxgene_schema_cli/tests/test_validate.py#L152) and probably other similar ones.

and then let's also set-up the experimental validator branch to run unit tests on prs

.github/workflows/gencode-process.yml
```diff
@@ -37,7 +37,9 @@ jobs:
         run: |
           make gene-processing -C cellxgene_schema_cli
           git add ./cellxgene_schema_cli/cellxgene_schema/gencode_files/genes_*.csv.gz
-          git add ./cellxgene_schema_cli/cellxgene_schema/gencode_files/*_diff.txt
+          ls ./cellxgene_schema_cli/cellxgene_schema/gencode_files/*_diff.txt 1> /dev/null 2>&1 && \
+          git add ./cellxgene_schema_cli/cellxgene_schema/gencode_files/*_diff.txt || \
+          echo "No matching files found. Skipping git add for diff files."
```
joyceyan, line 42: this should fix the issue of not having diff files if we have a new gene type

cellxgene_schema_cli/tests/fixtures/examples_validate.py
```diff
@@ -299,6 +299,19 @@
...
+        [False],
+        [False],
+        [False],
+    ],
+    index=["ERCC-00002", "ENSG00000127603", "ENSMUSG00000059552", "ENSSASG00005000004", "FBtr0472816_df_nrg"],
+    columns=["feature_is_filtered"],
+)
+
+# Valid var per schema
+good_var_four_genes = pd.DataFrame(
```
joyceyan, line 314: i added the fruit fly gene to `good_var`, and kept what we previously had as `good_var_four_genes` so that we wouldn't need to rewrite the matrix math in `test_determine_seurat_convertibility`

cellxgene_schema_cli/cellxgene_schema/write_labels.py
```diff
@@ -218,7 +218,7 @@ def _get_mapping_dict_feature_biotype(self, ids: List[str]) -> Dict[str, str]:
         for i in ids:
             if i.startswith("ERCC"):
                 mapping_dict[i] = "spike-in"
-            elif i.startswith("ENS"):
+            elif i.startswith("ENS") or i.startswith("FB"):
```
joyceyan, line 221: right now we have to explicitly state which gene prefixes we allow here, so this is another place we'd have to alter when we add new gene types. i'm neutral on keeping this as is, but we could also change this to

```
if i.startswith("ERCC"):
  mapping_dict[i] = "spike-in"
else:
  mapping_dict[i] = "gene"
```

to make it easier for when we add new gene types, if we can safely assume all future gene types will be "genes"

nayib-jose-gloria: I think switching to your suggested change makes sense; it would also be an easy enough change if we do get different feature_biotype requirements in the future

cellxgene_schema_cli/tests/fixtures/examples_validate.py
nayib-jose-gloria: do you mind leaving a comment linking this to seurat convertibility testing? so we know we can safely delete if we decide to merge this feature branch into main post seurat deprecation

cellxgene_schema_cli/tests/fixtures/examples_validate.py
```diff
@@ -316,9 +329,10 @@
         ["spike-in", False, "ERCC-00002 (spike-in control)", "NCBITaxon:32630", 1061, "synthetic"],
         ["gene", False, "MACF1", "NCBITaxon:9606", 2821, "protein_coding"],
         ["gene", False, "Trp53", "NCBITaxon:10090", 1797, "protein_coding"],
-        ["gene", False, "S", "NCBITaxon:2697049", 3822, "protein_coding"],
+        ["gene", False, "S_ENSSASG00005000004", "NCBITaxon:2697049", 3822, "protein_coding"],
```
nayib-jose-gloria, line 332: Why did feature_name change for this one if we didn't bump the version?

cellxgene_schema_cli/tests/fixtures/examples_validate.py
```diff
@@ -316,9 +329,10 @@
...
+        ["gene", False, "FBtr0472816_df_nrg", "NCBITaxon:7227", 22, ""],
```
nayib-jose-gloria, line 333: why is the feature_type an empty string for this one?

nayib-jose-gloria (changes requested):
one small fix required, and a few nits

cellxgene_schema_cli/tests/fixtures/examples_validate.py
joyceyan: so here: https://github.com/chanzuckerberg/single-cell-curation/blob/main/cellxgene_schema_cli/scripts/gene_processing.py#L211-L224

we deduplicate gene names across species. my guess is that fruit fly has a gene named `S`, which resulted in this being picked up and renamed to `[gene name]_[gene id]` instead of just `gene name`.

cellxgene_schema_cli/tests/test_validate.py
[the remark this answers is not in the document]
joyceyan: wait is `test_get_dictionary_mapping_feature_type` not covering that?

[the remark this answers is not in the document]
nayib-jose-gloria: there's [feature_type](https://github.com/chanzuckerberg/single-cell-curation/blob/main/schema/5.2.0/schema.md#feature_type) and there's also [feature_biotype](https://github.com/chanzuckerberg/single-cell-curation/blob/main/schema/5.2.0/schema.md#feature_biotype)

cellxgene_schema_cli/tests/fixtures/examples_validate.py
joyceyan: i started a discussion here to confirm we can generalize the logic a bit: https://czi-sci.slack.com/archives/C02TSE92DBM/p1729886608538879

nayib-jose-gloria (approved):

PR #1063, opened 2024-10-24
feat: require organism_cell_type_ontology_term_id column + support zebrafish validation

nayib-jose-gloria:
## Reason for Change

- #1049
- See schema definition [here](https://github.com/chanzuckerberg/single-cell-curation/blob/q4-2024/experimental-validator/schema/drafts/5.2.1-experimental.md#organism_cell_type_ontology_term_id)

## Changes

- add required organism_cell_type_ontology_term_id column in obs for all datasets
- Value should be 'NA' for all organisms but zebrafish (for now)
- added validation rules for when organism ontology term id is Zebrafish
- update _validate_spatial_cell_type_ontology_term_id for dataset rows with visium assay--now, both cell_type AND organism_cell_type should be "unknown" if in_tissue is 0 and assay is Visium. 
             - Note: while not explicit in the schema, I chose to also validate 'na' as an accepted value for organism_cell_type in this case, as the schema says it must be 'na' when organism_ontology_term_id is an organism not listed in the 'organism_cell_type_ontology_term_id' definition (i.e. Human/Mouse).  
              - While we could query for `(organism_cell_type_ontology_term_id == "unknown" ) or ("organism_cell_type_ontology_term_id==na" AND 'organism_ontology_term_id' is human/mouse/etc)`, There is separate validation that already ensures 'na' is only used in that scenario, so I didn't see a need to validate it again as part of this visium check. Let's double check with product + users if this makes sense before merging.

## Testing

- unit tests + some clean-up in TestZebrafish class
- updated the example anndata fixture to have new column

## Notes for Reviewer


cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -1542,10 +1542,14 @@ def _validate_spatial_cell_type_ontology_term_id(self):
         if (
             (self.adata.obs["assay_ontology_term_id"] == ASSAY_VISIUM)
             & (self.adata.obs["in_tissue"] == 0)
-            & (self.adata.obs["cell_type_ontology_term_id"] != "unknown")
+            & (
+                (self.adata.obs["cell_type_ontology_term_id"] != "unknown")
+                | (~self.adata.obs["organism_cell_type_ontology_term_id"].isin(["unknown", "na"]))
```
joyceyan, line 1547: can you update the code comment above this?

cellxgene_schema_cli/tests/test_schema_compliance.py
```diff
@@ -2558,13 +2582,76 @@ def test_development_stage_ontology_term_id_zebrafish__invalid(
...
-        obs.loc[
-            obs.index[0],
-            "self_reported_ethnicity_ontology_term_id",
-        ] = "na"
         validator.validate_adata()
         assert validator.errors == [error + " " + zebrafish_error_message_suffix]
+
+    @pytest.mark.parametrize(
+        "organism_cell_type_ontology_term_id",
+        ["ZFA:0000003", "unknown"],
```
joyceyan, line 2593: can you also add `na` as a valid test case?

joyceyan (approved):

brianraymor:
> update _validate_spatial_cell_type_ontology_term_id for dataset rows with visium assay--now, both cell_type AND organism_cell_type should be "unknown" if in_tissue is 0 and assay is Visium.
> - Note: while not explicit in the schema, I chose to also validate 'na' as an accepted value for organism_cell_type in this case, as the schema says it must be 'na' when organism_ontology_term_id is an organism not listed in the 'organism_cell_type_ontology_term_id' definition (i.e. Human/Mouse).

I should reorganize and clarify the schema.

1. If organism is not zebrafish or fly, then the value must be "na". Break.
2. If zebrafish or fly, then define the "unknown" cases. 
3. The value should never be "na" for zebrafish or fly. 

"unknown" is not equivalent to "na".

nayib-jose-gloria:
> > update _validate_spatial_cell_type_ontology_term_id for dataset rows with visium assay--now, both cell_type AND organism_cell_type should be "unknown" if in_tissue is 0 and assay is Visium.
> > 
> > * Note: while not explicit in the schema, I chose to also validate 'na' as an accepted value for organism_cell_type in this case, as the schema says it must be 'na' when organism_ontology_term_id is an organism not listed in the 'organism_cell_type_ontology_term_id' definition (i.e. Human/Mouse).
> 
> I should reorganize and clarify the schema.
> 
> 1. If organism is not zebrafish or fly, then the value must be "na". Break.
> 2. If zebrafish or fly, then define the "unknown" cases.
> 3. The value should never be "na" for zebrafish or fly.
> 
> "unknown" is not equivalent to "na".

Thanks for the clarified schema, Brian! 

It looks like we're on the same page re: this change--if visium AND in_tissue is 0 AND zebrafish or fly --> organism_cell_type is "unknown"
if NOT zebrafish or fly --> organism_cell_type is "na", including when visium AND in_tissue is 0

cellxgene_schema_cli/tests/test_schema_compliance.py
```diff
@@ -2558,13 +2582,94 @@ def test_development_stage_ontology_term_id_zebrafish__invalid(
...
```
joyceyan, line 2593: can you also add a non-zebrafish organism + `na` as a valid case?

cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -1528,7 +1528,11 @@ def _validate_spatial_assay_ontology_term_id(self):
 
     def _validate_spatial_cell_type_ontology_term_id(self):
         """
-        Validate cell type ontology term id is "unknown" if Visium, is_single is True and in_tissue is 0.
+        if dataset row obs.assay is Visium, uns.spatial.is_single is True, and obs.in_tissue is 0:
+
+        Validate cell type ontology term id is "unknown"
+        and organism cell type ontology term id is "unknown" (or "na", for when this field is not applicable
```
joyceyan, line 1534: i think it would be a bit more clear to have a different error message if you have a non-zebrafish/fly organism and this field is not `na`. maybe something like `For organism <organism name>, organism_cell_type_ontology_term_id must be 'na'`. Otherwise I feel like the error message is a bit confusing

joyceyan (changes requested):

cellxgene_schema_cli/tests/test_schema_compliance.py
nayib-jose-gloria: that case is covered in test_valid_anndata, since the example anndata already has non-zebrafish organisms + na

joyceyan (approved):

PR #1079, opened 2024-11-01
feat: map_species command for organism specific ontology terms to CL/UBERON mappings

nayib-jose-gloria:
## Reason for Change

- https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/single-cell-curation/1071

## Changes

- CLI function to map cell_type_ontology_term_id and tissue_ontology_term_id in anndata obs based on values in organism_*_ontology_term_id 
- if multiple matching values possible, will report accordingly in a log statement and not assign a value.
- if no matching values found, will report accordingly in a log warning

## Testing

- unit tests


joyceyan (approved):
if we actually want to release this, we'd probably want to add some documentation. but seems fine to add that later

PR #1115, opened 2024-11-18
feat: update "is visium" definition for cell_type_ontology_term_id

joyceyan:
## Reason for Change

https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/single-cell-curation/1103

## Changes

updates the definition of "is visium" to be "descendant of `EFO:0010961`" instead of "is `EFO:0010961`" for cell_type_ontology_term_id. this PR also introduces a new function, `_is_visium_including_descendants` which we can use for subsequent visium tickets.

i initially had tried updating the global `_is_visium` check to use the logic in `_is_visium_including_descendants` but found that it resulted in some additional things breaking. to make all the visium work be more parallelizable and test changes isolated, i think it makes more sense to just migrate each individual field to use `_is_visium_including_descendants` and then eventually remove `_is_visium` once everything has been migrated to use the descendant logic.

## Testing

- updated and wrote unit tests


Bento007 (approved):

cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -1760,6 +1760,37 @@ def _is_visium(self) -> bool:
...
+
+            if assay_ontology_term_id is not None:
+                # Convert to a regular Series if it's Categorical
+                assay_ontology_term_id = pd.Series(assay_ontology_term_id)
+
+                # Check if any term is a descendant of ASSAY_VISIUM
+                try:
+                    visium_results = assay_ontology_term_id.apply(
+                        lambda term: ASSAY_VISIUM
+                        in list(ONTOLOGY_PARSER.get_lowest_common_ancestors(ASSAY_VISIUM, term))
```
ejmolinelli, line 1781: @joyceyan While this implementation makes sense, is there a reason to not use ONTOLOGY_PARSER.get_term_descendants ([here](https://github.com/chanzuckerberg/cellxgene-ontology-guide/blob/main/api/python/src/cellxgene_ontology_guide/ontology_parser.py#L348C9-L348C29))?

joyceyan: @ejmolinelli There's no particular reason. They would both work, and I just picked the implementation that I felt made more sense to me.

PR #1119, opened 2024-11-20
chore: forbid any NaN in spatial embeddings #

ejmolinelli:
## Reason for Change
- #1105 
- compliance with 5.3 schema

## Changes
- add NaN check for spatial embeddings - no NaNs allowed
- convenience change: pyproject.toml - pytest block adds cellxgene_schema_cli to python path

## Testing
- added one test to test_validate
- modified existing test in test_compliance
```
```

## Notes for Reviewer
- notice the change to pyproject.toml. It's useful for vscode dev (allows me to interactively debug tests) and it shouldn't interfere with other environments. I'm happy to take out if it has unintended consequences.

### DIFF
--- a/cellxgene_schema_cli/tests/test_validate.py
+++ b/cellxgene_schema_cli/tests/test_validate.py
@@ -1050,8 +1050,7 @@ class TestCheckSpatial:
         validator.adata.obsm["spatial"][0, 1] = np.nan
         # Confirm spatial is valid.
         validator.validate_adata()
-        assert len(validator.errors) == 1
-        assert validator.errors[0] == "ERROR: adata.obs['spatial] contains at least one NaN value."
+        assert validator.errors == ["ERROR: adata.obs['spatial] contains at least one NaN value."]
 
 
 class TestValidatorValidateDataFrame:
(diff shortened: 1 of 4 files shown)


cellxgene_schema_cli/tests/test_validate.py
```diff
@@ -1040,6 +1040,19 @@ def test__validate_cell_type_ontology_term_id_error(
...
+        validator: Validator = Validator()
+        validator._set_schema_def()
+        validator.adata = adata_visium.copy()
+        validator.visium_and_is_single_true_matrix_size = 2
+
+        # invalidate spatial embeddings with NaN value
+        validator.adata.obsm["spatial"][0, 1] = np.nan
+        # Confirm spatial is valid.
+        validator.validate_adata()
+        assert len(validator.errors) == 1
```
joyceyan, line 1053: nit: you could accomplish this in just one assert
```
assert validator == ["ERROR: adata.obs['spatial] contains at least one NaN value."]
```

ejmolinelli: yep, absolutely. done.

Issue #1104, opened 2024-11-14
cellxgene-schema CLI must update validation for obs['in_tissue']


brianraymor:
# Changelog

* obs (Cell metadata)
  * Updated the requirements for `in_tissue` to include descendants of  _Visium Spatial Gene Expression_.

# Design

### in_tissue

<table><tbody>
    <tr>
      <th>Key</th>
      <td>in_tissue</td>
    </tr>
    <tr>
      <th>Annotator</th>
      <td>Curator MUST annotate if <code>assay_ontology_term_id</code> is a descendant of <a href="https://www.ebi.ac.uk/ols4/ontologies/efo/classes?obo_id=EFO%3A0010961"><code>"EFO:0010961"</code></a> for <i>Visium Spatial Gene Expression</i> and <code>uns['spatial']['is_single']</code> is <code>True</code>; otherwise, this key MUST NOT be present.</td>
    </tr>
    <tr>
      <th>Value</th>
        <td><code>int</code>. This MUST be the value for the corresponding spot from the <code>in_tissue</code> field in <code>tissue_positions_list.csv</code> or <code>tissue_positions.csv</code> which is either <code>0</code> if the spot falls outside tissue or <code>1</code> if the spot falls inside tissue. See <a href="https://www.10xgenomics.com/support/software/space-ranger/analysis/outputs/spatial-outputs">Space Ranger Spatial Outputs</a>.
        </td>
    </tr>
</tbody></table>
<br>

PR #1124, opened 2024-11-21
feat: update validation for obs['in_tissue'] to include descendants of Visiium

ejmolinelli:
## Reason for Change

- #1104 

## Changes
- update `_validate_spatial_tissue_position` to support descendants of Visium
- refactor `_is_visium_including_descendants` to work around non-idempotent dependence on ONTOLOGY_PARSER [COG#246](https://github.com/chanzuckerberg/cellxgene-ontology-guide/issues/246)
- 

## Testing
- `test_schema_compliance.test_column_presence_in_tissue` tests descendants and non descendants
```
```

## Notes for Reviewer
- the implementation for checking descendants relies on a new utility method. This was just a way to have a single method implementation that can be used as an `apply()` function and be cached to avoid unnecessary graph traversals in ontology
-

### DIFF
--- a/cellxgene_schema_cli/cellxgene_schema/utils.py
+++ b/cellxgene_schema_cli/cellxgene_schema/utils.py
@@ -15,8 +15,6 @@ logger = logging.getLogger(__name__)
 
 SPARSE_MATRIX_TYPES = {"csc", "csr", "coo"}
 
-ONTOLOGY_PARSER = OntologyParser(schema_version="v5.3.0")
-
 
 def replace_ontology_term(dataframe, ontology_name, update_map):
     column_name = f"{ontology_name}_ontology_term_id"
@@ -158,7 +156,7 @@ def get_hash_digest_column(dataframe):
 
 
 @lru_cache()
-def is_ontological_descendant_of(term: str, target: str, include_self: bool = True) -> bool:
+def is_ontological_descendant_of(onto:OntologyParser, term: str, target: str, include_self: bool = True) -> bool:
     """
     Determines if :term is an ontological descendant of :target and whether to include :term==:target.
 
@@ -166,4 +164,4 @@ def is_ontological_descendant_of(term: str, target: str, include_self: bool = Tr
 
     #TODO:[EM] needs testing
     """
-    return term in set(ONTOLOGY_PARSER.get_term_descendants(target, include_self))
+    return term in set(onto.get_term_descendants(target, include_self))
--- a/cellxgene_schema_cli/cellxgene_schema/utils.py
+++ b/cellxgene_schema_cli/cellxgene_schema/utils.py
@@ -2,21 +2,21 @@ import logging
 import os
 import sys
 from base64 import b85encode
-from typing import Dict, List, Union
 from functools import lru_cache
+from typing import Dict, List, Union
 
 import anndata as ad
 import numpy as np
+from cellxgene_ontology_guide.ontology_parser import OntologyParser
 from scipy import sparse
 from xxhash import xxh3_64_intdigest
-
-from cellxgene_ontology_guide.ontology_parser import OntologyParser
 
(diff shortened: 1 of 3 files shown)


cellxgene_schema_cli/cellxgene_schema/utils.py
```diff
@@ -2,17 +2,21 @@
...
 import numpy as np
+from cellxgene_ontology_guide.ontology_parser import OntologyParser
 from scipy import sparse
 from xxhash import xxh3_64_intdigest
 
 logger = logging.getLogger(__name__)
 
 SPARSE_MATRIX_TYPES = {"csc", "csr", "coo"}
 
+ONTOLOGY_PARSER = OntologyParser(schema_version="v5.3.0")
```
joyceyan, line 18: i would prefer passing in the `ONTOLOGY_PARSER` that we create an instance of in `validate.py` into `is_ontological_descendant_of`, that way we only have one place to update when we bump the schema version

cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -1477,18 +1477,25 @@ def _validate_spatial_cell_type_ontology_term_id(self):
 
         :rtype none
         """
-        # Exit if:
-        # - not Visium and is_single is True as no further checks are necessary
-        # - in_tissue is not specified as checks are dependent on this value
-        if not self._is_visium_including_descendants() and self._is_single() or "in_tissue" not in self.adata.obs:
+        self._is_visium_including_descendants()
```
joyceyan, line 1480: it might make sense to just put all of these calls further up the validator, and then rely on `self.is_visium` or `self.is_spatial` states directly instead of needing to call these throughout. feel free to refactor this separately though

cellxgene_schema_cli/cellxgene_schema/utils.py
ejmolinelli: yes, I see your point. I definitely agree we should minimize (to 1) the number of places a bump needs to be reflected. I'll do that.

One question I have is, why is the version in the code at all and not as a configuration option?

cellxgene_schema_cli/cellxgene_schema/validate.py
ejmolinelli: I agree. However, I don't see a performance issue given that it is only computed if the state is not already set.

joyceyan: yeah i think it's fine, it's just slightly confusing to follow. but honestly this entire codebase is so 🤷‍♀️

cellxgene_schema_cli/cellxgene_schema/utils.py
joyceyan: this was previously set to something like `get_current_schema_version()`. however, the current schema version is 5.2.0, because we bump the version number as part of the automated release process. and in the past, we would usually bump the ontology versions in COG completing all the feature changes to the validator. there's some more context in [this slack thread](https://czi-sci.slack.com/archives/C07AV4NU9D2/p1732047031943819) but basically i decided hard-coding this was easier than needing to change some of our automated release process to make sure we're not double-bumping the version number before creating the new 5.3.0 release.

cellxgene_schema_cli/cellxgene_schema/validate.py
ejmolinelli: I've created a "#tech" issue for this here: #1125

cellxgene_schema_cli/cellxgene_schema/utils.py
ejmolinelli: I've followed your suggestion. Good looking out @joyceyan

joyceyan (approved):

Issue #1128, opened 2024-11-22
experimental cellxgene-schema CLI must support C. elegans


brianraymor:
## Design

See [schema 5.2.1](https://raw.githubusercontent.com/chanzuckerberg/single-cell-curation/refs/heads/main/schema/drafts/5.2.1-experimental.md). This update adds `sex_ontology_term_id` due to specific requirements for *Caenorhabditis elegans*.

### schema v5.2.1-experimental

* Required Ontologies - addressed by [_ontology guide must add C. elegans ontologies_](https://github.com/chanzuckerberg/cellxgene-ontology-guide/issues/250)
  * Added C. elegans Development Ontology (WBls) release 2024-09-26 Wormbase WS295
  * Added C. elegans Gross Anatomy Ontology (WBbt) release 2024-09-24 Wormbase WS295
* Required Gene Annotations - addressed by [_experimental cellxgene-schema CLI must update gene references to Ensembl 113 release_](https://github.com/chanzuckerberg/single-cell-curation/issues/1127)
  * Added *Caenorhabditis elegans* WBcel235 (GCA_000002985.3) Ensembl 113
  * Added *Danio rerio* GRCz11 (GCA_000002035.4) Ensembl 113
  * Added *Drosophila melanogaster* BDGP6.46 (GCA_000001215.4) Ensembl 113
* obs (Cell metadata)
  * Updated `development_stage_ontology_term_id` to include:
    * *Caenorhabditis elegans*
  * Added `organism_cell_type_ontology_term_id`
  * Updated `organism_ontology_term_id` to require all observations to contain the same value for:
    * *Caenorhabditis elegans*
  * Added `organism_tissue_ontology_term_id`
  * Updated `sex_ontology_term_id` for *Caenorhabditis elegans*
  * Updated `tissue_type` to require `"tissue"` for:
    * *Caenorhabditis elegans*
* var and raw.var (Gene Metadata)
  * Updated `feature_reference` to include:
    * *Caenorhabditis elegans*

PR #1131, opened 2024-11-26
feat: add support for validating c. elegans

nayib-jose-gloria:
## Reason for Change

- #1128 (work for gencode / feature_reference done in separate PR)
- see corresponding c. elegans schema reqs in https://github.com/chanzuckerberg/single-cell-curation/blob/main/schema/drafts/5.2.1-experimental.md

## Changes

- add c. elegans validation rules for:
     - development stage ontology term ID
     - sex ontology term ID
     - organism_cell_type ontology term ID
     - organism_tissue_type ontology term ID
     - tissue_type
- COG ontology bump to use experimental COG release with c. elegans ontologies (1.3.2a0)
- bump CLI version to 5.2.1a2

## Testing

- unit tests (no real c. elegans examples yet to test against)

## Notes for Reviewer

### DIFF
--- a/cellxgene_schema_cli/cellxgene_schema/schema_definitions/schema_definition.yaml
+++ b/cellxgene_schema_cli/cellxgene_schema/schema_definitions/schema_definition.yaml
@@ -283,9 +283,6 @@ components:
                 ancestors:
                   WBbt:
                     - WBbt:0005766
-              forbidden:
-                terms:
-                  - WBbt:0005766
         # if organism is none of the above
         error_message_suffix: >-
           When 'organism_ontology_term_id'-specific requirements are not defined in the schema definition,
@@ -505,9 +502,6 @@ components:
                 ancestors:
                   WBls:
                     - WBls:0000075
-              forbidden:
-                terms:
-                  - WBls:0000075
               exceptions:
                 - unknown
         # If organism is none of the above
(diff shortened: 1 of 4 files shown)


cellxgene_schema_cli/cellxgene_schema/schema_definitions/schema_definition.yaml
```diff
@@ -254,6 +269,23 @@ components:
...
+            curie_constraints:
+              ontologies:
+                - WBbt
+              allowed:
+                ancestors:
+                  WBbt:
+                    - WBbt:0005766
+              forbidden:
+                terms:
+                  - WBbt:0005766
```
joyceyan, line 288: just curious, do you need to explicitly forbid `WBbt:0005766` if you only allow descendant terms of `WBbt:0005766` on lines 282-285? i had thought the `allowed` block would implicitly forbid the `WBbt:0005766` term, but now i'm not sure and it is different from how the rule for `organism_cell_type_ontology_term_id` is written

joyceyan (approved):
TIL female roundworms don't exist!

cellxgene_schema_cli/cellxgene_schema/schema_definitions/schema_definition.yaml
nayib-jose-gloria: good catch, I didn't need to add that! For some reason I thought that's what we did, but I was misremembering. fixing!

Issue #1134, opened 2024-11-28
bug: categorical encoding in obs['assay_ontology_term_id']


ejmolinelli:
## Motivation

We use the invert operand `~` to invert the boolean value of fields during validation. One place in particular that this happens is in checking if the `assay_ontology_term_id` is or is not an inclusive descendant of Visium Gene Expression. However, the assay_ontology_term_id is interpreted as a pandas `categorical` dtype, which does not work well with inversion (or other unary) operands.  

This results in failed integration tests done by the lattice team.

## Definition of Done
The task is done when (a) test exists that captures the reported error before a code change is made, and (b) passes after the change is made.

## Tasks
- [x]  in the offending areas, make sure the code works for both string encoded and categorical dtype. 
- [x] add test

## Other relevant information
- This error was raised by an external partner, and not caught by our existing test suite.

PR #1142, opened 2024-12-03
fix: handle both string and category encoded "assay_ontology_term_id"

ejmolinelli:
## Reason for Change
- #1134 

## Changes
- in cases where a unary operation is applied to "category" dtype column, be sure to cast the result as boolean

## Testing
- a single test was added specifically for the 'assay_ontology_term_id' obs field checking that it works when encoded as a normal string or as a category.

## Notes for Reviewer
- This should not introduce any new ambiguity, however, it does not remove existing ambiguity elsewhere in the codebase whereby some implicit assumptions about column encoding may be violated.

### DIFF
--- a/cellxgene_schema_cli/tests/test_schema_compliance.py
+++ b/cellxgene_schema_cli/tests/test_schema_compliance.py
@@ -633,14 +633,20 @@ class TestObs:
         ]
         assert validator.errors == [self.get_format_error_message(error_message_suffix, error)]
 
-    def test_assay_ontology_term_id__is_categorical(self, validator_with_visium_assay):
+    def test_assay_ontology_term_id__as_categorical(self, validator_with_visium_assay):
         """
         Formally, assay_ontology_term_id is expected to be a categorical variable of type string. However, it should work for categorical dtypes as well.
         """
         validator: Validator = validator_with_visium_assay
+        
+        # check encoding as string
+        validator._validate_obsm()
+        assert validator.errors == []
+        validator.reset()
+        
         # force encoding as 'categorical'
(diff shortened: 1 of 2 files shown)


cellxgene_schema_cli/tests/test_schema_compliance.py
```diff
@@ -632,6 +633,22 @@ def test_assay_ontology_term_id(self, validator_with_adata, assay_ontology_term_
...
         assert validator.errors == [self.get_format_error_message(error_message_suffix, error)]
 
+    def test_assay_ontology_term_id__as_categorical(self, validator_with_visium_assay):
+        """
+        Formally, assay_ontology_term_id is expected to be a categorical variable of type string. However, it should work for categorical dtypes as well.
+        """
+        validator: Validator = validator_with_visium_assay
+
+        # check encoding as string
+        validator._validate_obsm()
```
Bento007, line 643: this does not exercise your changes. `validator._check_spatial` would be better. Alternatively you can call `validator._validate_spatial_cell_type_ontology_term_id`, and `validator._validate_spatial_tissue_position` to reduce the test area.

Bento007 (changes requested):
I don't think your test covers the changes made

cellxgene_schema_cli/tests/test_schema_compliance.py
ejmolinelli: thanks @Bento007 good catch. This was a tab-complete oversight on my part. Updated to `check_spatial`.

Bento007 (approved):

Issue #1100, opened 2024-11-14
cellxgene-schema CLI must update validation for obs['array_col']


brianraymor:
# Changelog

* obs (Cell metadata)
  * Updated  the requirements for `array_col`:
    *  MUST be annotated if the `assay_ontology_term_id` is a descendant of _Visium Spatial Gene Expression_
    * Added ranges for _Visium CytAssist Spatial Gene Expression, 6.5mm_ and _Visium CytAssist Spatial Gene Expression, 11mm_ 

# Design

### array_col

<table><tbody>
    <tr>
      <th>Key</th>
      <td>array_col</td>
    </tr>
    <tr>
      <th>Annotator</th>
         <td>Curator MUST annotate if <code>assay_ontology_term_id</code> is  a descendant of <a href="https://www.ebi.ac.uk/ols4/ontologies/efo/classes?obo_id=EFO%3A0010961"><code>"EFO:0010961"</code></a> for <i>Visium Spatial Gene Expression</i> and <code>uns['spatial']['is_single']</code> is <code>True</code>; otherwise, this key MUST NOT be present.</td>
    </tr>
    <tr>
      <th>Value</th>
        <td><code>int</code>. This MUST be the value of the column coordinate for the corresponding spot from the <code>array_col</code> field in <code>tissue_positions_list.csv</code> or <code>tissue_positions.csv</code>. See <a href="https://www.10xgenomics.com/support/software/space-ranger/analysis/outputs/spatial-outputs">Space Ranger Spatial Outputs</a>.<br>
        <br>
          <table>
          <thead>
          <tr>
          <th>For Assay</th>
          <th>Value MUST be in<br>the range between</th>
          </tr>
          </thead>
          <tbody>
            <tr>
              <td><i>Visium Spatial Gene Expression V1</i> [<a href="https://www.ebi.ac.uk/ols4/ontologies/efo/classes?obo_id=EFO%3A0022857"><code>EFO:0022857</code></a>]</td>
              <td><code>0</code> and <code>127</code></td>
           </tr> 
            <tr>
              <td><i>Visium CytAssist Spatial Gene Expression, 6.5mm</i> [<a href="https://www.ebi.ac.uk/ols4/ontologies/efo/classes?obo_id=EFO%3A0022859"><code>EFO:0022859</code></a>]</td>
              <td><code>0</code> and <code>127</code></td>
           </tr>
            <tr>
[...]

PR #1143, opened 2024-12-03
feat: differential tissue position row/col max sizes for visium and visium 11

ejmolinelli:
## Reason for Change

- #1100 
- #1101 
- 5.3 schema compliance

## Changes
- validator returns appropriate tissue position max values for visium and visium 11, respectively

## Testing
- unit-tests cover min and max checks for visium and visium 11 assays

## Notes for Reviewer

### DIFF
--- a/cellxgene_schema_cli/cellxgene_schema/validate.py
+++ b/cellxgene_schema_cli/cellxgene_schema/validate.py
@@ -4,7 +4,7 @@ import numbers
 import os
 import re
 from datetime import datetime
-from typing import Dict, List, Mapping, Optional, Union
+from typing import Dict, List, Mapping, Optional, Tuple, Union
 
 import anndata
 import matplotlib.colors as mcolors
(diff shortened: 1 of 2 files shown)


joyceyan (approved):

cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -29,6 +29,8 @@
 
 VISIUM_AND_IS_SINGLE_TRUE_MATRIX_SIZE = 4992
 VISIUM_11MM_AND_IS_SINGLE_TRUE_MATRIX_SIZE = 14336
+VISIUM_TISSUE_POSITION_MAX = (77, 127)  # (row,col)
+VISIUM_11MM_TISSUE_POSITION_MAX = (127, 223)  # (row,col)
```
joyceyan, line 33: i would slightly prefer storing these as 4 separate variables, that way the reader isn't expected to understand the ordering of row/col in the tuple. but not a strong preference, so this isn't blocking

ejmolinelli: fair enough.

PR #1147, opened 2024-12-05
chore: add back return value from is_seurat_convertible

joyceyan:
## Reason for Change

makes the change to remove `is_seurat_convertible` more backwards compatible friendly by continuing to return `False`. we can remove this return value entirely with a future release, though i think it should technically be part of a major update if we do

## Changes

- returns `False` as the final tuple value always

## Testing

updated unit tests

## Notes for Reviewer


ejmolinelli (approved):
lgtm

I don't see anything missing.

PR #1152, opened 2024-12-11
feat: update to anndata 0.11 and memory efficient reads + writes

nayib-jose-gloria:
## Reason for Change

- https://czi.atlassian.net/browse/VC-1334

## Changes

- modify Isaac/Emanuele's alternate [backed h5ad read](https://github.com/chanzuckerberg/single-cell-curation/pull/1021/files#diff-4d93a63df86a2cbba75dbe4c025f4070481e8610948f60349351bb1f10b35abeR119) approach for a more memory-efficient h5ad read. Same concept, but returns matrix layers as dask arrays for chunked, memory-efficient writes.
- use single-threaded dask cluster for memory efficient h5ad writes
- update `_validate_raw_data_with_in_tissue_0` to traverse matrix in chunks rather than reading into memory, since the visium matrix size is no longer static 
- Make `count_matrix_nonzero` staticmethod so it may be re-used in data-portal, which currently redundantly implements its own matrix chunking + nonzero counter function. 
- update `count_matrix_nonzero` to count nonzeros among a subset of columns; used by `_validate_column_feature_is_filtered`. Refactor was done because the previous implementation in `_validate_column_feature_is_filtered` reads the entire matrix into memory rather than leveraging the chunked reads + nonzero counts in the already existing `count_matrix_nonzero`. 
- update `get_matrix_format` and simpify arg list
           - doesn't need adata passed in, as we don't support matrices with 0 cols or rows. Any such cases would be caught 
           by other validation.
           - now accounts for passing in dask arrays
- update 'chunk_matrix' to account for dask arrays 
- remove TODOs for re-evaluating anndata mixed type checks; anndata does not plan on supporting mixed types so the checks must stay
- remove chunk_matrix custom chunking function, and use dask chunking (map_blocks)
         - NOTE: using map_blocks when there is only 1 chunk is significantly less performant, so I added a check for number of chunks before invoking.

## Testing

- Tested with 20 GB H5AD that currently requires >20GB memory allocated for anndata processing, was able to process in comparable time on my machine using only 2.5 GB of memory.
- Also tested in rdev environment pointing to this branch CLI version. Was able to DownloadValidate the 20 GB H5AD in a 1 VCPU 8GB memory machine, when it was previously being allocated an 8 VCPU 64 GB machine, with minimal impact to speed.

## Notes for Reviewer


cellxgene_schema_cli/cellxgene_schema/utils.py
```diff
@@ -68,7 +71,7 @@ def remap_deprecated_features(*, adata: ad.AnnData, remapped_features: Dict[str,
     return adata
 
 
-def get_matrix_format(adata: ad.AnnData, matrix: Union[np.ndarray, sparse.spmatrix]) -> str:
+def get_matrix_format(matrix: Union[np.ndarray, sparse.spmatrix, DaskArray]) -> str:
```
Bento007, line 74: Do `np.ndarray, sparse.spmatrix,` still need to be supported?

cellxgene_schema_cli/cellxgene_schema/utils.py
```diff
@@ -116,6 +118,27 @@ def getattr_anndata(adata: ad.AnnData, attr: str = None):
         return getattr(adata, attr)
 
 
+def read_backed(f):
```
Bento007, line 121: typing?

Bento007: Add a description maybe some link.

cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -450,8 +447,8 @@ def _validate_feature_id(self, feature_id: str, df_name: str):
         return
 
     @staticmethod
-    def _chunk_matrix(
-        matrix: Union[np.ndarray, sparse.spmatrix],
+    def chunk_matrix(
+        matrix: Union[np.ndarray, sparse.spmatrix, DaskArray],
```
Bento007, line 451: Are there any cases where we use `chunk_matrix` that isn't a DaskArray. I think `read_backed` does a good job of converting that uses it a dask array. If it doesn't, then that element likely should be read into a dask array.

TLD: I think we can remove support for `np.ndarray, sparse.spmatrix,` in `chunk_matrix`

Bento007: We can even remove `chunk_matrix` all together if we don't need to support the other data types with this function. Dask is already doing the chunking. That work could be apart of a following PR to improve processing performance.

Bento007 (changes requested):
Looks good for improving memory read efficiency. We should do a follow up to improve the compute efficiency using [dask.distributed](https://distributed.dask.org/en/latest/) on a local cluster.

cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -449,43 +449,25 @@ def _validate_feature_id(self, feature_id: str, df_name: str):
...
-        for matrix_chunk, _, _ in self._chunk_matrix(matrix):
-            nnz += matrix_chunk.count_nonzero() if matrix_format != "dense" else np.count_nonzero(matrix_chunk)
-
-        self.number_non_zero[matrix_name] = nnz
-        return nnz
+    def count_matrix_nonzero(matrix: DaskArray, filter_by_column: pd.Series = None) -> int:
+        def count_nonzeros(
+            matrix_chunk: Union[np.ndarray, sparse.spmatrix], is_sparse_matrix: bool, filter_by_column: pd.Series
+        ) -> np.array:
+            if filter_by_column is not None:
```
Bento007, line 456: nit: you should move the filtering outside of count_nonzeros. This will reduce the amount of dataset sent between jobs. and should speed things up.

cellxgene_schema_cli/tests/fixtures/examples_validate.py
```diff
@@ -487,14 +488,14 @@
 #   the unittests
 
 # Valid anndata
-adata = anndata.AnnData(X=sparse.csr_matrix(X), obs=good_obs, uns=good_uns, obsm=good_obsm, var=good_var)
+adata = anndata.AnnData(X=X, obs=good_obs, uns=good_uns, obsm=good_obsm, var=good_var)
```
Bento007, line 491: These should be copies. Don't trust my syntax bellow. 
```suggestion
adata = anndata.AnnData(X=X.copy(), obs=good_obs, uns=good_uns, obsm=good_obsm, var=good_var)
```

Bento007 (changes requested):

Bento007 (approved):

PR #1153, opened 2024-12-11
fix: donor id error logging + obsm typo

joyceyan:
## Reason for Change

- the previous error logging was printing out all of the possible categorical values for `donor_id`, which would print out all possible donor id values. not just the donor id values for the invalid rows.

## Changes

- `self.adata.obs.loc[invalid_rows, "donor_id"]` will only fetch the donor_id fields when `invalid_rows` for that index is true. so this will only print out the donor ids that are actually invalid
- i also added in a typo fix for this [line](https://github.com/chanzuckerberg/single-cell-curation/blob/60673117145bb27edd83cde3fa7fb62052e3a5b7/cellxgene_schema_cli/cellxgene_schema/validate.py#L1121) which should say `adata.obsm['spatial']` instead of `adata.obs['spatial]`

## Testing

i tested this by just printing out the error log in `test_genetic_ancestry_same_donor_id`. previously, this error message would say `ERROR: obs rows with donor ids ['donor_1', 'donor_2']`. however, this should be `ERROR: obs rows with donor ids ['donor_1']` because the way that this test is set up, only `donor_1` has invalid values

```
ERROR    cellxgene_schema.validate:validate.py:2138 ERROR: obs rows with donor ids ['donor_1'] have invalid genetic_ancestry_* values. All observations with the same donor_id must contain the same genetic_ancestry_* values. If organism_ontolology_term_id is NOT 'NCBITaxon:9606' for Homo sapiens, then all geneticancestry values MUST be float('nan'). If organism_ontolology_term_id is 'NCBITaxon:9606' for Homo sapiens, then the value MUST be a float('nan') if unavailable; otherwise, the sum of all genetic_ancestry_* fields must be equal to 1.0
```

## Notes for Reviewer

### DIFF
--- a/cellxgene_schema_cli/cellxgene_schema/validate.py
+++ b/cellxgene_schema_cli/cellxgene_schema/validate.py
@@ -1118,7 +1118,7 @@ class Validator:
 
                 # spatial embeddings can't have any NaN; other embeddings can't be all NaNs
                 if key_is_spatial and np.any(np.isnan(value)):
-                    issue_list.append("adata.obsm['spatial] contains at least one NaN value.")
+                    issue_list.append("adata.obsm['spatial'] contains at least one NaN value.")
                 elif np.all(np.isnan(value)):
                     issue_list.append(f"adata.obsm['{key}'] contains all NaN values.")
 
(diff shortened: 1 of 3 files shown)


ejmolinelli (approved):
lgtm

Issue #1022, opened 2024-09-25
Require CSR encoding in X (Matrix Layers)


brianraymor:
## `X` (Matrix Layers)

The data stored in the `X` data matrix is the data that is viewable in CELLxGENE Explorer. For `X`, `raw.X`, and in any layer, if a matrix has 50% or more values that are zeros, it MUST be encoded as a [`scipy.sparse.csr_matrix`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.csr_matrix.html) with zero values encoded as <a href="https://docs.scipy.org/doc/scipy/tutorial/sparse.html#sparse-arrays-implicit-zeros-and-duplicates">implicit zeros</a>.

CELLxGENE's matrix layer requirements are tailored to optimize data reuse. Because each assay has different characteristics, the requirements differ by assay type. In general, CELLxGENE requires submission of "raw" data suitable for computational reuse when a standard raw matrix format exists for an assay. It is STRONGLY RECOMMENDED to also include a "normalized" matrix with processed values ready for data analysis and suitable for visualization in CELLxGENE Explorer. The schema imposes the following requirements:

*   All matrix layers MUST have the same shape, and have the same cell labels and gene labels.
*  ...

---

## Context

Per conversation on the September 25 DP Call, @nayib-jose-gloria proposed:

> CSC matrices–only found 2 in the corpus, and they are memory inefficient compared to CSR/COO. How do we feel about converting existing cases and codifying in the schema that we won’t take CSC? Or putting a size limit on CSC submissions?
“Backed” mode is prohibitively slow for validating CSC, requiring us to read into memory; we allocate processing resources relative to the worst case scenario, and we are quite close to making backed reads/writes happen in constant memory for CSR/COO matrices.

Also see continuing conversation in  [**cell-sci-platform**](https://czi-sci.slack.com/archives/C06AVGFV222/p1727283001596439).

jahilton:
I would think to also extend this to .raw.X, when present.

joyceyan:
@jahilton @brianraymor My understanding is that for existing datasets with CSC, they should be converted (or in the process of converting) to CSR. For any new datasets the validator sees with CSC, the expectation is that the validator should raise a failure message rather than converting to CSR.

It also seems reasonable to me that this requirement should extend to `.raw.X`

brianraymor:
Slight refinements but otherwise, agree:

[...]

PR #1154, opened 2024-12-11
feat: enforce CSR requirement for X [wip]

joyceyan:
## Reason for Change

https://github.com/chanzuckerberg/single-cell-curation/issues/1022

## Changes

- this is very much a WIP, but here's some work i started on enforcing the CSR requirement for `X` and `raw.X`

## Testing

TODO

## Notes for Reviewer


brianraymor:
Per my comments in slack, this PR is premature since the schema has not been updated with the final requirements. It should not be linked to #1022 which is a **schema** issue not a **CLI** issue.

Issue #1165, opened 2025-01-07
BUG: scipy 1.15.0 breaking validation


Bento007:
## Description
Here in the code when slicing the X matrix with a Pandas.Series and attribute error is thrown.

https://github.com/chanzuckerberg/single-cell-curation/blob/b4b1a5f54534eed0d40135492be8a4cd3f28540c/cellxgene_schema_cli/cellxgene_schema/validate.py#L614-L614

```
../cellxgene_schema/validate.py:2092: in validate_adata
    self._deep_check()
../cellxgene_schema/validate.py:2044: in _deep_check
    self._validate_dataframe(component_name)
../cellxgene_schema/validate.py:968: in _validate_dataframe
    self._validate_column(column, column_name, df_name, column_def)
../cellxgene_schema/validate.py:684: in _validate_column
    self._validate_column_feature_is_filtered(column, column_name, df_name)
../cellxgene_schema/validate.py:614: in _validate_column_feature_is_filtered
    n_nonzero = self.adata.X[:, column].count_nonzero()
../.venv/lib/python3.10/site-packages/anndata/_core/sparse_dataset.py:414: in __getitem__
    row_sp_matrix_validated, col_sp_matrix_validated = mtx._validate_indices(
../.venv/lib/python3.10/site-packages/scipy/sparse/_index.py:269: in _validate_indices
    index.extend(ix.nonzero())
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = gene_ids-Harvard-Nuclei
ERCC-00002          True
ENSG00000127472    False
ENSG00000142798    False
ENSG00000127603    ...00198938    False
ENSG00000198840    False
ENSG00000198727    False
Name: feature_is_filtered, Length: 136, dtype: bool
name = 'nonzero'

    @final
    def __getattr__(self, name: str):
        """
        After regular attribute access, try looking up the name
        This allows simpler access to columns for interactive use.
        """
        # Note: obj.x will always call obj.__getattribute__('x') prior to
        # calling obj.__getattr__('x').
        if (
            name not in self._internal_names_set
            and name not in self._metadata
            and name not in self._accessors
            and self._info_axis._can_hold_identifiers_and_holds_name(name)
        ):
            return self[name]
>       return object.__getattribute__(self, name)
[...]

PR #1166, opened 2025-01-07
fix: scipy sparse bug

Bento007:
## Reason for Change

- https://github.com/chanzuckerberg/single-cell-curation/issues/1165

## Changes

- pin scipy to scipy<1.15 

## Testing

- unit tests


nayib-jose-gloria (approved):

nayib-jose-gloria (commented):
as you (correctly) pointed out previously, the long-tail of the 5.3 updates in main means this is going to be annoying to release as a patch. 

do you think its worth reverting main to 5.2.1 and creating a feature branch now to re-apply all the 5.3 commits since then to a feature branch? cc: @joyceyan

Bento007:
@nayib-jose-gloria, why 5.2.1? Also we can do a 5.2.3 release with the patch and merge that back into main. That should retain all of the 5.3 changes and the 5.2.3 patch.

joyceyan:
> do you think its worth reverting main to 5.2.1 and creating a feature branch now to re-apply all the 5.3 commits since then to a feature branch?

@nayib-jose-gloria do you mean reverting main to 5.2.2? i think effort wise the two choices (revert main to 5.2.2 and release this change as 5.2.3 vs release this change as a 5.2.3 patch) seem fairly equivalent to me. i would probably slightly prefer reverting main to 5.2.2 and moving the 5.3 commits to a feature branch since we don't have a timeline set for actually releasing 5.3 yet

Bento007:
These changes were made in [5.2.3](https://github.com/chanzuckerberg/single-cell-curation/releases/tag/v5.2.3)

PR #1185, opened 2025-01-22
chore: use dask in migrate script template

nayib-jose-gloria:
## Reason for Change
- use dask chunked reads and writes for anndata matrices in migrate script template, for memory efficient read/writes

## Changes

- import dask, use dask-integrated read_h5ad from cellxgene_schema.utils, and use a single-threaded dask config for chunked writes by default in the migrate script jinja template

## Testing

- unit tests


scripts/migration_assistant/migration_template.jinja
```diff
@@ -100,5 +99,5 @@ def migrate(input_file, output_file, collection_id, dataset_id):
     if DEPRECATED_FEATURE_IDS:
         dataset = utils.remove_deprecated_features(adata=dataset, deprecated=DEPRECATED_FEATURE_IDS)
 
-    dataset.write(output_file, compression="gzip")
-
+    with dask.config.set(scheduler="single-threaded"):
```
Bento007, line 102: why single-threaded?

Bento007 (approved):

PR #1205, opened 2025-01-25
fix: remove dependency on Validator in AnnDataLabelAppender

Bento007:
## Reason for Change

This is part of the multiome work, where multiple validations steps will happen in parallel. If all steps pass then we will add labels. To accomplish this the Validator needed to be decoupled from the AnnDataLabelAppender. This allows the AnnDataLabelAppender to run independently of validation.

## Changes

- Cache gene_checker in gencode.
- Replace the `Validator` parameter in `AnnDataLabelAppender` with `AnnData`
- Replace uses of `Validator` in `AnnDataLabelAppender`
- add cli to add labels to h5ad
- fix relative imports
- remove AnnDataLabelAppender.was_writing_successful

## Testing
- Updates tests

## Notes for Reviewer

### DIFF
--- a/cellxgene_schema_cli/cellxgene_schema/cli.py
+++ b/cellxgene_schema_cli/cellxgene_schema/cli.py
@@ -55,6 +55,29 @@ def schema_validate(h5ad_file, add_labels_file, ignore_labels):
 
 
 @schema_cli.command(
+    name="add-labels",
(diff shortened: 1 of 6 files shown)


nayib-jose-gloria (approved):
good stuff

Issue #1189, opened 2025-01-23
cellxgene-schema CLI must update validation for obs['assay_ontology_term_id']


brianraymor:
# Context

See [Visium Spatial Gene Expression must be prohibited in assay_ontology_term_id](https://github.com/chanzuckerberg/single-cell-curation/issues/1183).

# Design

This MUST be an EFO term and either:

* the most accurate descendant of <a href="https://www.ebi.ac.uk/ols4/ontologies/efo/classes?obo_id=EFO%3A0002772"><code>"EFO:0002772"</code></a> for <i>assay by molecule</i> excluding <a href="https://www.ebi.ac.uk/ols4/ontologies/efo/classes?obo_id=EFO%3A0010961"><code>"EFO:0010961"</code></a> for <i>Visium Spatial Gene Expression</i> while allowing its descendants

* the most accurate descendant of <a href="https://www.ebi.ac.uk/ols4/ontologies/efo/classes?obo_id=EFO%3A0010183"><code>"EFO:0010183"</code></a>  for <i>single cell library construction</i> excluding <a href="https://www.ebi.ac.uk/ols4/ontologies/efo/classes?obo_id=EFO%3A0010961"><code>"EFO:0010961"</code></a> for <i>Visium Spatial Gene Expression</i> while allowing its descendants

ejmolinelli:
would you like us to update the error messages to reflect this language? That is, instead of:

>ERROR: A dict in uns['spatial'] is required when obs['assay_ontology_term_id'] is either a descendant of 'EFO:0010961' (Visium Spatial Gene Expression) or 'EFO:0030062' (Slide-seqV2)."

We would use
> ERROR: A dict in uns['spatial'] is required when obs['assay_ontology_term_id'] is either a descendant of  'EFO:0010961' (Visium Spatial Gene Expression) excluding 'EFO:0010961' or 'EFO:0030062' (Slide-seqV2)."

PR #1214, opened 2025-01-28
feat: exclude generic visium (EFO:0010961) from allowed assay_ontology_term

ejmolinelli:
## Reason for Change
- #1189 
- 
## Changes
- exclude self when checking if assay is a descendant of EFO:0010961
- remove use of EFO:0010961 from tests where it is expected to yield a successful validation
- add explicit test for EFO:0010961 and append a custom error message for this error.

## Testing

- Most of the test updates were to replace EFO:0010961 with a lower, valid term e.g. EFO:0022857 (v1) or EFO:0022858 (v2) or EFO:0022859 (6.5mm)

```
```

## Notes for Reviewer
- Lattice confirmed that the existing error messages related to valid spatial assays is sufficient. 
- it would be nice to extend the schema validator to include visium checks - perhaps this is a maintenance task.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -5,3 +5,4 @@ types-requests
 types-setuptools
 typing_extensions
 types-PyYAML
+pre-commit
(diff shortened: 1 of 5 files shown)


requirements-dev.txt
```diff
@@ -4,4 +4,5 @@ mypy
 types-requests
 types-setuptools
 typing_extensions
-types-PyYAML
\ No newline at end of file
+types-PyYAML
+pre-commit
```
joyceyan, line 8: i'm guessing pre-commit was added because you discovered it was a dev dependency?

joyceyan (approved):

PR #1229, opened 2025-01-30
feat: add support for zebrafish, fruit fly, and roundworm

joyceyan:
## Reason for Change

https://czi.atlassian.net/browse/VC-1959

## Changes

- moves all of the logic for `organism_cell_type_ontology_term_id` and `organism_tissue_ontology_term_id` into `cell_type_ontology_term_id` and `tissue_ontology_term_id`, respectively
- there have been some changes in the logic here between Q4 and now. there are some additional forbidden terms, which i've updated.
- i did make a pretty substantial change to how we parse through these organism-specific requirements. right now, the custom schema definition we've implemented in `schema_definition.yaml` is a bit restrictive with how we write rules. it is not easy for example, to implement the rules for `tissue_ontology_term_id` without basically rewriting this because we don't properly support rules where there are two column dependencies (for `tissue_ontology_term_id`, we would need to break off logic based on `tissue_type` _and_ `organism_ontology_term_id`). the easiest way to support this was to move all the organism-specific logic into the `validate.py` script, and rely on the nice built-in ontology support in `schema_definition.yaml` to make sure that all ontologies are valid. this does result in the add-on logic in `validate.py` being somewhat janky but i think this is the cleanest way to solve this requirement without overly complicating things.

## Testing

- a lot of existing, unrelated tests broke due to change in copy in the error messages. i tried to resolve these by just asserting that there were errors, which is mostly what we're concerned with. we go through fairly extensive review with lattice for ensuring the copy in our error messages are useful, so i don't think it's necessary to add this to our unit tests as well when they ~mostly just cause annoyance when making future changes. what is primarily important is that when we make new changes, the existing failures we expect to see will still occur. the english for exactly how the failure pops up is less important.
- this PR also skips over an additional set of visium tests, `test_cell_type_ontology_term_id__visium_in_tissue_0__na` because i wasn't quite able to figure out how these tests should be re-written
- i did not do any testing to check if datasets with multiple organisms in it can be supported... that will come later

## Notes for Reviewer


ejmolinelli (commented):
Hi @joyceyan . 

Largely these seem good. The only concern relates to a design choice to simplify the schema_definition logic (independent permissions to valid prefixed values) at the expense of inserting additional dependent logic (tissues from worm organisms must be uberon or worm specific). Alternatively, keeping the rule specific checks would mean that the schema_definition.yaml is closer to the single source of truth and lowers the amount of code needed to maintain the overall validator.

If that's the design choice the team is ok with, these changes look ready to go.

joyceyan:
> Hi @joyceyan .
> 
> Largely these seem good. The only concern relates to a design choice to simplify the schema_definition logic (independent permissions to valid prefixed values) at the expense of inserting additional dependent logic (tissues from worm organisms must be uberon or worm specific). Alternatively, keeping the rule specific checks would mean that the schema_definition.yaml is closer to the single source of truth and lowers the amount of code needed to maintain the overall validator.
> 
> If that's the design choice the team is ok with, these changes look ready to go.

So the additional organism-specific logic is actually fairly easy to add into the `schema_definition` with the current design. You would just add an additional dependency saying, "for zebrafish, terms must be either UBERON or ZFA" etc. The part that's tricky with the current schema design is how you would write the rule: "for organisms that are not zebrafish, fruit fly, or roundworm, only UBERON terms are allowed". Needing to build that rule into the schema definition would require writing it as something like:
```
tissue_ontology_term_id:
  dependencies:
    - rule:
      column: organism_ontology_term_id
      exclude_exact: zebrafish, fruit fly, roundworm
      curie_constraints:
        ontologies:
          - UBERON
```

In addition to all of the organism-specific logic for zebrafish, fruit fly, and roundworm. Writing out all of this in the schema_definition would make it much more difficult to understand imo, even if it does keep the logic in one place. We would also have to add support for a new `exclude_exact` rule. This isn't even getting in to the additional complexity that is introduced from the fact that tissue_ontology_term_id has tissue_type rules as well as organism-level rules, and it isn't currently possible to write dependency-level rules that check against multiple column dependencies.

The real issue I think, is that our current system is designed to assume that field-level rules are more loose than dependency-level rules, and the assumption that dependency-level rules are independent of each other. But what we're looking for here is we want the field-level rule to be more strict than the organism-dependency-level rules. I think what we may want to support is some sort of `if / ifelse / else` structure for dependency rules, but I think overhauling the semantic language of the schema definition is sort of out of scope given that we're supposed to have this work done in a week.

I would say for now it's fine to push through with a somewhat hacky solution. We're going to need to support additional species in Q2 that require pulling in full ontologies, so when we do that I think we should just make sure to scope in some infra improvements to the Q2 multi species work.

ejmolinelli (approved):

PR #1231, opened 2025-01-30
chore: move organism-specific validation logic to schema_definition.yaml

nayib-jose-gloria:
## Reason for Change

- follow-up proposal to discussion [here](https://github.com/chanzuckerberg/single-cell-curation/pull/1229#issuecomment-2624989955)

## Changes
- move logic for validating organism-specific cell type and tissue requirements into schema_definition.yaml
- for tissue and cell type, defined a "default" rule for columns not matching any dependency rules in schema_definition.yaml. We use this pattern (similar to an else statement) in other places like [development_stage](https://github.com/chanzuckerberg/single-cell-curation/blob/main/cellxgene_schema_cli/cellxgene_schema/schema_definitions/schema_definition.yaml#L460) 
- while I can see certain column dependency combinations becoming complicated in the future, tissue_type and  organism_ontology_term_id dependencies for tissue ontology term ID actually can be enforced in the same list of dependencies without issue (pls correct me if you saw an issue I'm missing, but tests are passing)

## Testing

- kept same tests as in main, just changed some error msgs around to correspond to different approach 
- skipped some failing map species tests since the schema has changed since that was written and its being deprecated

## Notes for Reviewer


nayib-jose-gloria:
also whether we go with this pattern or not, definitely makes me think we should add a maintenance task to create at least a simple doc or json schema defining the keys we use in schema_definition.yaml. And we should add an explicit key for things like the "else" statement functionality (we use this pattern in a few spots, and its baked into the validator logic that we intentionally support this pattern, but there's no key or doc to actually signal this is a feature)

joyceyan (approved):

PR #1235, opened 2025-01-31
fix: tissue_ontology_term_id and cell_type_ontology_term_id fields for 3 new species

joyceyan:
## Reason for Change

there's a few issues here that this PR addresses:
- we need to make sure that zebrafish / fruit fly / roundworm still accepts valid CL terms for `tissue_ontology_term_id`, but _only_ when `tissue_type == cell culture`
- also, UBERON terms should be supported for `tissue_ontology_term_id` even if species is zebrafish / fruit fly / roundworm
- and CL terms should be supported for `cell_type_ontology_term_id`  even if species is zebrafish / fruit fly / roundworm

this PR: https://github.com/chanzuckerberg/single-cell-curation/pull/1231/files introduced some issues with these 3 cases weren't being properly handled with moving the logic to the schema def. so i've moved the organism-specific logic back into the validate script directly.

we will revisit an approach to writing out the validation rules here sometime next week.

## Testing

- adds test coverage for these 3 cases, which was missing earlier

## Notes for Reviewer

### DIFF
--- a/cellxgene_schema_cli/cellxgene_schema/schema_definitions/schema_definition.yaml
+++ b/cellxgene_schema_cli/cellxgene_schema/schema_definitions/schema_definition.yaml
@@ -108,85 +108,33 @@ components:
     columns:
       cell_type_ontology_term_id:
         type: curie
-        dependencies:
-          - # If organism is Zebrafish
-            rule:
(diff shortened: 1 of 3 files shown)


ejmolinelli (approved):
looks good.

PR #1237, opened 2025-01-31
MACAQUE - DO NOT MERGE

ejmolinelli:
## Reason for Change
- supporting macaque NCBITaxon:9541

## Changes
- enumerate allowed NCBITaxons for `organism_ontology_term_id` as in 5.3 spec
- update gene references and run `gene-processing` action
- add NCBITaxon:9541 to supported species in gencode.py

## Testing
- some tests fail due to static expectations in pytest fixtures.


cellxgene_schema_cli/tests/fixtures/examples_validate.py
```diff
@@ -402,7 +402,8 @@
 var_expected = pd.DataFrame(
     [
         ["spike-in", False, "ERCC-00002 (spike-in control)", "NCBITaxon:32630", 1061, "synthetic"],
-        ["gene", False, "MACF1", "NCBITaxon:9606", 2821, "protein_coding"],
+        #["gene", False, "MACF1", "NCBITaxon:9606", 2821, "protein_coding"],
```
ejmolinelli, line 405: @joyceyan  @nayib-jose-gloria 

This is a pretty gnarly downstream consequence of adding species. After speaking with @joyceyan, this is what's happening.

# TLDR
unique-ifying feature-names across all genecode files (e.g. MACF1 -> MACF1_ENSG00000127603 for humans, and MACF1 -> MACF1_ENSMFAG00000036050 for macaque) may break tests that have explicitly hardcoded feature-names for testing e.g. "MACF1".

# Additional Detail
When you add a new species to `gene_info.yaml` and run `make gene-processing`, the code will download all of the source annotation files and rebuild `gene_<species>.csv.gz` datasets. If a feature-name collision is detected (i.e. MACF1 exists in both humans and macaque), it will unique-ify the feature name by suffixing the feature name with the feature-id (i.e. MACF1_ENSG00000127603) for each occurrence. 

This means:
1: updating one `.gtf` could result in changes to multiple `*.csv.gz` files if new collisions are detected
2: old feature-name for each of those collisions (i.e. MACF1) will disappear, potentially breaking hardcoded tests.

## Resolution
### option 1 - manually update hardcoded values
As shown here, I've manually deleted (commented) the breaking test case and updated it with the expected value.

This can be tedious, but given the small set of exemplar data in the fixtures, this is tolerable. This is the quickest solution.

One risk, is that the tester may be unaware of any third-party expectations on the presence of specific gene names. "Will anything break if we remove MACF1 and replace with MACF1_ENSG00000127603?"

### option 2 - generate fixtures by reading `*.csv.gz` directly
This defers the responsibility to the unique-ification step in `make gene-processing`, but would make it easier to test by sampling randomly from lines in the `*.csv.gz` file.

This is somewhat risky to do on a short turn around, and I don't recommend this in the short term.

nayib-jose-gloria: I think we use such a small set of feature names that updating it manually is tolerable IMO. And useful to ensure unique names are being persisted. 

I think the second option is also good but I wouldn't prioritize it for this project; at most, add it to the maintenance bucket.

PR #1249, opened 2025-02-04
fix: cell culture edge case for zebrafish, fly, roundworm

joyceyan:
## Reason for Change

feedback from brian mott: https://czi-sci.slack.com/archives/C07AV4NU9D2/p1738627367824459?thread_ts=1738345846.792719&cid=C07AV4NU9D2

When tissue_type == cell culture, then the rules for cell_type_ontology_term_id should apply for tissue_ontology_term_id. However, only CL terms are valid in this case, instead of both CL and the respective organism cell ontology (FBbt, ZFA, or WBbt)

i also picked up one of the minor nits which was to remove the `\n` from error messages

## Changes

- updates the validator to ensure that this case is now caught properly. this does make the schema definition even more janky.
- @ejmolinelli i do recall you mentioning that we can easily extend the validator to name rules and apply rules in certain areas. i wasn't sure if you made much (or any) progess actually writing code for that, but this would definitely be a good place for that. i decided against actually trying to implement that change in this PR since i didn't want to duplicate work if you had already started working on this.

## Testing

- wrote tests to verify correctness
- i updated some existing tests to just validate yes / no instead of asserting on exact error message copy

## Notes for Reviewer

### DIFF
--- a/cellxgene_schema_cli/cellxgene_schema/schema_definitions/schema_definition.yaml
+++ b/cellxgene_schema_cli/cellxgene_schema/schema_definitions/schema_definition.yaml
@@ -192,8 +192,7 @@ components:
                   terms:
                     - NCBITaxon:6239
             error_message_suffix: >-
-              When 'organism_ontology_term_id' is 'NCBITaxon:6239' (Caenorhabditis elegans),
-              'sex_ontology_term_id' MUST be 'PATO:0000384' for male, 'PATO:0001340' for hermaphrodite, or 'unknown'.
+              When 'organism_ontology_term_id' is 'NCBITaxon:6239' (Caenorhabditis elegans), 'sex_ontology_term_id' MUST be 'PATO:0000384' for male, 'PATO:0001340' for hermaphrodite, or 'unknown'.
             type: curie
             curie_constraints:
               ontologies:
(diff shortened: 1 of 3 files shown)


brianraymor:
Hi @joyceyan -please do not update the CLI code *before* the schema is updated. Clarifying the requirement with @jahilton on [**#sci-data-eng**](https://czi-sci.slack.com/archives/C07AV4NU9D2/p1738632581006799?thread_ts=1738345846.792719&cid=C07AV4NU9D2).

ejmolinelli:
Hi @joyceyan I did say that we can used named rule sets. However, that was an experimental branch that we abandoned to shift our focus to delivering on this release instead of spending time updating the validator.

It wold be a relatively easy addition to the schema but could potentially introduce infinite loops so some caution and extra testing would be required. We should add this to the maintenance tasks and flag this issue as an example where that feature would come in handy.

ejmolinelli (approved):
Hey Joyce. This works as far as I can tell. A few non-urgent nits.

1. Add to the `test_cell_culture_tissue_ontology_term_id_invalid` tests using the wrong organism specific ontology e.g. ZFA for worm, WBt for zebrafish,
2. maintain the allowed ontology prefix map as a constant outside of the validation function, so it's perhaps easier to identify as a global constraint and not an implementation detail.

PR #1253, opened 2025-02-07
fix: determine species from gene file, not from gene prefix

joyceyan:
## Reason for Change

it seems like the issue with `RR` prefixed genes with fruit fly was a bit of a patchwork solution to a slightly bigger problem, which is that inferring organism species based on gene id prefix is not entirely correct. see [this](https://czi-sci.slack.com/archives/C07AV4NU9D2/p1738957962258369?thread_ts=1738896679.479739&cid=C07AV4NU9D2) issue with a gorilla gene prefixed with `ENSGGOG` gets incorrectly mapped to homo sapiens

## Changes

- updates implementation of `gencode.get_organism_from_feature_id` to go through each individual gene checker, rather than rely on gene prefix

## Testing

- updates a test in `test_gencode` to assert that we're mapping the organism from feature id correctly

## Notes for Reviewer

### DIFF
--- a/cellxgene_schema_cli/cellxgene_schema/gencode.py
+++ b/cellxgene_schema_cli/cellxgene_schema/gencode.py
@@ -30,7 +30,7 @@ def get_organism_from_feature_id(
     feature_id: str,
 ) -> Union[SupportedOrganisms, None]:
     """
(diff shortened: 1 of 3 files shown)


Bento007 (approved):
that's a nice fix.

PR #1262, opened 2025-02-24
feat: verify anndata.obs["organism_ontology_term_id"] is support Atac…

Bento007:
## Reason for Change

… seq species.

Previously feature_reference was used, but this term is not populated during validation. Therefor, organism_ontology_term_id must be used during validation.

## Changes
- replace the use of anndata.var["feature_references"] with anndata.obs["organism_ontology_term_id"] during atac validation.
- remove var from test anndata.
- update tests

## Testing

- tests have been updated


ivirshup (approved):
lgtm

Issue #1283, opened 2025-03-04
Multiome: using incorrect chromosome table returns an ambiguous pandas exception rather than a validation error message


brianraymor:
# Context

Reported by @brian-mott on [**sci-data-eng**](https://czi-sci.slack.com/archives/C07AV4NU9D2/p1741118567470999?thread_ts=1741117809.630769&cid=C07AV4NU9D2):

When using human chromosome in a mouse fragment (and vice versa), get this error: pandas.errors.IntCastingNaNError: Cannot convert non-finite values (NA or inf) to integer

Issue #1282, opened 2025-03-04
Multiome: negative values in the fifth_column must fail validation


brianraymor:
# Context

Reported by @brian-mott on [**sci-data-eng**](https://czi-sci.slack.com/archives/C07AV4NU9D2/p1741118567470999?thread_ts=1741117809.630769&cid=C07AV4NU9D2):

Read support with negative value is valid, with 0 is not valid

Issue #1281, opened 2025-03-04
Multiome: incorrect types in fragment columns return ambiguous pandas exception rather than a validation error message


brianraymor:
# Context

Reported by @brian-mott on [**sci-data-eng**](https://czi-sci.slack.com/archives/C07AV4NU9D2/p1741118567470999?thread_ts=1741117809.630769&cid=C07AV4NU9D2):

For various fragment columns, setting the dtype to something other than expected raises a pandas exception instead of a validation error

Issue #1280, opened 2025-03-04
Mulitome: unsupported organism_ontology_term_id returns an ambiguous KeyError rather than a validation error message


brianraymor:
# Context

Reported by @brian-mott on [**sci-data-eng**](https://czi-sci.slack.com/archives/C07AV4NU9D2/p1741118567470999?thread_ts=1741117809.630769&cid=C07AV4NU9D2):

Non-human or non-mouse organism terms fail in this function `validate_fragment_stop_coordinate_within_chromosome` but not due to a check for correct organism term but because a KeyError is raised when trying to determine the correct organism chromosome length table through the dictionary `organism_ontology_term_id_by_chromosome_length_table`

PR #1284, opened 2025-03-04
fix: atac fragment processing suggestion

Bento007:
## Reason for Change

- [Mulitome: unsupported organism_ontology_term_id returns an ambiguous KeyError rather than a validation error message](https://github.com/chanzuckerberg/single-cell-curation/issues/1280)
- [Multiome: negative values in the fifth_column must fail validation](https://github.com/chanzuckerberg/single-cell-curation/issues/1282)
- [Multiome: using incorrect chromosome table returns an ambiguous pandas exception rather than a validation error message](https://github.com/chanzuckerberg/single-cell-curation/issues/1283)
- [Multiome: validation error messages are "lost in all the output"](https://github.com/chanzuckerberg/single-cell-curation/issues/1277)
- [Multiome: anndata validation should occur before fragment file validation](https://github.com/chanzuckerberg/single-cell-curation/issues/1278)
- [Multiome: include paired/unpaired assay check in process-fragment instead of as a separate cli command](https://github.com/chanzuckerberg/single-cell-curation/issues/1279)
- [Multiome: incorrect types in fragment columns return ambiguous pandas exception rather than a validation error message](https://github.com/chanzuckerberg/single-cell-curation/issues/1281)
## Changes

- silencing dask distributed log messages to make output easier to understand.
- reformat error message to make them easier to find in the output.
- check if anndata is atac before any other checks.
- perform anndata checks before fragments checks to speed of error feedback loop.
- check that `read support` is not <= 0
- check for mismatch chromosome with organism
- return organisms that are not allowed in the error response.
- catch pandas error when converting to parquet.

## Testing

version at HEAD of main with 
```
```

## Notes for Reviewer

### DIFF
--- a/cellxgene_schema_cli/cellxgene_schema/atac_seq.py
+++ b/cellxgene_schema_cli/cellxgene_schema/atac_seq.py
@@ -19,7 +19,7 @@ from dask.delayed import Delayed
 
 from .utils import is_ontological_descendant_of
 
-logger = logging.getLogger(__name__)
+logger = logging.getLogger("cellxgene-schema")
 
 # TODO: these chromosome tables should be calculated from the fasta file?
 # location of fasta https://www.gencodegenes.org/human/release_44.html and file name GRCh38.primary_assembly.genome.fa
@@ -276,7 +276,13 @@ def validate_fragment_stop_coordinate_within_chromosome(parquet_file: str, annda
             return "Anndata.obs.organism_ontology_term_id must have a unique value."
     chromosome_length_table = organism_ontology_term_id_by_chromosome_length_table[organism_ontology_term_id[0]]
     df = ddf.read_parquet(parquet_file, columns=["chromosome", "stop coordinate"])
-    df["chromosome_length"] = df["chromosome"].map(chromosome_length_table).astype(int)
+    # the chromosomes in the fragment must match the chromosomes for that organism
+    mismatched_chromosomes = set(df["chromosome"].unique().compute()) - chromosome_length_table.keys()
+    if mismatched_chromosomes:
+        return f"Chromosomes in the fragment do not match the organism({organism_ontology_term_id}).\n" + "\t\n".join(
+            mismatched_chromosomes
+        )
+    df["chromosome_length"] = df["chromosome"].map(chromosome_length_table, meta=int).astype(int)
     df = df["stop coordinate"] <= df["chromosome_length"]
     if not df.all().compute():
         return "Stop coordinate must be less than the chromosome length."
@@ -307,7 +313,8 @@ def validate_anndata_organism_ontology_term_id(anndata_file: str) -> Optional[st
 def detect_chromosomes(parquet_file: str) -> list[str]:
     logger.info("detecting chromosomes")
     df = ddf.read_parquet(parquet_file, columns=["chromosome"]).drop_duplicates()
-    return df["chromosome"].values.compute()
+    df = df["chromosome"].compute()
+    return df.tolist()
 
 
 def get_output_file(fragment_file: str, output_file: Optional[str]) -> str:
--- a/cellxgene_schema_cli/cellxgene_schema/atac_seq.py
+++ b/cellxgene_schema_cli/cellxgene_schema/atac_seq.py
@@ -13,10 +13,10 @@ import dask.distributed as dd  # TODO: see if distributed mode can be avoided.
 import h5py
 import pandas as pd
 import pysam
-from cellxgene_ontology_guide.ontology_parser import OntologyParser
 from dask import delayed
 from dask.delayed import Delayed
 
+from .ontology_parser import ONTOLOGY_PARSER
 from .utils import is_ontological_descendant_of
 
 logger = logging.getLogger("cellxgene-schema")
@@ -141,11 +141,10 @@ def check_anndata_requires_fragment(anndata_file: str) -> bool:
     :param anndata_file: The anndata file to validate.
     :return:
     """
-    onto_parser = OntologyParser()
 
     def is_atac(x: str) -> str:
-        if is_ontological_descendant_of(onto_parser, x, "EFO:0010891"):
-            if is_ontological_descendant_of(onto_parser, x, "EFO:0008913"):
+        if is_ontological_descendant_of(ONTOLOGY_PARSER, x, "EFO:0010891"):
+            if is_ontological_descendant_of(ONTOLOGY_PARSER, x, "EFO:0008913"):
                 return "p"  # paired
             else:
                 return "u"  # unpaired
@@ -249,7 +248,7 @@ def convert_to_parquet(fragment_file: str, tempdir: str) -> str:
 
 def report_errors(header: str, errors: list[str]) -> list[str]:
     if any(errors):
-        errors = [f"{i}: {e})" for i, e in enumerate(errors) if e is not None]
+        errors = [e for e in errors if e is not None]
         errors = [header] + errors
         logger.error("\n\t".join(errors))
         return errors
@@ -259,7 +258,7 @@ def report_errors(header: str, errors: list[str]) -> list[str]:
 
 def validate_anndata(anndata_file: str) -> list[str]:
     errors = [validate_anndata_organism_ontology_term_id(anndata_file), validate_anndata_is_primary_data(anndata_file)]
-    return report_errors("Errors found in Anndata file", errors)
+    return report_errors("Errors found in Anndata file. Skipping fragment validation.", errors)
 
 
 def validate_anndata_with_fragment(parquet_file: str, anndata_file: str) -> list[str]:
@@ -341,7 +340,7 @@ def validate_anndata_organism_ontology_term_id(anndata_file: str) -> Optional[st
         organism_ontology_term_ids = ad.io.read_elem(f["obs"])["organism_ontology_term_id"].unique().astype(str)
     if organism_ontology_term_ids.size > 1:
         error_message = (
-            "Anndata.obs.organism_ontology_term_id must have a unique value. Found the following values:\n"
+            "Anndata.obs.organism_ontology_term_id must have exactly 1 unique value. Found the following values:\n"
         ) + "\n\t".join(organism_ontology_term_ids)
         return error_message
     organism_ontology_term_id = organism_ontology_term_ids[0]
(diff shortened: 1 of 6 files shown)


cellxgene_schema_cli/cellxgene_schema/atac_seq.py
```diff
@@ -187,56 +187,95 @@ def process_fragment(
...
+        return errors
+    else:
+        return []
+
+
+def validate_anndata(anndata_file: str) -> list[str]:
+    try:
+        check_anndata_requires_fragment(anndata_file)
+    except ValueError as e:
+        return _report_errors("Errors found in Anndata file", str(e))
```
brian-mott, line 264: Just a quick comment, looks like it should be `[str(e)]` otherwise I get this:
```
ERROR:cellxgene-schema:Errors found in Anndata file
	A
	n
	n
	d
	a
	t
	a
	.
	o
	b
	s
	....
```

cellxgene_schema_cli/cellxgene_schema/atac_seq.py
```diff
@@ -273,18 +308,34 @@ def validate_fragment_stop_coordinate_within_chromosome(parquet_file: str, annda
...
+            error_message = (
+                "Anndata.obs.organism_ontology_term_id must have a unique value. Found the following " "values:\n"
+            ) + "\n\t".join(organism_ontology_term_id)
+            return error_message
+        organism_ontology_term_id = organism_ontology_term_id[0]
+    try:
+        chromosome_length_table = organism_ontology_term_id_by_chromosome_length_table[organism_ontology_term_id]
+    except KeyError:
+        return (
+            f"Anndata.obs.organism_ontology_term_id must be one of NCBITaxon:9606 or NCBITaxon"
```
nayib-jose-gloria, line 320: list keys from table instead of hard-coding the accepted values

cellxgene_schema_cli/cellxgene_schema/atac_seq.py
```diff
@@ -273,18 +308,34 @@ def validate_fragment_stop_coordinate_within_chromosome(parquet_file: str, annda
...
+            f":10090. Received {organism_ontology_term_id}."
+        )
```
nayib-jose-gloria, line 322: consolidate these checks with validate_anndata_ontology_term_id?

cellxgene_schema_cli/cellxgene_schema/atac_seq.py
```diff
@@ -143,7 +143,7 @@ def check_anndata_requires_fragment(anndata_file: str) -> bool:
     """
     onto_parser = OntologyParser()
```
nayib-jose-gloria, line 144: nit: pin to a schema_version? anndata CLI is doing so, we should do this to avoid potential mismatches

cellxgene_schema_cli/cellxgene_schema/atac_seq.py
```diff
@@ -19,7 +19,7 @@
 
 from .utils import is_ontological_descendant_of
 
-logger = logging.getLogger(__name__)
+logger = logging.getLogger("cellxgene-schema")
 
 # TODO: these chromosome tables should be calculated from the fasta file?
```
nayib-jose-gloria, line 24: Note: spoke to Trent about this in irl sync. Agreed that this issue should be tracked as a fast-follow for atac-seq validation, as this table should be aligned to the GENCODE version we're using for each pertinent species

cellxgene_schema_cli/cellxgene_schema/atac_seq.py
```diff
@@ -187,56 +187,91 @@ def process_fragment(
...
+        errors = [header] + errors
+        logger.error("\n\t".join(errors))
+        return errors
+    else:
+        return []
+
+
+def validate_anndata(anndata_file: str) -> list[str]:
+    errors = [validate_anndata_organism_ontology_term_id(anndata_file), validate_anndata_is_primary_data(anndata_file)]
+    return report_errors("Errors found in Anndata file", errors)
```
nayib-jose-gloria, line 262: Maybe add a warning / note that because the anndata failed these basic checks, we could not validate the fragment-based rules (to account for someone seeing these, fixing them, then being surprised when they get new errors)

cellxgene_schema_cli/cellxgene_schema/cli.py
```diff
@@ -77,6 +77,20 @@ def add_labels(input_file, output_file):
         sys.exit(1)
 
 
+def _check_anndata_requires_fragment(h5ad_file):
+    from .atac_seq import check_anndata_requires_fragment, report_errors
+
+    try:
+        fragment_required = check_anndata_requires_fragment(h5ad_file)
+        if fragment_required:
+            logger.info("Andata requires an ATAC fragment file.")
```
ejmolinelli, line 86: super nit - `Andata` typo?

cellxgene_schema_cli/cellxgene_schema/cli.py
```diff
@@ -77,6 +77,20 @@ def add_labels(input_file, output_file):
...
+        else:
+            logger.info("Andata does not require an ATAC fragment file.")
+    except Exception as e:
+        report_errors("Andata does not support ATAC fragment files for the follow reason", [str(e)])
```
nayib-jose-gloria, line 90: same typo

cellxgene_schema_cli/cellxgene_schema/atac_seq.py
ejmolinelli: definitely agree here! It's just waiting to get out of sync.

cellxgene_schema_cli/cellxgene_schema/atac_seq.py
ejmolinelli: could you import the existing instance from `validate.py`?

cellxgene_schema_cli/cellxgene_schema/atac_seq.py
```diff
@@ -187,56 +187,91 @@ def process_fragment(
...
-                logger.error(errors)
+            # quick checks
+            errors = validate_anndata(anndata_file)
+            if errors:
+                return errors
+
+            # convert the fragment to a parquet file for faster processing
+            try:
+                parquet_file = convert_to_parquet(fragment_file, tempdir)
+            except Exception:
```
brian-mott, line 207: Is it possible to propagate the Exception up the to the error message?

It looks like for the int columns in the fragment file, I will get a pandas parse exception with different dtypes.

For the chromosome column (col 1), there's this exception if there's a value that's not part of the expected categories:
`pyarrow.lib.ArrowInvalid: No non-null segments were available for field 'chromosome'; couldn't infer type`

And for the the barcode column, I think any given value is coerced to a string, so then the I get the validation error 'Barcodes don't match anndata.obs.index'

If it's too much to have more specific error messages for why the conversion failed, then maybe a message like : "Error converting fragment to parquet, check that dtypes for fragment file columns are consistent/match the schema"

cellxgene_schema_cli/cellxgene_schema/atac_seq.py
```diff
@@ -187,56 +187,91 @@ def process_fragment(
...
+        # remove the unzipped file
+        logger.debug(f"Removing {unzipped_file}")
+        unzipped_file.unlink()
+    parquet_file = str(parquet_file_path)
+    return parquet_file
+
+
+def report_errors(header: str, errors: list[str]) -> list[str]:
+    if any(errors):
+        errors = [f"{i}: {e})" for i, e in enumerate(errors) if e is not None]
```
brian-mott, line 252: Enumerate is a nice touch but I'd say not needed as long as each error is its own string/prints on a new line. Makes it easier to just check against the error instead of an error string with the enumeration plus the error message.

nayib-jose-gloria (commented):
left some more notes, but largely looks good. thanks!

cellxgene_schema_cli/cellxgene_schema/atac_seq.py
Bento007: moved ONTOLOGY_PARSER to it's own files to share across modules

cellxgene_schema_cli/cellxgene_schema/atac_seq.py
Bento007: https://github.com/chanzuckerberg/single-cell-curation/issues/1285

cellxgene_schema_cli/cellxgene_schema/atac_seq.py
Bento007: There errors should appear that way now

brian-mott: I'm still getting just the `Error converting fragment to parquet.` error message or the this raised exception for the first column: `pyarrow.lib.ArrowInvalid: No non-null segments were available for field 'chromosome'; couldn't infer type`

Ideally, if there's a parsing error at this step, it would be best to know which column or even which row failed so we could go find and correct the issue. With the current messages, I can very roughly know where to start to try and find the issue because I purposefully introduced it, but in a normal curation workflow, I'd be very lost as to what went wrong and how to fix it.

Bento007: What type of bad values are you thinking you will see?

Bento007: > `pyarrow.lib.ArrowInvalid: No non-null segments were available for field 'chromosome'; couldn't infer type`

Doesn't this tell you what column had the error?

brian-mott: That's a good question, I'm testing against different random data types that go against what is expected for each column, but it's hard to say what's actually out there in the wild as we get real submissions. And also if we curate and concatenate fragments from various donors/samples into one file, do we accidentally introduce an error?

We can wait and see what we come across, but at the very least, the error message should state something like parsing error, check that columns match schema definition if it's too hard to get more specific than that

brian-mott: > Doesn't this tell you what column had the error?
Yes, that at least mentions the right column

Bento007: fix the error message to be

> Error Parsing the fragment file. Check that columns match schema definition. Error:

with the message from the pandas or error appends to the end.

PR #1286, opened 2025-03-06
feat: schema 5.3.0 gene bumps

ejmolinelli:
## Reason for Change
Update (a) gene info files and (b) COG version to comply with cxg schema 5.3.0.

## Changes
- update COG version in requirements
- update to gene info.

## Testing
version at HEAD of main with 
```
```

## Notes for Reviewer

### DIFF
--- a/scripts/schema_bump_dry_run_genes/gene_bump_dry_run.py
+++ b/scripts/schema_bump_dry_run_genes/gene_bump_dry_run.py
@@ -205,7 +205,6 @@ def generate_deprecated_public(base_url: str, diff_map: Dict) -> Dict:  # type: 
             run_reporter.public_errored_datasets += 1
             logger.error(
                 f"Error processing public dataset {dataset['dataset_id']}, in collection {dataset['collection_id']}: "
-                f"{e}"
             )
     for collection in public_deprecated.values():
         # convert dataset_groups from a dictionary to a list of its values.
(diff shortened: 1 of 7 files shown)


scripts/schema_bump_dry_run_genes/gene_bump_dry_run.py
```diff
@@ -201,11 +201,10 @@ def generate_deprecated_public(base_url: str, diff_map: Dict) -> Dict:  # type:
             public_deprecated, is_deprecated_genes_found = compare_genes(dataset, diff_map, public_deprecated)
             if is_deprecated_genes_found:
                 run_reporter.public_deprecated_datasets += 1
-        except Exception as e:
+        except Exception:
             run_reporter.public_errored_datasets += 1
             logger.error(
                 f"Error processing public dataset {dataset['dataset_id']}, in collection {dataset['collection_id']}: "
-                f"{e}"
```
joyceyan, line 208: is there any reason why we would not want to log the actual exception that occurred?

scripts/schema_bump_dry_run_genes/gene_bump_dry_run.py
```diff
@@ -237,9 +236,9 @@ def generate_deprecated_private(base_url: str, diff_map: Dict) -> Tuple[Dict, Li
                 if revision_of and revision_of not in non_auto_migrated:
                     non_auto_migrated.add(revision_of)
                     private_deprecated[dataset["collection_id"]]["revision_of"] = revision_of
-        except Exception as e:
+        except Exception:
             run_reporter.private_errored_datasets += 1
-            logger.error(f"Error processing private dataset: {e}")
+            logger.error("Error processing private dataset")
```
joyceyan, line 241: i understand that we don't want to log stuff about private datasets

joyceyan (approved):
looks good, i think we may want to add back some additional logging info for dataset failures (ex: just the dataset id seems fine?) but i'm ok with removing it for now to err on the side of being more private / secure

scripts/schema_bump_dry_run_genes/gene_bump_dry_run.py
ejmolinelli: For expedience only. We have another issue to handle meaningful logs.

Issue #1297, opened 2025-03-17
cellxgene-schema CLI must update validation of X


brianraymor:
# Context

See [**#sci-data-eng**](https://czi-sci.slack.com/archives/C07AV4NU9D2/p1742242029050519). 

@Bento007 - please check the current validation of X to ensure that @jahilton's points are addressed. This current implementation may already address this?

> The specification of unpaired Accessibility not requiring raw counts means that paired Accessibility MUST have raw counts
> 
> So 5.3 should block a 10x multiome Dataset from being submitted without raw counts

brianraymor:
The related schema change has already been made and approved by @jahilton.

Bento007:
Is checking if raw exists in the anndata  enough for this check?

jahilton:
Datasets of paired assay (as defined in the schema) should be treated like RNA-seq datasets
>REQUIRED. Values MUST be de-duplicated molecule counts. Each cell MUST contain at least one non-zero value. All non-zero values MUST be positive integers stored as numpy.float32.

location...
>AnnData.raw.X unless no "normalized" is provided, then AnnData.X

PR #1299, opened 2025-03-18
fix: Atac Seq validate X if assay-ontology-id is a descendant of paired

Bento007:
## Reason for Change

- https://github.com/chanzuckerberg/single-cell-curation/issues/1297
- https://github.com/chanzuckerberg/single-cell-curation/issues/1296

## Changes

- Atac seq validation checks for raw if assay-ontology-id is a descendant of paired'
- Atac seq validation conversion sets NA value to "" when parsing the tsv. This will ensure missing barcodes are caught, as well as other NA values. 

## Testing

- Updated unit tests to cover the new checks.


cellxgene_schema_cli/cellxgene_schema/atac_seq.py
```diff
@@ -340,7 +356,7 @@ def validate_anndata_organism_ontology_term_id(anndata_file: str) -> Optional[st
         organism_ontology_term_ids = ad.io.read_elem(f["obs"])["organism_ontology_term_id"].unique().astype(str)
     if organism_ontology_term_ids.size > 1:
         error_message = (
-            "Anndata.obs.organism_ontology_term_id must have exactly 1 unique value. Found the following values:\n"
+            "Anndata.obs.organism_ontology_term_id must have exactly 1 unique value. Found the " "following values:\n"
```
ejmolinelli, line 359: Is the nested double quote a mistake?

Bento007:
We still need to resolved this comment https://github.com/chanzuckerberg/single-cell-curation/issues/1297#issuecomment-2731242567

ejmolinelli (approved):

cellxgene_schema_cli/cellxgene_schema/atac_seq.py
```diff
@@ -257,10 +257,24 @@ def report_errors(header: str, errors: list[str]) -> list[str]:
...
     return report_errors("Errors found in Anndata file. Skipping fragment validation.", errors)
 
 
+def validate_anndata_raw_counts(anndata_file: str) -> Optional[str]:
+    # check if the raw counts are required and present
+    with h5py.File(anndata_file) as f:
+        assay_ontology_term_ids = ad.io.read_elem(f["obs"])["assay_ontology_term_id"]
+        if is_atac(assay_ontology_term_ids.iloc[0]) != "p":
+            return
+        if "raw/X" not in f:
```
nayib-jose-gloria, line 274: We can't assume no raw/X == no raw counts. We must validate if X is a raw count matrix, and fail if both no raw.X AND X is not raw counts. 

That said, I think we already validate for this in the anndata validation. Not sure if there's an extra nuance to checking here or not, my understanding is that in practice this will always run in tandem with anndata validation so not sure if we need to add it here (someone else should validate whether that is a fair assumption on my part)

cellxgene_schema_cli/.coverage.66f0141.Evan-CZI-Laptop.local.32135.XeulyeFx
nayib-jose-gloria, line 1: @ejmolinelli a local coverage file of yours got committed

cellxgene_schema_cli/cellxgene_schema/atac_seq.py
```diff
@@ -323,7 +323,9 @@ def validate_fragment_stop_coordinate_within_chromosome(parquet_file: str, annda
 
 def validate_fragment_read_support(parquet_file: str) -> Optional[str]:
     # check that the read support is greater than 0
-    df = ddf.read_parquet(parquet_file, columns=["read support"], filters=[("read support", "<=", 0)])
+    df = ddf.read_parquet(
+        parquet_file, columns=["read support"], filters=[("read support", "<=", 0)], keep_default_na=False
```
Bento007, line 327:
```suggestion
        parquet_file, columns=["read support"], filters=[("read support", "<=", 0)]
```

PR #1309, opened 2025-03-20
Store intermediates compressed

ivirshup:
## Reason for Change

Intermediate sorted files were being stored uncompressed. Though one would expect dask dataframe to infer compression like pandas does, it doesn't appear to work correctly.

I would note that we should probably be able to write directly to bgzipped files, but I just haven't figured out how to do that elegantly through dask yet.

## Changes

- Explicitly compress intermediate sorted fragment files
- Decompress when re-compressing with bgzip

## Testing

version at HEAD of main with 
```
```

## Notes for Reviewer


Bento007 (approved):

ivirshup:
Unclear on the failures (I was only testing the pysam implementation locally), investigating now.

PR #1321, opened 2025-03-26
fix: scramble atac test data

Bento007:
## Reason for Change

Adding scrambled test data to verify sorting of the data is actually performed

## Changes

- generated files are in a temporary directory
- replace sorted test data with scrambled test data.


ivirshup (commented):
What exactly changed about the file? I wouldn't expect the values in `expected_chromosome_row_count` to change from a shuffle.

Bento007:
I looked at the test data closer because those numbers seemed off and verfied the file was being written incorrectly. Likely because the file not being locked. When I made the sample data, I pulled 1002 of each chromosome, but for some there were less than 1002. So any chromosome count over 1002 seemed strange.

ivirshup:
Maybe the test could compare the result to the input file instead of hard coding?

E.g. I believe we can assume:

* The result has the same number of total rows
* The same number of rows per chromosome

For small test cases this could be:

```python
import pandas as pd

def count_fragments_per_chromosome(fragment_file):
    fragments = pd.read_csv(fragment_file, sep="\t", header=None, names=["chromosome", "start coordinate", "stop coordinate", "barcode", "read support"])
    fragments_per_chromosome = fragments["chromosome"].value_counts().sort_index()
    return fragments_per_chromosome

orig_chrom_counts = count_fragments_per_chromosome(INPUT_FRAGMENTS)
new_chrom_counts = count_fragments_per_chromosome(PROCESSED_FRAGMENTS)

pd.testing.assert_series_equal(orig_chrom_counts, new_chrom_counts)
```

---

I'm not sure I understand yet how this test passed previously, especially against the PR that should have fixed the race condition.

Bento007:
We will still want to check that the tabix index is fetch the correct number of rows from the fragment file.

ivirshup (approved):

PR #1322, opened 2025-03-26
feat: schema 5.3 migration

ejmolinelli:
## Reason for Change
These fixes address issues that emerged as part of the schema migration. 
- ERROR: 'CL:0000000' in 'cell_type_ontology_term_id' is not an allowed term id.
- sparse array length is ambiguous; use getnnz() or shape[0]  - this error occurs as a result of a sparse matrix in `anndata.uns`

## Changes
- allow specific terms in schema_definition e.g. CL:0000000
- support reading `csc` sparse matrices in `uns` element
- support for validating dask arrays in `uns` element

## Testing
- added tests for validating empty/non-empty `anndata.uns` sparse matrices, regardless of sparse format.

## Notes for Reviewer

### DIFF
--- a/cellxgene_schema_cli/cellxgene_schema/schema_definitions/schema_definition.yaml
+++ b/cellxgene_schema_cli/cellxgene_schema/schema_definitions/schema_definition.yaml
@@ -121,6 +121,9 @@ components:
           exceptions:
             - unknown
           allowed:
+            terms:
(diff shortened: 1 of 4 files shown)


nayib-jose-gloria (approved):
address the lint errors, looks good otherwise!

PR #1346, opened 2025-04-24
feat: validate that no two rows in obs are identical by row counts

joyceyan:
## Reason for Change

https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/single-cell-curation/1317

## Changes

- after spending ~30 mins writing my own hash algorithm that all seemed to break my local machine, i pivoted and based this PR off of this code snippet lattice provided for evaluating duplicate obs rows: https://github.com/Lattice-Data/lattice-tools/blob/a9f7129f43e5cc1e910e513d0efd3ca1412e59cb/cellxgene_resources/cellxgene_mods.py#L450-L513
- i made some changes to this, notably:
  - switched from `sha1` to `sha224`, which is faster
  - updated it to support reading in a dask array instead of a csr matrix
  - some basic code clean up like changing variable names, using list comprehension, etc

note that i also put all of this logic in a separate file, `check_duplicates.py` instead of throwing everything into `validate.py`, and putting all relevant tests into `test_check_duplicates.py`. personally i would prefer moving some of our validation logic out of `validate.py` and into separate files / modules, so if this structure makes sense, maybe we can do this moving forward and potentially refactor our existing code.

## Testing

i wrote some tests to verify that we're raising a validation error for duplicate rows. a lot of our test data was actually seeded with duplicate rows (ex: matrices with all 0s), so some unrelated tests had to be fixed as well.

to settle on `sha224`, i benchmarked all of the hash algorithms in `hashlib` with this script, and using one very large dataset:
```
import numpy as np
import pandas as pd
import hashlib
import anndata
import time
from scipy import sparse

HASH_FUNCTIONS = {
    "sha1": lambda b: hashlib.sha1(b).hexdigest(),
    "sha224": lambda b: hashlib.sha224(b).hexdigest(),
    "sha256": lambda b: hashlib.sha256(b).hexdigest(),
    "sha384": lambda b: hashlib.sha384(b).hexdigest(),
    "sha512": lambda b: hashlib.sha512(b).hexdigest(),
    "sha3_224": lambda b: hashlib.sha3_224(b).hexdigest(),
    "sha3_256": lambda b: hashlib.sha3_256(b).hexdigest(),
    "sha3_384": lambda b: hashlib.sha3_384(b).hexdigest(),
    "sha3_512": lambda b: hashlib.sha3_512(b).hexdigest(),
    "shake_128": lambda b: hashlib.shake_128(b).hexdigest(16),
    "shake_256": lambda b: hashlib.shake_256(b).hexdigest(16),
    "blake2b": lambda b: hashlib.blake2b(b).hexdigest()
}


def _check_duplicate_obs_sparse(adata, hash_name="blake2b") -> list[str]:
    """
    Checks for duplicate rows in a CSR matrix using a specified hash function.
    """
    start_time = time.time()
    errors = []

    hash_fn = HASH_FUNCTIONS[hash_name]

    if "in_tissue" in adata.obs.columns:
        obs_to_keep = adata.obs[adata.obs["in_tissue"] != 0].index
        adata = adata[obs_to_keep, :]

    matrix = adata.raw.X if adata.raw else adata.X

    if not isinstance(matrix, sparse.csr_matrix):
        errors.append("Matrix not in sparse csr format, please convert before hashing")
        return errors

    if not matrix.has_canonical_format:
        matrix.sort_indices()
        matrix.sum_duplicates()
    
    assert matrix.has_canonical_format, "Matrix still in non-canonical format"

    indptr = matrix.indptr
    data = matrix.data
    indices = matrix.indices

    def row_hash(row_slice):
        return hash_fn(row_slice.tobytes())

    data_hashes = [row_hash(data[indptr[i] : indptr[i + 1]]) for i in range(matrix.shape[0])]
    index_hashes = [row_hash(indices[indptr[i] : indptr[i + 1]]) for i in range(matrix.shape[0])]

    hash_df = adata.obs.copy()
    hash_df["data_hash"] = data_hashes
    hash_df["index_hash"] = index_hashes

    dup_mask = hash_df.duplicated(subset=["data_hash", "index_hash"], keep=False)
    dup_df = hash_df[dup_mask].copy()

    if not dup_df.empty:
        errors.append(f"[{hash_name}] Found {len(dup_df)} duplicated raw counts in obs")

    print(f"[{hash_name}] _check_duplicate_obs_sparse took", round(time.time() - start_time, 2), "seconds")
    return errors

adata = anndata.read_h5ad("big_boi.h5ad")

for hash_name in HASH_FUNCTIONS:
    errs = _check_duplicate_obs_sparse(adata, hash_name=hash_name)
    for err in errs:
        print(err)
```
which showed sha224 to be the fastest:
```
[sha1] _check_duplicate_obs_sparse took 33.67 seconds
[sha224] _check_duplicate_obs_sparse took 28.74 seconds
[sha256] _check_duplicate_obs_sparse took 30.52 seconds
[sha384] _check_duplicate_obs_sparse took 40.58 seconds
[sha512] _check_duplicate_obs_sparse took 42.03 seconds
[sha3_224] _check_duplicate_obs_sparse took 49.4 seconds
[sha3_256] _check_duplicate_obs_sparse took 51.16 seconds
[sha3_384] _check_duplicate_obs_sparse took 63.15 seconds
[sha3_512] _check_duplicate_obs_sparse took 80.38 seconds
[shake_128] _check_duplicate_obs_sparse took 44.04 seconds
[shake_256] _check_duplicate_obs_sparse took 53.53 seconds
[blake2b] _check_duplicate_obs_sparse took 48.85 seconds
```

## Notes for Reviewer

### DIFF
--- a/cellxgene_schema_cli/cellxgene_schema/validation_internals/check_duplicates.py
+++ b/cellxgene_schema_cli/cellxgene_schema/validation_internals/check_duplicates.py
@@ -23,7 +23,7 @@ def check_duplicate_obs(adata) -> list[str]:
 
     if adata.raw:
         to_validate.append(adata.raw.X)
-    
+
     for matrix in to_validate:
         matrix_format = get_matrix_format(matrix)
         if matrix_format == "csr" or matrix_format == "dense":
@@ -32,6 +32,7 @@ def check_duplicate_obs(adata) -> list[str]:
             return [
                 f"Unsupported matrix format: {matrix_format} Sparse matrices must be encoded as scipy.sparse.csr_matrix. Dense matrices can be np.ndarray"
             ]
+
 
 def _check_duplicate_obs_dask(adata, matrix: da.Array) -> list[str]:
     errors = []
@@ -44,16 +45,19 @@ def _check_duplicate_obs_dask(adata, matrix: da.Array) -> list[str]:
         if sparse.issparse(block):
             block = block.toarray()
 
-        return np.array([
-            hashlib.sha224(row.tobytes()).hexdigest()
-            for row in block
-        ], dtype=object).reshape(-1, 1)  # shape (rows, 1) for map_blocks
+        return np.array([hashlib.sha224(row.tobytes()).hexdigest() for row in block], dtype=object).reshape(
+            -1, 1
+        )  # shape (rows, 1) for map_blocks
 
-    row_hashes = matrix.map_blocks(
-        rowwise_hash_block,
-        dtype=object,
-        chunks=(matrix.chunks[0], (1,)),  # one hash per row
-    ).compute().ravel()
+    row_hashes = (
+        matrix.map_blocks(
+            rowwise_hash_block,
+            dtype=object,
+            chunks=(matrix.chunks[0], (1,)),  # one hash per row
+        )
+        .compute()
+        .ravel()
+    )
 
     hash_df = adata.obs.copy()
     hash_df["row_hash"] = row_hashes
--- a/cellxgene_schema_cli/cellxgene_schema/validation_internals/check_duplicates.py
+++ b/cellxgene_schema_cli/cellxgene_schema/validation_internals/check_duplicates.py
@@ -1,47 +1,54 @@
 import hashlib
 
+import anndata
 import dask.array as da
 import numpy as np
 from scipy import sparse
 from utils import get_matrix_format
 
 
-def check_duplicate_obs(adata) -> list[str]:
+def check_duplicate_obs(adata: anndata.AnnData) -> list[str]:
     """
     Checks for duplicate rows in obs by raw count.
 
-    The logic is split by whether it's a sparse or dense matrix.
-
     :rtype List of errors that were seen. If no errors, validation is good.
     """
+
+    errors = []
 
     if "in_tissue" in adata.obs.columns:
         obs_to_keep = adata.obs[adata.obs["in_tissue"] != 0].index
         adata = adata[obs_to_keep, :]
 
-    to_validate = [adata.X]
+    to_validate = [(adata.X, "adata.X")]
 
     if adata.raw:
-        to_validate.append(adata.raw.X)
+        to_validate.append((adata.raw.X, "adata.raw.X"))
 
-    for matrix in to_validate:
+    for matrix, matrix_name in to_validate:
         matrix_format = get_matrix_format(matrix)
         if matrix_format == "csr" or matrix_format == "dense":
-            return _check_duplicate_obs_dask(adata, matrix)
+            errors.extend(_check_duplicate_obs_dask(adata, matrix, matrix_name))
         else:
-            return [
+            errors.append(
                 f"Unsupported matrix format: {matrix_format} Sparse matrices must be encoded as scipy.sparse.csr_matrix. Dense matrices can be np.ndarray"
-            ]
+            )
+
+    return errors
 
 
-def _check_duplicate_obs_dask(adata, matrix: da.Array) -> list[str]:
+def _check_duplicate_obs_dask(adata: anndata.AnnData, matrix: da.Array, matrix_name: str) -> list[str]:
+    """
+    For each row in the matrix, we compute a hash of the row and store it in row_hashes.
+    We then check for duplicates in the row_hashes.
+    """
     errors = []
 
     if matrix.ndim != 2:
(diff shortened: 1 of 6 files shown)


cellxgene_schema_cli/tests/test_schema_compliance.py
```diff
@@ -162,6 +162,20 @@ def test_shapes(self, validator_with_adata):
             in validator.errors
         )
 
+    def test_csc_matrix_invalid(self, validator_with_adata):
```
joyceyan, line 165: while looking at the issue brought up in [this thread](https://czi-sci.slack.com/archives/C08MF1AJ6F5/p1745437756457979), i thought that we weren't raising a validation error at all for unsupported CSC matrices, so i added test coverage to check that. it turns out we already are, but i figured having test coverage for this is a good idea so i kept it in this PR lol

nayib-jose-gloria (commented):
how large was the matrix in the dataset used for benchmarking hash function performance? out of curiosity

cellxgene_schema_cli/cellxgene_schema/validation_internals/check_duplicates.py
```diff
@@ -0,0 +1,81 @@
...
+        .ravel()
+    )
+
+    hash_df = adata.obs.copy()
+    hash_df["row_hash"] = row_hashes
+    dup_df = hash_df[hash_df.duplicated(subset="row_hash", keep=False)].copy()
+
+    if not dup_df.empty:
+        duplicate_rows_to_print = []
+        for i, idx in enumerate(hash_df.index[:10]):
```
nayib-jose-gloria, line 75: nit: can we move "10" to a var named "max_dup_rows_to_print" or something similar? to more easily track where we're setting the cutoff for how many we're printing

nayib-jose-gloria: also, should this be `dup_df.index`?

cellxgene_schema_cli/cellxgene_schema/validation_internals/check_duplicates.py
```diff
@@ -0,0 +1,81 @@
...
+            duplicate_rows_to_print.append(f"row {i}: index = {idx}")
```
nayib-jose-gloria, line 76: it looks like this isn't printing which groups of rows are duplicates of each other--is that something worthwhile to include for lattice?

brianraymor: @jahilton @corismall - thoughts ^^^

cellxgene_schema_cli/cellxgene_schema/validation_internals/check_duplicates.py
joyceyan: yes, this should be `dup_df.index` 😬

i'll get some more feedback from lattice on what we should include in the error messages, but moving that into a separate variable seems reasonable

cellxgene_schema_cli/cellxgene_schema/validation_internals/check_duplicates.py
joyceyan: @jahilton i know the ask in slack [here](https://czi-sci.slack.com/archives/C08MF1AJ6F5/p1745421541504909) was to just print the number of duplicate rows, since the expectation is that the curator would open the anndata object for further investigation.

however, while i was debugging some of our existing tests, i found it helpful to print out at least some information about which rows were duplicates so i could examine how we were setting up the test data more closely. let me know if you think this would be helpful or if we can just stick to the count of duplicates.

nayib-jose-gloria (commented):
overall looks good! couple comments to address 

out of scope for this pr but bigger picture--the ingestion job runs on a 1 core machine in favor of resource conservation, so this check might be slower there than on our local. If it doesn't scale well we may want to introduce parallelization (that may be an involved change lol). 

it's probably fine in most cases but we do have an 11M and a few 4M row datasets.

cellxgene_schema_cli/cellxgene_schema/validation_internals/check_duplicates.py
corismall: Our workflow produces a dataframe (hash_df) of all the duplicates and the corresponding metadata, which is what we send to the contributors. In my experience when there are duplicates there are usually many of them so printing out the rows wouldn't be very useful. Having the count is a good double check.

PR #1349, opened 2025-04-29
feat: update validation of feature_is_filtered

joyceyan:
## Reason for Change

https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/single-cell-curation/137

## Changes

- updates `_validate_column_feature_is_filtered` as described in the schema. so if raw is missing, then we raise an error if there are any true values in `feature_is_filtered`
- additionally, if any of the columns in `adata.X` sum to 0 (aka they are all 0s), then we validate that either `feature_is_filtered = True` or the `adata.raw.X` column is also all 0s

## Testing

- wrote a test where the first column in `adata.X` is all 0s, feature_is_filtered is False, and `adata.raw.X` has non-zero values, which raises a validation error

## Notes for Reviewer


cellxgene_schema_cli/cellxgene_schema/matrix_utils.py
```diff
@@ -0,0 +1,135 @@
...
+import numpy as np
+from anndata.compat import DaskArray
+from dask.array import map_blocks
+from scipy import sparse
+
+SPARSE_MATRIX_TYPES = {"csr", "csc", "coo"}
+SUPPORTED_SPARSE_MATRIX_TYPES = {"csr"}
+
+
+def get_matrix_format(matrix: DaskArray) -> str:
```
joyceyan, line 13: i refactored some of the matrix util functions from `utils.py` into `matrix_utils.py`. just seemed a bit cleaner this way and easier to find helper methods.

cellxgene_schema_cli/tests/test_validate.py
```diff
@@ -1329,8 +1329,7 @@ def test_has_valid_raw(self, data, matrix_format, expected_result):
         validator = self.create_validator(data, matrix_format)
         assert validator._has_valid_raw() == expected_result
 
-    @mock.patch("cellxgene_schema.validate.get_matrix_format", return_value="unknown")
```
joyceyan, line 1332: i removed this mock since we weren't using it

cellxgene_schema_cli/cellxgene_schema/matrix_utils.py
```diff
@@ -0,0 +1,135 @@
...
+
+    for layer in adata.layers:
+        format = get_matrix_format(adata.layers[layer])
+        if format != "csr" and get_sparsity(adata.layers[layer], format) >= 0.5:
+            adata.layers[layer] = adata.layers[layer].map_blocks(sparse.csr_matrix, dtype=adata.layers[layer].X.dtype)
+
+    return adata
+
+
+def debug_print_matrix(matrix: ad.AnnData, matrix_name: str, max_rows=20, max_cols=20):
```
joyceyan, line 127: i added this to debug things earlier, not actually using it in actual code i'm committing. but i figured it could be useful in the future too

cellxgene_schema_cli/cellxgene_schema/matrix_utils.py
nayib-jose-gloria: I like this refactor, but [data portal imports get_matrix_format](https://github.com/chanzuckerberg/single-cell-data-portal/blob/e30a8593cb1a785dc4d71b5f81062f186cafacd0/backend/layers/processing/utils/cxg_generation_utils.py#L8) so this'll break it.

Maybe lets mark the other as deprecated, keep both for now, and then make the switch later? and then delete the utils one

nayib-jose-gloria: for safety, let's do that for any other functions that got moved over

cellxgene_schema_cli/cellxgene_schema/matrix_utils.py
```diff
@@ -0,0 +1,135 @@
...
+    Returns a NumPy array of sums.
+
+    For example, this matrix:
+    [
+        [1, 0, 0],
+        [1, 2, 0],
+        [1, 0, 1],
+    ]
+
+    would return a NumPy array of [3, 12, 1]
```
nayib-jose-gloria, line 54: nit: 
```suggestion
    would return a NumPy array of [3, 2, 1]
```

joyceyan: woops, nice catch

cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -578,8 +577,31 @@ def _validate_column_feature_is_filtered(self, column: pd.Series, column_name: s
...
+            for i, _ in enumerate(sum_X):
+                is_filtered = column[i]
+                raw_column_sum = sum_raw_X[i]
+                # If adata.X has all 0s for the column, then is_filtered must be True or adata.raw.X should also be all 0s
+                if sum_X[i] == 0:
+                    if is_filtered or raw_column_sum == 0:
+                        continue
+                    else:
+                        gene_name = self.adata.var_names[i]
+                        self.errors.append(
```
nayib-jose-gloria, line 598: the original ticket ask was for a warning--has that changed, or should we fix here?

also not blocking for the ticket but if it is meant to be a warning only, I was thinking whether it'd be worthwhile to gate this check behind a --warnings flag in the CLI. since warnings aren't really actionable/exposed in the ingestion pipeline, it isn't really useful to run this check in a deployed environment, only locally.

cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -578,8 +577,31 @@ def _validate_column_feature_is_filtered(self, column: pd.Series, column_name: s
...
+            return
+        else:
+            sum_X = compute_column_sums(self.adata.X)
+            sum_raw_X = compute_column_sums(self.adata.raw.X)
+            for i, _ in enumerate(sum_X):
+                is_filtered = column[i]
+                raw_column_sum = sum_raw_X[i]
+                # If adata.X has all 0s for the column, then is_filtered must be True or adata.raw.X should also be all 0s
+                if sum_X[i] == 0:
+                    if is_filtered or raw_column_sum == 0:
```
nayib-jose-gloria, line 594: Is it guaranteed that the order of columns in var, raw.X, and X will always match? i recall we've had this concern with other matrix checks, we might have to enforce canonical format first here (unless thats already guaranteed by a previous step?)

nayib-jose-gloria: if we don't guarnatee canonical format earlier, whether the order can be safely assumed to be the same could be a question for lattice ^

cellxgene_schema_cli/cellxgene_schema/validate.py
joyceyan: i was confused about that as well, but the final schema says we should raise an error if we encounter this issue. i guess the part of the ticket body that says "feature request: raise a warning when X" is actually part of the _context_ for the ticket, not the actual implementation details of the ticket

jason and brian confirmed we want an error here: https://czi-sci.slack.com/archives/C08MF1AJ6F5/p1745937982488189

cellxgene_schema_cli/cellxgene_schema/utils.py
```diff
@@ -11,19 +11,35 @@
...
+"""
+Ideally, these methods should all only live within matrix_utils. However, we
+currently import these into single-cell-data-portal, so we need to keep these
+in here for backwards compatibility until we can refactor that.
+"""
+SPARSE_MATRIX_TYPES = {"csr", "csc", "coo"}
+SUPPORTED_SPARSE_MATRIX_TYPES = {"csr"}
+
+
+def get_matrix_format(matrix: DaskArray) -> str:
```
joyceyan, line 32: i slightly renamed all the methods between `utils.py` and `matrix_utils.py` because i got a warning from python saying that i was redefining a variable on line 32 that was imported on line 14. it didn't actually seem like it caused an issue with test correctness, but it seemed a bit safer to just rename the methods entirely.

cellxgene_schema_cli/cellxgene_schema/validate.py
joyceyan: seems like according to isaac, we can generally assume that `adata.var_names[i]` will match `adata.X[i]`: https://czi-sci.slack.com/archives/C08MF1AJ6F5/p1746127011430139

it seems like it should be correct most of the time, but i updated the error to also include the index of the mismatch

nayib-jose-gloria (approved):

PR #1350, opened 2025-04-30
fix: package setup with validation_internals

joyceyan:
## Reason for Change

discovered that despite the tests passing, we can't actually run the CLI because of how we're doing the package setup

## Changes

- primary change is to update the package setup to just pull in everything from `cellxgene_schema`, rather than needing to further specify all the subdirectories in there. this seems fine to me as i think we can reasonably assume anything within that subdirectory we'd want to include

## Testing

- ran `pip install ~/src/single-cell-curation/cellxgene_schema_cli` and verified i could then run `cellxgene-schema validate input.h5ad`

## Notes for Reviewer

### DIFF
--- a/cellxgene_schema_cli/cellxgene_schema/validate.py
+++ b/cellxgene_schema_cli/cellxgene_schema/validate.py
@@ -13,7 +13,6 @@ import pandas as pd
 from anndata.compat import DaskArray
 from dask.array import map_blocks
 from scipy import sparse
-from validation_internals.check_duplicates import check_duplicate_obs
 
 from . import gencode, schema
 from .gencode import get_gene_checker
@@ -27,6 +26,7 @@ from .utils import (
     is_ontological_descendant_of,
     read_h5ad,
 )
+from .validation_internals.check_duplicates import check_duplicate_obs
 
 logger = logging.getLogger(__name__)
 
--- a/cellxgene_schema_cli/cellxgene_schema/validation_internals/check_duplicates.py
+++ b/cellxgene_schema_cli/cellxgene_schema/validation_internals/check_duplicates.py
@@ -3,8 +3,8 @@ import hashlib
 import anndata
 import dask.array as da
 import numpy as np
+from cellxgene_schema.utils import get_matrix_format
 from scipy import sparse
-from utils import get_matrix_format
 
 
 def check_duplicate_obs(adata: anndata.AnnData) -> list[str]:
--- a/cellxgene_schema_cli/setup.py
+++ b/cellxgene_schema_cli/setup.py
@@ -1,4 +1,4 @@
-from setuptools import setup
+from setuptools import find_packages, setup
 
 with open("requirements.txt") as fh:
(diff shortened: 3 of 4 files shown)


cellxgene_schema_cli/cellxgene_schema/validation_internals/check_duplicates.py
```diff
@@ -3,8 +3,8 @@
 import anndata
 import dask.array as da
 import numpy as np
+from cellxgene_schema.utils import get_matrix_format
```
joyceyan, line 6: this was a separate issue i caught, the import needs to be absolute instead of relative

cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -27,6 +26,7 @@
     is_ontological_descendant_of,
     read_h5ad,
 )
+from .validation_internals.check_duplicates import check_duplicate_obs
```
joyceyan, line 29: minor copy change to just have this be consistent with the other subdirectory imports

cellxgene_schema_cli/setup.py
```diff
@@ -15,15 +15,6 @@
...
     packages=["cellxgene_schema"],
-    package_dir={"cellxgene_schema": "cellxgene_schema"},
-    package_data={
-        "cellxgene_schema": [
-            "gencode_files/*gz",
-            "migrate_files/*json",
-            "schema_definitions/*yaml",
-            "migrate_files/*.txt",
-        ]
-    },
```
nayib-jose-gloria, line 26: i believe we need this, or else non-py files won't be included in the package

nayib-jose-gloria: nvm, we have a MANIFEST file already so this should be fine to remove

nayib-jose-gloria: https://github.com/chanzuckerberg/single-cell-curation/blob/main/cellxgene_schema_cli/MANIFEST.in

we should add to this manifest, however, `migrate_files/*json` and `migrate_files/*txt`. this directory contains files used by lattice for the migrate script, as needed

cellxgene_schema_cli/setup.py
```diff
@@ -15,15 +15,6 @@
     install_requires=requirements,
     python_requires=">=3.10",
     packages=["cellxgene_schema"],
```
nayib-jose-gloria, line 17: i tried locally, and this definition still doesn't include validation_internals for me, since it doesn't look for nested directories. 

however, we can import "find_packages" from setuptools to find nested subdirectories without having to explicitly list them here

```suggestion
    packages=find_packages(exclude=["tests", "scripts"]),
```

joyceyan: hm weird, okay i wonder why my local environment is different? i just pushed the change to use ` packages=find_packages(exclude=["tests", "scripts"]),` so hopefully that works for you

nayib-jose-gloria (approved):

PR #1353, opened 2025-05-02
fix: take smaller step when sorting.

Bento007:
## Reason for Change

- this PR aims to resolve https://czi.atlassian.net/jira/software/c/projects/VC/boards/1534?selectedIssue=VC-2554

## Changes

- Sort the fragments in smaller chunks. This will slowdown the sorting but it less likely to run out of memory. The additional slow down is in the additional times we need to filter the parquet files.
 
## Testing

version at HEAD of main with 
```
```

## Notes for Reviewer
- This needs to be tested on our compute infra with a large dataset to make sure it doesn't hurt the performance too much.

### DIFF
--- a/cellxgene_schema_cli/cellxgene_schema/atac_seq.py
+++ b/cellxgene_schema_cli/cellxgene_schema/atac_seq.py
@@ -312,9 +312,10 @@ def validate_fragment_stop_greater_than_start_coordinate(parquet_file: str) -> O
 
 @log_calls
 def validate_fragment_stop_coordinate_within_chromosome(parquet_file: str, anndata_file: str) -> Optional[str]:
+    organism_ontology_term_id = None
     with h5py.File(anndata_file) as f:
-        organism_ontology_term_id = ad.io.read_elem(f["obs"])["organism_ontology_term_id"].unique().astype(str)[0]
-    chromosome_length_table = organism_ontology_term_id_by_chromosome_length_table[organism_ontology_term_id]
+        organism_ontology_term_id = ad.io.read_elem(f["uns"]).get("organism_ontology_term_id")
+    chromosome_length_table = organism_ontology_term_id_by_chromosome_length_table.get(organism_ontology_term_id)
     t = ibis.read_parquet(f"{parquet_file}/**", hive_partitioning=True)
     df = t.group_by("chromosome").aggregate(max_stop_coordinate=t["stop coordinate"].max()).execute()
 
@@ -346,14 +347,9 @@ def validate_anndata_is_primary_data(anndata_file: str) -> Optional[str]:
 
 @log_calls
 def validate_anndata_organism_ontology_term_id(anndata_file: str) -> Optional[str]:
+    organism_ontology_term_id = None
     with h5py.File(anndata_file) as f:
-        organism_ontology_term_ids = ad.io.read_elem(f["obs"])["organism_ontology_term_id"].unique().astype(str)
-    if organism_ontology_term_ids.size > 1:
-        error_message = (
-            "Anndata.obs.organism_ontology_term_id must have exactly 1 unique value. Found the " "following values:\n"
-        ) + "\n\t".join(organism_ontology_term_ids)
-        return error_message
-    organism_ontology_term_id = organism_ontology_term_ids[0]
+        organism_ontology_term_id = ad.io.read_elem(f["uns"]).get("organism_ontology_term_id")
     allowed_terms = [*organism_ontology_term_id_by_chromosome_length_table.keys()]
     if organism_ontology_term_id not in allowed_terms:
         return f"Anndata.obs.organism_ontology_term_id must be one of {allowed_terms}. Got {organism_ontology_term_id}."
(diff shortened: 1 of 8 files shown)


cellxgene_schema_cli/cellxgene_schema/atac_seq.py
```diff
@@ -475,8 +479,14 @@ def prepare_fragment(
     :param write_algorithm:
     :return:
     """
+    step_size = 1_000_000_000  # 1 million
```
joyceyan, line 482: whoa, i didn't know you could just add underscores to python integers and it would work? 😮

cellxgene_schema_cli/cellxgene_schema/atac_seq.py
```diff
@@ -475,8 +479,14 @@ def prepare_fragment(
...
     for chromosome in chromosomes:
-        logger.info(f"Processing chromosome: {chromosome}")
-        temp_data = sort_fragment(parquet_file, tempdir, chromosome)
-        write_algorithm(temp_data, bgzip_output_file)
+        chromosome_length = human_chromosome_by_length[chromosome]
+        steps = max(chromosome_length // step_size, 1)
```
joyceyan, line 485: would we ever have an issue where chromosome_length > step_size ?

cellxgene_schema_cli/cellxgene_schema/atac_seq.py
Bento007: you can also do scientific notation, but I found the `_` easier to read for numbers less than a billion

cellxgene_schema_cli/cellxgene_schema/atac_seq.py
Bento007: Most of the chromsome_lengths > step_size. Can you expand on your question?

joyceyan: sorry i meant to ask if there would be an issue if chromosome_length < step_size, not the other way around 😅

joyceyan: i wrote a bit of python to double check that the loop is processing the full chromosome length, and this seems to be working as expected:
```
step_size = 1_000_000_000
chromosome_length = 5_000_000
steps = max(chromosome_length // step_size, 1)
for i in range(steps):
    chunk_start = i * step_size
    chunk_end = min((i + 1) * step_size, chromosome_length)
    print(f"Processing chromosome: chunk: {i}, range: {chunk_start}-{chunk_end}")
```

but i think i found a different edge case:
```
step_size = 1_000_000_000
chromosome_length = 2_500_000_000
steps = max(chromosome_length // step_size, 1)
for i in range(steps):
    chunk_start = i * step_size
    chunk_end = min((i + 1) * step_size, chromosome_length)
    print(f"Processing chromosome: chunk: {i}, range: {chunk_start}-{chunk_end}")
```
this outputs
```
Processing chromosome: chunk: 0, range: 0-1000000000
Processing chromosome: chunk: 1, range: 1000000000-2000000000
```
but i'd expect there to be another step that processes the range 2000000000-2500000000, right?

Bento007: nvm, I forgot how range works. Fixing it.

cellxgene_schema_cli/cellxgene_schema/atac_seq.py
```diff
@@ -475,9 +479,12 @@ def prepare_fragment(
     :param write_algorithm:
     :return:
     """
+    step_size = 10_000_000  # 10 million
     for chromosome in chromosomes:
-        logger.info(f"Processing chromosome: {chromosome}")
-        temp_data = sort_fragment(parquet_file, tempdir, chromosome)
-        write_algorithm(temp_data, bgzip_output_file)
-        Path(temp_data).unlink(missing_ok=True)  # clean up temporary file
```
nayib-jose-gloria, line 482: why remove this cleanup? seems like it'd be useful if we're increasing the # of files created

cellxgene_schema_cli/cellxgene_schema/atac_seq.py
```diff
@@ -475,9 +479,12 @@ def prepare_fragment(
...
+        chromosome_length = human_chromosome_by_length[chromosome]
```
nayib-jose-gloria, line 484: is this step guaranteed to only be run in human?

cellxgene_schema_cli/cellxgene_schema/atac_seq.py
joyceyan: that's a good question - i didn't notice it had removed it when i first reviewed this. i'm inclined to keep it unless there's a reason to remove it, @Bento007 ?

joyceyan:
@nayib-jose-gloria i updated this PR to use the organism specific chromosome table, but since i have another PR open that would produce conflicting changes [here](https://github.com/chanzuckerberg/single-cell-curation/pull/1366), i updated the base branch of this PR to `joyce/fix-organism`

cellxgene_schema_cli/cellxgene_schema/atac_seq.py
```diff
@@ -209,10 +209,16 @@ def process_fragment(
         else:
             logger.info("Fragment and Anndata file are valid")
 
+        organism_ontology_term_id = None
+        with h5py.File(anndata_file) as f:
+            organism_ontology_term_id = ad.io.read_elem(f["uns"]).get("organism_ontology_term_id")
+
```
nayib-jose-gloria, line 215: this gets called in a couple of the functions called by validate_anndata_with_fragment. probably worth fetching once and passing as an arg, I think

PR #1357, opened 2025-05-07
feat: support only one organism

joyceyan:
## Reason for Change

https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/single-cell-curation/1311

## Changes

this is the first change, which is altering the validator to only allow one organism per dataset. this involves moving the `organism_ontology_term_id` out of `obs` and into `uns`. at a high level, this meant:
- moving `organism_ontology_term_id` from `obs` to `uns` in `schema_definition.yml`. note that i kept the definition for `add_labels` the same, so it uses the key `to_column` keyword even though we're not writing `organism` as a column, it'll just be an additional key. but i figured it would be easier / more obvious to reuse code and schema defs this way.
- updating the code for `write_labels` and `remove_labels` to account for labeling curie keys in `uns`

## Testing

- i ended up needing to rewrite ~all of the test fixtures 🥲 because our base `adata` object had human + mouse data mixed in. so a lot of this PR is rewriting test fixtures not just to move this from a required column in `obs` to a required key in `uns`, but also to separate out human and mouse anndata
- the mouse tests now use a new fixture `adata_mouse`
- we had an existing test checking for a validation error when `organism_ontology_term_id` doesn't exist, i updated that test to remove that key in `uns` and validate that still fails
- all existing tests for valid `organism_ontology_term_id` were rewritten to read from `uns` instead of `obs`
- wrote a new test, `test_uns_added_labels` to verify that we write the `organism` label from `organism_ontology_term_id`

## Notes for Reviewer


cellxgene_schema_cli/cellxgene_schema/migrate.py
```diff
@@ -34,8 +34,6 @@
     "development_stage": DEV_STAGE_AUTO_MIGRATE_MAP,
     "disease": {
     },
-    "organism": {
```
joyceyan, line 37: i know we'll need to rewrite the migrate script, but planning to do that in a separate PR since this one is already a lot

cellxgene_schema_cli/cellxgene_schema/remove_labels.py
```diff
@@ -30,23 +30,33 @@ def remove_labels(self):
             if component is None:
                 continue
 
+            component_def = self.schema_def["components"][component_name]
```
joyceyan, line 33: minor refactor here to just use a variable name instead of referencing `self.schema_def["components"][component_name]` seven times

cellxgene_schema_cli/cellxgene_schema/remove_labels.py
```diff
@@ -30,23 +30,33 @@ def remove_labels(self):
...
 
             # Doing it for index
-            if "index" in self.schema_def["components"][component_name]:
-                index_def = self.schema_def["components"][component_name]["index"]
+            if "index" in component_def:
+                index_def = component_def["index"]
                 if "add_labels" in index_def:
                     self._remove_columns(component, index_def)
 
+            # Remove any labels added as dict keys
```
joyceyan, line 52: this is the bulk of the logic change to `remove_labels`, to just strip off any labels that were added as part of `uns`

cellxgene_schema_cli/cellxgene_schema/schema_definitions/schema_definition.yaml
```diff
@@ -37,6 +37,42 @@ components:
...
+                - NCBITaxon:9541
+                - NCBITaxon:9544
+                - NCBITaxon:10090
+                - NCBITaxon:9986
+                - NCBITaxon:9598
+                - NCBITaxon:10116
+                - NCBITaxon:9823
+        add_labels:
+          - type: curie
+            to_column: organism
```
joyceyan, line 75: i just copied and pasted this from `obs` to `uns` in the schema def. the `to_column` keyword doesn't make as much sense since this is a key not a column. i think it's slightly better to just keep the keywords consistent but i'm open to changing it

cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -701,51 +706,62 @@ def _validate_column_dependencies(
...
-                terms_to_match.update(dependency_def["rule"]["match_exact"]["terms"])
+                    terms_to_match.update(get_descendants(ONTOLOGY_PARSER, ancestor, include_self=True))
+            if "match_exact" in rule:
+                terms_to_match.update(rule["match_exact"]["terms"])
+
             try:
-                match_query = df[column_to_match].isin(terms_to_match)
+                if column_to_match:
+                    match_query = df[column_to_match].isin(terms_to_match)
+                elif uns_key_to_match:
```
joyceyan, line 734: `_validate_column_dependencies` was extended to support `uns_key` for fields like `development_stage_ontology_term_id` that depend on `uns["organism_ontology_term_id"]`

cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -964,6 +980,10 @@ def _validate_uns_dict(self, uns_dict: dict) -> None:
             elif value is not None and not isinstance(value, (np.bool_, bool, numbers.Number)) and len(value) == 0:
                 self.errors.append(f"uns['{key}'] cannot be an empty value.")
 
+            value_def = dict_def["keys"].get(key, None)
```
joyceyan, line 983: i had to update `_validate_uns_dict` to support validating against curie constraints

cellxgene_schema_cli/cellxgene_schema/write_labels.py
```diff
@@ -339,6 +339,21 @@ def _add_labels(self):
             if "add_labels" in index_def:
                 self._add_column(component, "index", index_def)
 
+        uns_def = self.schema_def["components"]["uns"]
```
joyceyan, line 342: had to add this here to support writing curie labels in `uns`

cellxgene_schema_cli/tests/fixtures/examples_validate.py
```diff
@@ -40,7 +40,6 @@
             "CL:0000066",
             "EFO:0009899",
             "MONDO:0100096",
-            "NCBITaxon:9606",
```
joyceyan, line 43: all of the fixtures here had to be rewritten because:
1. all existing test fixtures had human + mouse data combined, which is now not allowed
2. `organism_ontology_term_id` now moved to `uns`

cellxgene_schema_cli/tests/fixtures/examples_validate.py
```diff
@@ -512,14 +570,21 @@
     var=good_var,
 )
 
+adata_mouse = anndata.AnnData(
```
joyceyan, line 573: this adata fixture is used for the mouse tests now

cellxgene_schema_cli/cellxgene_schema/migrate.py
nayib-jose-gloria: You'll want to remove this from the migrate script [template](https://github.com/chanzuckerberg/single-cell-curation/blob/main/scripts/migration_assistant/migration_template.jinja#L36) rather than the migrate script directly

nayib-jose-gloria (commented):
side note--have you already looked into what changes need to be made in the data-portal to support this change?

cellxgene_schema_cli/cellxgene_schema/migrate.py
joyceyan: ah thanks for the call out, updated!

cellxgene_schema_cli/cellxgene_schema/schema_definitions/schema_definition.yaml
nayib-jose-gloria: I slightly prefer renaming it to `to_key` (or if you can think of a neutral term that captures both), but could be convinced otherwise

nayib-jose-gloria (approved):
logic and refactor looks good, thanks! 

non-blocking note: for add labels, i think we should use to_key for uns labels or otherwise switch to a neutral term if we want it to match what is used to add column labels.

cellxgene_schema_cli/cellxgene_schema/schema_definitions/schema_definition.yaml
joyceyan: i can rename it!

nayib-jose-gloria (approved):

PR #1369, opened 2025-05-28
fix: 6.0 one organism implementation

joyceyan:
## Reason for Change

- there were a few requested changes in this thread here: https://czi-sci.slack.com/archives/C08MF1AJ6F5/p1747337258902249

## Changes

- Raises a more explicit "not allowed" error for the exempt organisms (SARS COV-2 and ERCC), which are both no longer valid term ids for `uns["organism_ontology_term_id"]`
- Validates that each gene id provided in `var` belongs to the organism (or one of its descendants) in `uns["organism_ontology_term_id"]`

## Testing

- I had to modify a lot of our existing test fixtures which mostly covers the cases we would want to handle, but I did also add some explicit test coverage for the exempt organisms

## Notes for Reviewer

### DIFF
--- a/cellxgene_schema_cli/cellxgene_schema/validate.py
+++ b/cellxgene_schema_cli/cellxgene_schema/validate.py
@@ -479,7 +479,7 @@ class Validator:
                 is_descendant = organism_ontology_id in ONTOLOGY_PARSER.get_term_ancestors(dataset_organism, True)
                 if not is_descendant and organism_ontology_id not in gencode.EXEMPT_ORGANISMS:
                     invalid_gene_organisms.append(organism)
-        
+
         if len(invalid_gene_organisms) > 0:
             self.errors.append(
                 f"uns['organism_ontology_term_id'] is '{dataset_organism}' but feature_ids are from {invalid_gene_organisms}."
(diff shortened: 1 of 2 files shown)


cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -399,19 +400,26 @@ def _validate_curie(self, term_id: str, column_name: str, curie_constraints: dic
 
         :rtype None
         """
+        # Check if term_id is forbidden by schema definition. Sometimes, these are also invalid ontology
+        # terms, but it's preferred to report these as "not allowed" terms rather than "invalid ontology terms"
+        if "forbidden" in curie_constraints:
```
Bento007, line 405: Is there different if login here or is this just a reorganizing of logic?

Bento007 (approved):

cellxgene_schema_cli/cellxgene_schema/validate.py
joyceyan: no, it's just a reorganizing of logic so the forbidden terms check can occur before validating that they're valid ontology terms. but they're still required to be valid ontology terms before proceeding into the forbidden ancestors check.

PR #1377, opened 2025-06-02
feat: run gene processing script

joyceyan:
## Reason for Change

we need to generate new gene assets since we previously modified the `gene_processing.py` script

## Changes

- just made a whitespace change to trigger the GHA

## Testing

n/a

## Notes for Reviewer


joyceyan:
looks like i actually already ran this earlier in this PR: https://github.com/chanzuckerberg/single-cell-curation/pull/1359/files

PR #1382, opened 2025-06-05
feat: update migrate script

joyceyan:
## Reason for Change

a few changes here, based on feedback from lattice in [this thread](https://czi-sci.slack.com/archives/C08MF1AJ6F5/p1749151615477039?thread_ts=1749139807.104259&cid=C08MF1AJ6F5)

## Changes

- added a new util method `replace_delimiter` to migrate the delimiter of `self_reported_ethnicity_ontology_term_id`
- i updated the logic in `move_column_from_obs_to_uns` to log a warning instead of raising a keyerror if the column isn't found. mostly just because it makes writing the migration unit test easier, but i also think we shouldn't fail the migration if the column isn't found in case we need to re-run migration for something that is partially migrated or something like that.

## Testing

- added test coverage for `replace_delimiter`
- fixed a few random tests that were failing that were introduced when i merged my earlier PR when GHA's were down

## Notes for Reviewer

### DIFF
--- a/cellxgene_schema_cli/cellxgene_schema/migrate.py
+++ b/cellxgene_schema_cli/cellxgene_schema/migrate.py
@@ -91,10 +91,10 @@ def migrate(input_file, output_file, collection_id, dataset_id):
             {
                 "Healthy": "PATO:0000461",
                 "HIV": "MONDO:0005109",
-                "VL_HIV": "MONDO:0005109 || MONDO:0011989",
+                "VL_HIV": "MONDO:0005109 || MONDO:0005445",
                 "aL_HIV": "MONDO:0005109 || MONDO:0011989",
-                "cVL_HIV": "MONDO:0005109 || MONDO:0011989",
-                "pVL_HIV": "MONDO:0005109 || MONDO:0011989",
+                "cVL_HIV": "MONDO:0005109 || MONDO:0005445",
+                "pVL_HIV": "MONDO:0005109 || MONDO:0005445",
             }
         )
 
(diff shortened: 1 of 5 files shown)


cellxgene_schema_cli/cellxgene_schema/migrate.py
```diff
@@ -72,6 +80,40 @@ def migrate(input_file, output_file, collection_id, dataset_id):
...
+
+    if collection_id == "126afc71-47fb-4e9d-8aaf-9d9f61e0ac77":
+        utils.map_ontology_term(
+            dataset.obs,
+            "disease",
+            "disease_group",
+            {
+                "Healthy": "PATO:0000461",
+                "HIV": "MONDO:0005109",
+                "VL_HIV": "MONDO:0005109 || MONDO:0011989",
```
jahilton, line 94: "VL" is MONDO:0005445 (visceral leishmaniasis) instead of MONDO:0011989
so that can be updated for VL_HIV, cVL_HIV, pVL_HIV

jahilton (changes requested):
just 1 set of MONDO term updates

joyceyan:
@nayib-jose-gloria for both of those issues, i think we'd want to resolve them manually by updating the datasets. so if they fail validation during the dev migration i would just ping lattice and we'd figure out how to resolve them. for (1), we'd probably want to manually look at the dataset and see if we want to remove any rows. for (2), i could see that potentially being an issue if we somehow have datasets with multiple gene versions that were previously distinct from one another, but i don't really see how we'd programmatically resolve that so i think that would also require manual inspection / resolution.

nayib-jose-gloria (approved):

PR #1383, opened 2025-06-06
feat: downgrade feature_is_filtered errors to to warning

joyceyan:
## Reason for Change

https://czi-sci.slack.com/archives/C08MF1AJ6F5/p1749222387913889

## Changes

- seems like there are some legitimate cases where you would have all 0's, so we're downgrading this to a warning. i wrote up a ticket [here](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/single-cell/759) to pull in all the datasets we raise these warnings on, so lattice can evaluate them and figure out what we want to do long term

## Testing

updated tests as needed

## Notes for Reviewer


nayib-jose-gloria (commented):
can you fix the failing unit test? seems related to the version bump rather than this, logic here looks good

PR #1395, opened 2025-06-30
fix: memory optimization in feature_is_filtered validation

nayib-jose-gloria:
## Reason for Change

- OOMs on large datasets during latest migration dev run, identified this as a point of memory leakage (still OOMs on datasets with 2M+ cells even after increasing to 64GB RAM machines)
- original implementation introduced [here](https://github.com/chanzuckerberg/single-cell-curation/pull/1395/files)

## Changes

- use dask's built-in 'sum' function, which can sum over dask array sparse matrix columns in a memory-optimized way (manages intermediate sums efficiently using tree-reduction)
- set optional 'split_every' arg to 8, to limit number of intermediate artifacts held in RAM
- requires converting blocks to COO matrices to be compatible with built-in sum function
- custom implementation was holding many intermediate artifacts (dense array blocks) concurrently in memory, causing OOMs.
         -  Alternative approach would require limiting parallelization or otherwise re-creating dask's built-in sparse matrix sum function to work for csr matrices 
- added more debug logging for future investigation into potentially memory-intensive jobs 

## Tests
- dev migration run successfully on previously failing datasets using this branch

### DIFF
--- a/cellxgene_schema_cli/cellxgene_schema/cli.py
+++ b/cellxgene_schema_cli/cellxgene_schema/cli.py
@@ -14,8 +14,8 @@ logger = logging.getLogger("cellxgene_schema")
 )
 @click.option("-v", "--verbose", help="When present will set logging level to debug", is_flag=True)
 def schema_cli(verbose):
-    logging.basicConfig(level=logging.ERROR)
-    logger.setLevel(logging.DEBUG if verbose else logging.INFO)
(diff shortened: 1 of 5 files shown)


Bento007 (commented):
What is your justification for adding the `sparse` library as opposed to using scipy.sparse?

Bento007 (approved):

PR #1424, opened 2025-08-13
fix: resolve parquet duplication of rows

Bento007:
## Reason for Change

- https://czi.atlassian.net/browse/VC-3784
- When reading chunks of the parquet into bgzip, duplicate rows were created. To avoid this we can go directly from the source file to bgzip without convert to parquet. The original file will still be converted to parquet to speed up validation check.

## Changes

- When reading chunks of the parquet into bgzip, duplicate rows were created. To avoid this we can go directly from the source file to bgzip without convert to parquet. The original file will still be converted to parquet to speed up validation check.
- The complexity of converting to bgzip has been offloaded to common linux tools.

## Testing
- test have been updated to reflect the changes.

## Notes for Reviewer


Bento007:
>Do we have any reproducible cases of the bug that we can add as a test case?

I wasn't able to reproduce it, but narrowed down the possible area of code it could be caused.

> And do we know that there are no duplicate rows in the converted parquet file?

We know there are no duplicate in the parquet file because we have an explicit check for that as part of our validation. It's only after we write it back out to bgzip that we end up with duplicates.

PR #1438, opened 2025-09-05
feat: update cell_type and cell_type_ontology_term_id for 7.0

joyceyan:
## Reason for Change

https://czi.atlassian.net/browse/VC-3384

## Changes

- Updated the requirements for cell_type to require "na" when the cell_type_ontology_term_id is "na"
- Updated the requirements for cell_type_ontology_term_id to allow "na" when the tissue_type is "cell line"

## Testing

- added unit test coverage for this

## Notes for Reviewer

### DIFF
--- a/cellxgene_schema_cli/cellxgene_schema/validate.py
+++ b/cellxgene_schema_cli/cellxgene_schema/validate.py
@@ -572,7 +572,7 @@ class Validator:
         }
 
         def is_valid_row(row):
-            if row[cell_type_column] == "unknown":
+            if row[cell_type_column] == "unknown" or row[cell_type_column] == "na":
                 return True
             allowed = allowed_prefixes.get(organism_term, ("CL",))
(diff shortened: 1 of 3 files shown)


cellxgene_schema_cli/cellxgene_schema/validate.py
```diff
@@ -572,7 +572,7 @@ def _validate_cell_type_ontology_term_id(self):
         }
 
         def is_valid_row(row):
-            if row[cell_type_column] == "unknown":
+            if row[cell_type_column] == "unknown" or row[cell_type_column] == "na":
```
joyceyan, line 575: this function does organism-specific logic to determine if the `cell_type_ontology_term_id` is valid. we can assume that `na` will be valid here because the tissue_type logic is implemented in schema_definition and isn't dependent on what type of organism the cell is.

PR #1519, opened 2026-01-12
feat: Integrate genetic_perturbation annotations into cellxgene cli

Bento007:
… and genomic regions

### DIFF
--- a/cellxgene_schema_cli/cellxgene_schema/annotate_guides.py
+++ b/cellxgene_schema_cli/cellxgene_schema/annotate_guides.py
@@ -642,7 +642,7 @@ def _build_genomic_regions(valid_rows: pd.DataFrame) -> List[str]:
     return list(regions)
 
 
-def _build_target_features(valid_rows: pd.DataFrame) -> Dict[str, str]:
+def _build_derived_features(valid_rows: pd.DataFrame) -> Dict[str, str]:
     """Build derived_features dict from annotation rows.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -59,16 +59,6 @@ If there are multiple closest ancestors of the same distance with a match, the c
...
-
-This command requires [guidescan2](https://github.com/pritykinlab/guidescan-cli) to be installed on your system before use.
-
-```
-cellxgene-schema annotate-perturbations input.h5ad output.h5ad
-```
-
-The command will extract guide sequences from `genetic_perturbations`, run guidescan2 to find genomic matches, identify overlapping genes using bioframe, and update the h5ad file with the annotations.
-
 ---
```
Bento007:
```suggestion
```

cellxgene_schema_cli/cellxgene_schema/annotate_guides.py
```diff
@@ -643,7 +643,7 @@ def _build_genomic_regions(valid_rows: pd.DataFrame) -> List[str]:
 
 
 def _build_target_features(valid_rows: pd.DataFrame) -> Dict[str, str]:
```
surygeng: Do you want to keep "target" in the function name while the features are updated to "derived"?

Bento007: we should change it to be consistent. Good catch

PR #1550, opened 2026-04-13
Update schema.md

BAevermann:
Updated: 
Contact into
Removed:
OBSM requirements for pre-analysis schema
Added:
genetic_perturbations[id][‘intended_features']
genetic_perturbations[id][‘intended_features'][feature_id]

## Reason for Change

- #TICKET_NUMBER
- If the reason for this PR's code changes are not clear in the issue, state value/impact

## Changes

- 

## Testing

version at HEAD of main with 
```
```

## Notes for Reviewer

### DIFF
--- a/schema/7.1.0/schema.md
+++ b/schema/7.1.0/schema.md
@@ -1459,7 +1459,7 @@ When a dataset is uploaded, CELLxGENE Discover MUST annotate a unique observatio
 
 The value for each `str` key MUST be a  `numpy.ndarray` of shape `(n_obs, m)`, where `n_obs` is the number of rows in `X` and `m >= 1`. 
 
-Otherwise, to display a dataset in CELLxGENE Explorer, Curators MUST annotate **one or more** embeddings of at least two-dimensions (e.g. tSNE, UMAP, PCA, spatial coordinates) as `numpy.ndarrays` in `obsm`.<br><br>
+In order to display a dataset in CELLxGENE Explorer, uns['is_pre_analysis'] MUST NOT be True AND Curators MUST annotate one or more embeddings of at least two-dimensions (e.g. tSNE, UMAP, PCA, spatial coordinates) as numpy.ndarrays in obsm.<br><br>
 
 ### spatial
 
--- a/schema/7.1.0/schema.md
+++ b/schema/7.1.0/schema.md
@@ -1775,12 +1775,12 @@ Curators MUST annotate the following keys and values in `uns`:
     </tr>
     <tr>
       <th>Annotation</th>
-      <td>CELLxGENE Discover MUST annotate.  The key MUST be the <code>gene_id</code> attribute from the <a href="#required-gene-annotations">corresponding gene reference</a> of the <code>organism_ontology_term_id</code>.<br><br> Version numbers MUST be removed from the <code>gene_id</code> if it is prefixed with <code>"ENS"</code> for <i>Ensembl stable identifier</i>. See <a href="https://ensembl.org/Help/Faq?id=488">I have an Ensembl ID, what can I tell about it from the ID?</a> For example, if the <code>gene_id</code> is <code>“ENSG00000186092.7”</code>, then the <code><i>feature_id</i></code> MUST be <code>“ENSG00000186092”</code>.</td>
+      <td>CELLxGENE Curator MAY annotate.  The key MUST be the <code>gene_id</code> attribute from the <a href="#required-gene-annotations">corresponding gene reference</a> of the <code>organism_ontology_term_id</code>.<br><br> Version numbers MUST be removed from the <code>gene_id</code> if it is prefixed with <code>"ENS"</code> for <i>Ensembl stable identifier</i>. See <a href="https://ensembl.org/Help/Faq?id=488">I have an Ensembl ID, what can I tell about it from the ID?</a> For example, if the <code>gene_id</code> is <code>“ENSG00000186092.7”</code>, then the <code><i>feature_id</i></code> MUST be <code>“ENSG00000186092”</code>.</td>
     </tr>
     <tr>
       <th>Value</th>
         <td>
-          <code>str</code>. If a <code>gene_name</code> attribute is assigned to the <code>gene_id</code> attribute in the <a href="#required-gene-annotations">corresponding gene reference</a> of the <code>organism_ontology_term_id</code>, the value MUST be the <code>gene_name</code>. Otherwise, the value MUST be the <code><i>feature_id</i></code> key.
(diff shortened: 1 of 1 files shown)


schema/7.1.0/schema.md
```diff
@@ -1740,6 +1742,52 @@ Curators MUST annotate the following keys and values in `uns`:
...
 
+#### genetic_perturbations[<i>id</i>]['intended_features']
+<table><tbody>
+    <tr>
+      <th>Key</th>
+      <td>intended_features</td>
+    </tr>
+    <tr>
+      <th>Annotation</th>
+      <td>CELLxGENE Curator MAY annotate when <code>genetic_perturbations[<i>id</i>]['derived_features']</code> is annotated; otherwise, this key MUST NOT be present.
```
jahilton: The way this is written, one would NOT be able to submit intended_features if there were no derived_features calculated.
Which means, an intergenic target can't have intended_features.

I think we want to allow intended_features independent of derived_features (isn't the whole point that derived_features is insufficient for some use cases?). So would suggest just `CELLxGENE Curator MAY annotate`

schema/7.1.0/schema.md
```diff
@@ -1740,6 +1742,52 @@ Curators MUST annotate the following keys and values in `uns`:
...
+      </td>
+    </tr>
+    <tr>
+      <th>Annotation</th>
+      <td>CELLxGENE Curator MAY annotate.  The key MUST be the <code>gene_id</code> attribute from the <a href="#required-gene-annotations">corresponding gene reference</a> of the <code>organism_ontology_term_id</code>.<br><br> Version numbers MUST be removed from the <code>gene_id</code> if it is prefixed with <code>"ENS"</code> for <i>Ensembl stable identifier</i>. See <a href="https://ensembl.org/Help/Faq?id=488">I have an Ensembl ID, what can I tell about it from the ID?</a> For example, if the <code>gene_id</code> is <code>“ENSG00000186092.7”</code>, then the <code><i>feature_id</i></code> MUST be <code>“ENSG00000186092”</code>.</td>
+    </tr>
+    <tr>
+      <th>Value</th>
+        <td>
+          <code>str</code>. If a <code>gene_name</code> attribute is assigned to the <code>gene_id</code> attribute in the <a href="#required-gene-annotations">corresponding gene reference</a> of the <code>organism_ontology_term_id</code>, the value MUST be the <code>gene_name</code>. Otherwise, the value MUST be the <code><i>feature_id</i></code> key.
```
jahilton: I think the annotator needs to be specified since it deviates from the annotator of feature_id
`str`. CELLxGENE Discover MUST annotate. If a gene_name...

schema/7.1.0/schema.md
```diff
@@ -1459,8 +1459,6 @@ When a dataset is uploaded, CELLxGENE Discover MUST annotate a unique observatio
 
 The value for each `str` key MUST be a  `numpy.ndarray` of shape `(n_obs, m)`, where `n_obs` is the number of rows in `X` and `m >= 1`. 
 
-`obsm` MUST NOT be present if <code>uns['is_pre_analysis']</code> is <code>True</code>.
```
jahilton, line 1462: This alone won't work. As now pre-analysis data MUST have one or more valid obsm.
If we go this route, we'll want to add the opt out in the obsm section (maybe something like "Otherwise, _if uns['is_pre_analysis'] is not True_, to display a dataset in CELLxGENE Explorer, Curators MUST annotate one or more embeddings")

jahilton (changes requested):
Also need to update changelog

schema/7.1.0/schema.md
nayib-jose-gloria: On the data portal UI side--the issues, as written, state to block cxg conversion (and therefore visualization in Explorer) entirely for all pre-analysis datasets. Is that going to be a problem for pre-analysis spatial datasets? 

We may want to edit the paragraph below this one to read something like:
_In order to display a dataset in CELLxGENE Explorer, uns['is_pre_analysis'] MUST NOT be True AND Curators MUST annotate **one or more** embeddings of at least two-dimensions (e.g. tSNE, UMAP, PCA, spatial coordinates) as `numpy.ndarrays` in `obsm`.<br><br>_

BAevermann: added Nayibs text. Replaced the paragraph below as it had high overlap

schema/7.1.0/schema.md
BAevermann: Updated to your suggestion. Was hoping to enshrine that the dict must exist if the "intended_target" would be annotated. With the current update, this will be implied only..

schema/7.1.0/schema.md
BAevermann: Updated! 
To me having MUST here implies that Discover will be populating the gene_ID and not just the gene_name. Is it just me?

schema/7.1.0/schema.md
```diff
@@ -1740,6 +1742,52 @@ Curators MUST annotate the following keys and values in `uns`:
...
+<table><tbody>
+    <tr>
+      <th>Key</th>
+      <td>
+        <i>feature_id</i>
+      </td>
+    </tr>
+    <tr>
+      <th>Annotation</th>
+      <td>CELLxGENE Discover MUST annotate.  The key MUST be the <code>gene_id</code> attribute from the <a href="#required-gene-annotations">corresponding gene reference</a> of the <code>organism_ontology_term_id</code>.<br><br> Version numbers MUST be removed from the <code>gene_id</code> if it is prefixed with <code>"ENS"</code> for <i>Ensembl stable identifier</i>. See <a href="https://ensembl.org/Help/Faq?id=488">I have an Ensembl ID, what can I tell about it from the ID?</a> For example, if the <code>gene_id</code> is <code>“ENSG00000186092.7”</code>, then the <code><i>feature_id</i></code> MUST be <code>“ENSG00000186092”</code>.</td>
```
nayib-jose-gloria: I think we should keep Curators MAY annotate here under 'Annotation', and move the CELLxGENE Discover MUST annotate to the 'Value' block

BAevermann: Updated!
@jahilton is this intuitively written?

schema/7.1.0/schema.md
BAevermann: Addressed in other ticket.

nayib-jose-gloria (commented):
Are we looking to bump all the ontology versions as well? If so, that'll need to be updated here as well (can be a separate PR). @jahilton @BAevermann

jahilton:
> Are we looking to bump all the ontology versions as well? If so, that'll need to be updated here as well (can be a separate PR). @jahilton @BAevermann

Yes, ontology versions should be updated as the last action like previous versions

nayib-jose-gloria (approved):

jahilton (approved):

PR #1576, opened 2026-06-03
fix: unblock PyPI publish workflow

nayib-jose-gloria:
## Problem

The **Publish to PyPI** workflow fails at the install/test step on `main`:

```
ModuleNotFoundError: No module named 'pkg_resources'
```

`bioframe==0.4.0` imports the deprecated `pkg_resources` at import time, which the current `setuptools` (installed via `pip install -U setuptools` in the build step) no longer ships, breaking `import cellxgene_schema`.

Separately, because PyPI/Test PyPI are immutable, a re-run also fails: the prior run already published the (broken) version to Test PyPI, so the "Publish to Test PyPI" step 400s on the duplicate, and the "Install and Test from Test PyPI" step would only ever re-test the frozen broken artifact.

## Changes

- **Bump `bioframe` `0.4.0 → 0.8.0`** — newer bioframe drops the `pkg_resources` import entirely. Our only usage (`bf.overlap(..., how, cols1, cols2)`) is unchanged across these versions, and 0.8.0 fits all existing pins (python ≥3.10, numpy<3, pandas>2,<3).
- **Add a local install/validate gate** on the freshly built wheel *before* any publish — this is the real gate that blocks a bad build from reaching prod PyPI.
- **`skip-existing: true`** on the Test PyPI publish + a pre-check that skips the Test PyPI round-trip test when the version is already published, so re-runs of an already-published version proceed to prod instead of failing on the immutable duplicate.

## Notes

- This branch does **not** include the `7.1.0` version bump (that lives on `feat-schema7.1.0`); it only fixes the publish pipeline.
- The workflow runs off `main`, so the fix only takes effect once merged.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/publish-to-pypi.yml
+++ b/.github/workflows/publish-to-pypi.yml
@@ -24,21 +24,36 @@ jobs:
           cd cellxgene_schema_cli
           echo "version=`python setup.py --version`" >> $GITHUB_OUTPUT
           cd ..
+      - name: Test built distribution locally
+        run: |
+          pip install -r cellxgene_schema_cli/requirements.txt
+          pip install cellxgene_schema_cli/dist/*.whl
+          cellxgene-schema validate cellxgene_schema_cli/tests/fixtures/h5ads/example_valid.h5ad
+      - name: Check if version already on Test PyPI
+        id: testpypi_check
+        run: |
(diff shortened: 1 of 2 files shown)


surygeng (approved):
LGTM

